PR #17, opened 2015-02-12
Spec 7

lipari:


garlick:
Sorry I thought I merged this.

I meant to comment that the language still needs to use the RFC 2119 words, but no reason not to get this out there and refine it later.

PR #20, opened 2015-04-21
rfc3: update CMB1 protocol after broker refactor

garlick:
* Rename `cmbd` to `flux-broker`
* Reserve `FLUX_NODEID_UPSTREAM`
* Add `FLUX_MSGFLAG_UPSTREAM`

### DIFF
--- a/spec_3.adoc
+++ b/spec_3.adoc
@@ -36,8 +36,8 @@ the following specific goals:
 
 == Background
 
-`cmbd` is a message broker daemon for the Flux resource manager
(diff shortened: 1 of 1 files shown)


garlick:
Good point.  Maybe _task_ would be better than _instance_ here?

garlick:
Does that look better?

grondo:
Yeah, I like that better. (It would be nice to have one more person's ACK though) :-)

grondo:
Actually I think this is probably safe to merge and others can comment separately. No reason to leave this update hanging.

PR #21, opened 2015-04-21
rfc8: Initial rough draft for the Flux Program Interface rfc

grondo:
Let's open up a *pending* PR for rfc8. We can capture required changes to the document here for merging into flux-framework/rfc. Once we've got something most people are happy with, 
someone can merge the doc...

### DIFF
--- a/spec_8.adoc
+++ b/spec_8.adoc
@@ -24,14 +24,19 @@ services required to execute a Flux workload.
 
 == Terminology
 
-* **program** A set of one or more tasks running under a single identifier
-  within a Flux instance.
+* **task**  A process, and its child processes, threads and fibers,
+  launched directly by flux and assigned a unique, zero origin,
+  unsigned integer rank. This SHALL serve as the basic granularity
+  of I/O aggregation and state tracking in Flux.
+
+* **task slot** A set of resources, which MAY be constrained by a container,
+  each of which MUST be sufficient to run at least one task.
+
+* **program**  A single unit of execution within a flux instance, defining
+  the environment, initial working directory, arguments, IO handling,
+  and granularity and placement of task slots for its child tasks.
 
 * **job** A synonym for _program_.
-
-* **task** A process that is part of a Flux program. Each task within
-  a program has a unique, zero origin, integer rank.  Any child processes
-  directly spawned by a task are nominally an extension of the task.
 
 * **program parameters** Data for a program that impacts how its tasks are
   executed. For example, arguments, environment, working directory.
@@ -40,8 +45,7 @@ services required to execute a Flux workload.
   program parameters to a Flux instance. 
 
 * **program context** a per-program datastore in the enclosing instance
-  where a program may store persistent state and exchange information
-  between tasks.
+  where a program MAY store persistent state.
 
 * **instance** A set of Flux framework services running within a single
   Flux _comms session_ [RFC 3], that includes a capability to launch
@@ -49,8 +53,8 @@ services required to execute a Flux workload.
 
 * **enclosing instance** The instance in which a program is running.
 
-* **task manager** A service of a Flux instance which manages tasks
-  that are part of a Flux workload.
+* **task manager** A service provided by a Flux instance for the management,
+  execution, and manipulation of tasks within a program.
 
 * **instance owner** The user on behalf of which a Flux instance is running.
 
@@ -60,7 +64,7 @@ services required to execute a Flux workload.
 
 A Flux instance SHALL offer basic services to start programs on behalf of users
 that have access to the instance. Programs SHALL only be started on resources
-to which the current Flux instance has access.
+to which the current Flux instance has been allocated.
 
 == Program Parameters
 
@@ -68,7 +72,6 @@ Program parameters accepted by a Flux instance SHALL be sufficient
 to fully define creation, distribution, and environment of tasks
 executed as part of the program. These parameters MAY include:
 
-* Scheduling Criteria (RFC XX)
 * Arguments (MAY be per task)
 * Working directory (MAY be inherited from instance)
 * Environment
@@ -76,13 +79,13 @@ executed as part of the program. These parameters MAY include:
 * Task mapping
 * Dependencies on other programs
 
-Program parameters MAY be inherited where effective. For example,
-the environment for a task within a program MAY be inhereted from
-the parent program, and a program MAY inherit its environment from
-the enclosing instance. A mechanism to override inhereted program
+Program parameters SHALL be inherited where effective. For example,
+the environment for a task within a program SHALL be inhereted from
+the parent program, and a program SHALL inherit its environment from
+the program that created it. A mechanism to override inhereted program
 parameters SHALL be provided.
 
-A Flux instance MUST support persistence of all program data for normal
+A Flux instance MUST support persistence of program parameters for normal
 program execution.
 
 == Program specification
@@ -95,7 +98,8 @@ in this document it is suggested that the program specification have
 at a minumum the following properties
 
 * scriptable interface
-* ability to describe arbitrarily complex task layout
+* ability to specify task slots
+* ability to map tasks to task slots with arbitrary complexity
 * ability to describe dynamic program properties
 * optionally include binary data (e.g. include a tarball or other
    package format of executables, input data, etc.) 
@@ -151,9 +155,8 @@ as parent, and MAY include:
 * Credentials of the enclosing instance owner
 * Contain a proper subset of enclosing instance
 
-The rank or ranks on which to run the initial program SHALL be at
-the discretion of the enclosing instance, but MAY be influenced by
-the owner during launch of program comprising the enclosing instance.
+The task slot on which to run the initial program MAY be influenced
+by the program parameters of the instance.
 
 The initial program of an instance MAY be used to further customize
 the enclosing instance, e.g. by loading extra modules, spawning
--- a/spec_8.adoc
+++ b/spec_8.adoc
@@ -144,7 +144,7 @@ A Flux instance SHALL complete and release resources upon exit
 of the initial program.
 
 Parameters of the initial program SHALL be set by the enclosing instance
-as parent, and MAY
+as parent, and MAY include:
 
 * Environment and namespace such that enclosing instance is default
   Flux instance for all subprocesses
--- a/spec_8.adoc
+++ b/spec_8.adoc
@@ -29,8 +29,9 @@ services required to execute a Flux workload.
 
 * **job** A synonym for _program_.
 
-* **task** A single process that is part of a Flux program. Each task within
-  a program has a unique identifier called the rank.
+* **task** A process that is part of a Flux program. Each task within
+  a program has a unique, zero origin, integer rank.  Any child processes
+  directly spawned by a task are nominally an extension of the task.
 
 * **program parameters** Data for a program that impacts how its tasks are
   executed. For example, arguments, environment, working directory.
@@ -46,8 +47,6 @@ services required to execute a Flux workload.
 
 * **task manager** A service of a Flux instance which manages tasks
   that are part of a Flux workload.
-
-* **task slot** One or more resources which run a single executable task.  
 
 * **instance owner** The user on behalf of which a Flux instance is running.
 
@@ -211,10 +210,13 @@ and control methods:
  * **exec** Execute all pending tasks in program P. 
    _exec(P)_
 
- * **shrink** Remove resource set R' from program P. Tasks currently
-   executing on task slots removed from P SHALL be terminated. 
-   _shrink (P, R')_. If _R' == R_ then P becomes empty again and SHALL
-   have no running tasks.
+ * **shrink** Remove resource set R' from program _P_.
+   Tasks within _P_ will be constrained to the new resource set for _P_.
+   If migration of a task to the new resource set is impossible, the
+   task MAY be terminated, stopped, or hibernated.
+   If _R' == R_ then _P_ becomes an empty program
+   and all running tasks are terminated.
+
 
  * **wait** Wait on status changes in program P.
 
--- a/spec_8.adoc
+++ b/spec_8.adoc
@@ -39,6 +39,10 @@ services required to execute a Flux workload.
 * **program specification** The language and/or API used to communicate
   program parameters to a Flux instance. 
 
+* **program context** a per-program datastore in the enclosing instance
+  where a program may store persistent state and exchange information
+  between tasks.
+
 * **instance** A set of Flux framework services running within a single
   Flux _comms session_ [RFC 3], that includes a capability to launch
   programs. A Flux instance is such a _program_.
@@ -224,6 +228,9 @@ and control methods:
   
  * **terminate** Terminate program P and _wait_ for completion.
 
+ * **reap** Transfer data from the _program context_ of a completed program
+   to the _program context_ of the enclosing instance.
+
 Other methods MAY be built using these primitives. For instance, a
 _run_ or _launch_ compound command may combine the _allocate_,
 _new_, _grow_, _map_, and _exec_ into a single interface.
@@ -245,4 +252,4 @@ calls based on security policy and ownership of the instance.
 
 As a program, a Flux instance MAY utilize any of the methods above
 as needed to make requests of its enclosing instance.
-   
+
--- a/spec_8.adoc
+++ b/spec_8.adoc
@@ -1,0 +1,246 @@
+8/Flux Task and Program Execution Services
+==========================================
+
+
+* Name: github.com/flux-framework/rfc/spec_8.adoc
+* Editor: Mark grondona <mgrondona@llnl.gov>
+* State: raw
+
+== Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to
+be interpreted as described in http://tools.ietf.org/html/rfc2119[RFC 2119].
+
+
+== Goals
+
+A core service of Flux is to launch, monitor, and handle I/O for
+distributed sets of tasks in order to execute a parallel workload.
+A Flux workload can include further instances of Flux, to arbitrary
+recursive depth. The goal of this RFC is to specify in detail the
+services required to execute a Flux workload.
+
+
+== Terminology
+
+* **program** A set of one or more tasks running under a single identifier
+  within a Flux instance.
+
+* **job** A synonym for _program_.
+
+* **task** A single process that is part of a Flux program. Each task within
+  a program has a unique identifier called the rank.
+
+* **program parameters** Data for a program that impacts how its tasks are
+  executed. For example, arguments, environment, working directory.
+
+* **program specification** The language and/or API used to communicate
+  program parameters to a Flux instance. 
+
+* **instance** A set of Flux framework services running within a single
+  Flux _comms session_ [RFC 3], that includes a capability to launch
+  programs. A Flux instance is such a _program_.
+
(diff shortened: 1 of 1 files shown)


dongahn:
Since "program" is equal to "job", we should consider adding other job states into "Program States" section later on. My job status and control API has it currently defined as the following. I reserve my right to refine those states once scheduler framework service refactoring work is done. And we should also understand if "growing/shrinking" are all that's required for elastic job management as part of work with Suraj. 

typedef enum {
    J_NULL = 1, /*!< the state has yet to be assigned */
    J_RESERVED, /*!< a job has a reservation in KVS*/
    J_SUBMITTED, /*!< a job added to KVS */
    J_UNSCHED, /*!< a job never gone through sched_loop */
    J_PENDING, /*!< a job set to pending */
    J_ALLOCATED, /*!< a job got allocated to resource */
    J_RUNREQUEST,/*!< a job is requested to be executed */
    J_STARTING, /*!< a job is starting */
    J_STOPPED, /*!< a job stopped including hitting the initial barrier for tools */
    J_RUNNING, /*!< a job is running */
   J_CANCELLED, /*!< a job got cancelled */
   J_COMPLETE, /*!< a job completed */
   J_REAPED, /*!< a job reaped */
   J_FOR_RENT /*!< Space For Rent */
} job_state_t;

grondo:
Thanks, that is a great start!

dongahn:
Program Interface section

Looks like a great start!

Perhaps we can describe the role of these program interface a bit more? I initially mistakenly thought whether you also covered the scheduler operations here as well, which you apparently did not. 

Perhaps, something like 

A flux instance SHALL support at least the following program initialization and control methods. The Flux instance SHALL use these runtime methods to relate a program to a subset (or full set) of the resources owned and managed by it and/or to relate a program to another program.

Re) allocate:
Can we use a different interface name from what scheduler uses (allocate and later perhaps grow) to reduce confusion (this at least confused me when I looked at the scheduler service framework vs. scheduler plugin division)? 

provision -- provision resources R from the enclosing instance using a resource description Rdecs ?

Re) grow
I see "grow" is the primary method that relates a program to a set of resources. Is it right? Do you cover the case where P at the point of grow has existing resources provisioned to it? Under elastic job management, P can have some existing resources, in that case, would it be better to state "R" will be added to the existing resource P already holds? And say the existing R can be an empty set as a special case...

Re) map
Do we want to have the whole set beyond map here? which includes bind and contain (or confine)t? I see program containers are described in a separate section above, but it is not clear why it is not described in the interface level. 

I would like to express sync (P1, P2) as a way to relate one set of program to another set. This will be common operations for runtime tools. 

There are other runtime service originally envisioned like record, query and split. But those either can come later into this section or be described in a separate section as an advanced interface: logging, querying etc.

grondo:
These are good discussion points, thanks.

My opinion is that scheduler parameters and operation is out of scope for this document. The scheduler will decide how and when to satisfy queued request for resources (i.e. _allocate_), but that is a black box in this rfc.

I am a bit unclear on exactly what is meant by 

> to relate a program to a subset (or full set) of the resources owned and managed by it and/or to relate a program to another program.

I am fine with using that language, but I think we'd need a definition of what "relating programs to resources and other programs" means.

> Re) allocate:
>Can we use a different interface name from what scheduler uses (allocate and later perhaps grow) to >reduce confusion (this at least confused me when I looked at the scheduler service framework vs. >scheduler plugin division)?

I am fine with another name. However, in my mind a scheduler implementation will provide a mechanism for _allocate_, so I'm not sure I follow why the program interface name should be different. 
Since _grow_ is a program interface method, if a scheduler/scheduler uses _grow_ it becomes a user of the program interface (by _grow (P, R)_).

> provision -- provision resources R from the enclosing instance using a resource description Rdecs ?

Is _provision_ your suggestion for an alternative to allocate? I would be hesitant using this term since it is already overloaded in cloud in virtual machine deployment. It tends to mean "to bring up and install necessary software". At least in my initial draft of this rfc, the _allocate_ operation simply means "to be granted access to a proper subset of resources from the enclosing instance".

>Re) grow
> I see "grow" is the primary method that relates a program to a set of resources. Is it right? Do you cover the case where P at the point of grow has existing resources provisioned to it? Under elastic job management, P can have some existing resources, in that case, would it be better to state "R" will be added to the existing resource P already holds? And say the existing R can be an empty set as a special case...

Yes, I thought _grow_ implied that resource set of _P_ would become the union of its existing resource set and the new set _R_. I will try to improve the wording here.

> Re) map
> Do we want to have the whole set beyond map here? which includes bind and contain (or confine)? I see program containers are described in a separate section above, but it is not clear why it is not described in the interface level.
 
I don't think I saw a difference between _map_, _bind_, or _contain_.

More specifically, containment is not at the program interface level. Containment is obligatory as set by the parent (enclosing instance). That is, a user running a program/job within an instance does not get to choose whether their job is run inside a container. The container would be the mandatory restriction of their job to the resources which they were granted in _allocate_.

They can, however, _map_ the processes of their program however they want within the granted resource set. Once the mapping of individual tasks to resources within the program is set, the launch phase would take care of binding tasks based on user option. Does that make sense? If not, we should continue discussing this, I may not be rigorous enough in these definitions.

>  would like to express sync (P1, P2) as a way to relate one set of program to another set. This will be common operations for runtime tools.

I was thinking _sync_ was a higher order operation built from the primitives here. I may not be understanding the _sync_ operation though. However, it seemed like a sync operation was something like:

 * `getinfo (P1) -> R1`   Use _getinfo_ method to get resource set _R1_ for program _P1_
 * `grow (P2, R1)`  Grow empty program _P2_ using resource set _R1_
 * `map (T2, P2)`  Map tasks for _P2_
 * `exec (P2)` Run tasks of _P2_

Similarly, various _query_ operations can be built from _getinfo_.

I'm a bit worried about _split_ as I don't see a practical mechanism to actually perform that operation...

lipari:
Three comments:
It seems like you are missing a word after this line:
> Parameters of the initial program SHALL be set by the enclosing instance as parent, and MAY
I think the missing word is "inherit" or "include".

While you have a section on program containers, I wonder whether you want to include in the "Program Parameters" bullet items whether a program will elect to be contained to its mapped spec.

Finally, dropping of the term "ephemeral program", leaves me wondering whether we need to mention that executions of user applications that are not started by the task manager and are therefore not registered with a Flux ID fall outside the context of the RFP.  This tells the reader that we thought of this scenario, and will allow it, but don't consider this a relevant component to the RFC.

garlick:
We define a task slot as a set of resources that run a single process, and a task as a single process.  When a task directly spawns a child process, do we say that it's outside of the task slot?  Would it be more natural to think of the task slot as a container for a task and any children?

grondo:
@garlick: good point. 

The issue I was trying to address was that a task slot is any resource set that is a valid target for a runnable process. E.g. a core+memory is a valid task slot, but a block of memory alone is not. 
A node is also a valid slot, but multiple nodes probably are not...

I don't think the task is required to be bound to its task slot, so I'm not sure what to say about children forked from tasks spawned by Flux.  We have no control over these processes, so I am unsure what to say about them. 

I'm fine with saying a task is the process launched by flux and all its children(?)

garlick:
How about:

* **task** A process that is part of a Flux program. Each task within a program has a unique, zero-origin, integer rank.  Any child processes directly spawned by a task are nominally an extension of the task.

* **task slot** A set of resources capable of running a task.

grondo:
Nice, thanks.

garlick:
Is this RFC an appropriate place to define how we signal programs, what we do with a program when one task exits, how we clean up or _reap_ a program?

lipari:
I think of task slot as a scheduling construct.  A task may need 2 cores, 4GB memory, and a gpu.  Hence one could have 4 slots available on a node and be able to schedule a task to each of those slots.  Hence, I suggest that task slot be more like, "the minimal set of resources required by a task".

grondo:
Yes, program control methods like signaling and _reap_ seem like they should be in here, if nothing else to capture the program life-cycle. We do have _signal_ and _wait_ methods for a program, but I'm not sure how to capture the _reap_ operation.

It would be great if we could figure out _reap_. It seems an important operation on instances, but I'm not sure if it is noop or what for non-instance programs.

Action taken on a program when one task exits seems like it should be left up the user, but maybe we should define a set of user-selected methods that must be supported?

garlick:
Thinking about @lipari's comment, _task slot_ is only used once in the RFC draft:

> **shrink** Remove resource set R' from program P. Tasks currently executing on task slots removed from P SHALL be terminated.  _shrink (P, R')_. If _R' == R_ then P becomes empty again and SHALL  have no running tasks.

Maybe we could leave the term out of this RFC, substituting  _resources_ above, and leave the precise definition of _task slot_ to a scheduling RFC?

grondo:
@garlick, @lipari -- good point. I didn't know "task slot" was specifically a scheduling term. In this document it is meant as a minimum/valid resource set capable of running of unix process. 

@garlick has a good suggestion above. Maybe we could even remove the language about terminating tasks. (Imagine a program running on one node with 16 cores, which is then shrunk to 15 cores -- possibly for short duration. The program may be able to absorb running all current processes on the node on 15 of 16 cores, and no tasks need to be terminated. Also, if not using binding, how could the enclosing instance or task manager decide *which* task to terminate)

grondo:
What about something like (I know, wordsmithing not good)

 * **shrink** Remove resource set R' from program _P_.
   Tasks within _P_ will be constrained to the new resource set for _P_.
   If migration of a task to the new resource set is impossible, the
   task MAY be terminated, stopped, or hibernated.
   If _R' == R_ then _P_ becomes an empty program
   and all running tasks are terminated.

garlick:
@grondo: for context, this is how we introduced _reap_ in the vision doc:

> When a job terminates, either due to a time limit or the work submitted for the job has completed, the job releases most resources back to the parent job. The control functions of the job remain and wait for an asynchronous reap operation from the parent. When the job is reaped, the parent job reads in all data from the child’s local job including its resource database. The parent then instantiates that data in its own databases. In this way, global information about jobs percolates back up through the job hierarchy, eventually back to the root job. The root job periodically updates the global, persistent databases.

garlick:
@grondo _shink_ reword good.

garlick:
It seems like a terminating program could dump an opaque blob of data for retrieval by the enclosing instance.  In the case of a terminating instance, this blob could be sifted for global resource tags or whatever else the enclosing instance wants to look for.  In other cases it might be stored as program output that could potentially provide input to other programs.

grondo:
@garlick thanks! In the case of the vision doc, were we talking about `job == instance` (?)

I was thinking along the same lines as you,  but in the case of user-level programs within an instance, is an active _reap_ needed? I was thinking programs would just write stuff in the kvs...

garlick:
@grondo good point!

I wasn't thinking about the fact that programs (including instances) have direct access to the enclosing instance's KVS.  Maybe that's all we need provided there is workable synchronization so the program has a chance to complete its dump before the enclosing instance gets impatient and SIGKILL's it.

garlick:
What about adding a new term _program context_ defined as a KVS subtree in the enclosing instance where a program can store persistent state and exchange information between tasks?   Then the _reap_ operation could be the culling of this information at the end of a job, and it could cover the PMI KVS namespace for the program as well.

grondo:
@garlick, that is a great idea

grondo:
Given the above, does a _reap_ operation simply become

 * **reap** A transfer of data from the _program context_ of a completed job to the enclosing instance's _program context_

This, at least guarantees that persistent data moves up the hierarchy as programs are reaped, and treats instances the same as other programs (since all their completed programs have data in their enclosing instance's own _program context_ datastore)

I'm not sure it is a sufficient definition though...

garlick:
Maybe good for now...

grondo:
As a placeholder I added _reap_ to the program methods section but I don't think it ultimately belongs there. It feels like we need a Program Lifecycle section or something?

garlick:
Wait shouldn't be more like:

* **reap** Post-processing of the _program context_ of a completed program by its enclosing instance.

Since presumably the enclosing instance wouldn't update _its_ program context until it completes?

grondo:
I assumed it would update as it goes. A _reap_ operation would move persistent data from child's program context to the instances own program context. (Much like _wait(2)_ collects exit status then the process goes 'poof')

This is because I was thinking of _reap_ as an operation by the enclosing instance (the parent), so the data needs to be in the program context already by the time the _reap_ operation occurs. Also, much like a normal program, the data in the _program context_ would be kept up to date during runtime, not at _reap_ time.

I don't feel strongly about this though

grondo:
Hm, yeah you are right. Post processing will probably be involved, but isn't the main point the movement of data from the child context to the parent's, so it isn't lost? Though I guess movement of data as-is isn't required so I think your description is probably more general and more correct.

garlick:
Yeah, it feels like some data reduction is missing if the program context is transferred directly to the enclosing instance's program context after every program completion.

garlick:
Not to mention that in reaping the program context, the enclosing instance may want to update other state like its resource db

garlick:
@trws might want to jump in (mentioning his name so he gets notification that we're over here now)

trws:
Thanks Jim.  I've read through the intervening comments, but I'm sure I've missed some of the subtleties.

Lots of great discussion, but before launching into a response, let me say, I'm trying to keep my response to the items that fall specifically under the program execution role of flux rather than flux as a whole (though I may not succeed...). Flux is in a relatively unique position here, in that it is a resource provisioning system, job scheduler, application bootstrap system, container manager, overlay network, and distributed database all at once.  For the purpose of this RFC, I'll try and keep myself to the purpose I derive from the title, the execution services, which I would define as the portion of flux that serves the purpose of an application bootstrap system, i.e. srun, mpd, hydra, mpirun etc.

## Terminology

1. program is defined before task but in terms of task, task is then defined in terms of program, can we make this non-cyclic? task slot also seems like a useful definition, but I see it's definition as a combination of the above discussed points. Possibly something like this:
  * Task: A process, and its child processes, threads and fibers, launched directly by flux and assigned a unique, zero origin, unsigned integer rank.  This SHALL serve as the basic granularity of I/O aggregation and state tracking in flux.
  * Task slot: A set of resources, which MAY be constrained by a container, each of which MUST be sufficient to run at least one task.
  * program/job: A single unit of execution within a flux instance, defining the environment, initial working directory, arguments, IO handling, and granularity and placement of task slots for its child tasks.  A program MUST consist of at least one task slot and one task
2. program context: It makes sense to me to define that there is a persistent context in flux storing data about a given program, but given the desire for lightweight launches, do we want to specify that it is sufficient for task data-exchange?  That implies to me a reliance on the KVS module, unless it's meant to refer to the overlay network, in which case maybe it should just be re-worded.
3. Under task manager, what is a flux workload? Given the context, I think it might be better worded something like this:
  * **task manager** A service provided by a flux instance for the management, execution and manipulation, of tasks within a program.

## Basic Execution...

Is "has access" what we want to say here? It's a nitpicky point, but I would think "has been assigned," "controls," or "has been allocated" might be closer to the mark.

## Program parameters

1. I'm thinking leave out the scheduling criteria, if this is the program/task execution specification then this is, to me, either what comes *out* of the scheduler or what is sent to a flux by a run script after scheduling has completed and an allocation initially assigned.   That would also imply that one would provide this to the scheduler, but the scheduler itself seems orthogonal.
2. Dependencies on other programs: should this be part of the program parameters, or part of the scheduling parameters? I would lean toward the latter, since I think we want to provide a "thin" layer for launching that doesn't imply services like this, but it might not be too hard to convince me otherwise.
3. Personally, I would go with SHALL be inherited, then note that an interface SHALL be provided to override inheritance where desired.  This may just be personal bias, but the admin-configuration-dependent nature of environment inheritance in current schedulers caused me a great deal of heartburn when working on Hydra, so having it be a certainty except when undesirable is attractive to me.
4. What is *program data*?  It isn't defined, and the words could imply anything from the specification, I/O and execution times to the entire input/output data and checkpoints.

## Program specification

:+1: 

My only comment here is that the task layout might be better split into a task slot specification and a mapping of tasks to task slots, but that depends somewhat on my above definitions.

## Scheduling program execution

I would remove this completely.  The program doesn't depend on scheduling logic, but scheduling depends on program execution logic.  Maybe we should have a scheduling RFC that defines its interfaces in terms of this one?

## Program Input/Output

My only comment here is that we probably want to retain the option to have a frontend program that does not hook stdout/stderr directly to its own.  Perhaps "standard output and standard error SHALL be copied to the frontend program during execution," so just removing the "to the stdout/stderr" part?

## Initial Program

This looks great for the most part.  I'm not sure I know what "rank or ranks" refers to in this context though, should that map back to task slot or task?

## Interface

I think all of this is good, but as I mentioned before it might be good to soften it a bit or define it in terms of an execution interface and a control interface, in case what's available inside and outside a given program is different.

( I can't seem to avoid these book-length replies lately... need to work on that)

grondo:
@trws: Thanks! Excellent replies. I think I will take all of your suggestions verbatim, and your improvements to terminology are really superb...

grondo:
@trws: 
>  I'm not sure I know what "rank or ranks" refers to in this context though, should that map back to task slot or task?

In this case "rank or ranks" refers the broker rank. I'm not sure the best way to word this.. but actually I think "task slot within the instance" might work given your definition above.

The point I'm trying to get across is that the initial program is not required to run on broker rank 0, and that the program parameters for the program that becomes the new instance should probably influence this.

So maybe I should just say that?

_The task slot on which the initial program runs within the instance MAY be influenced by the program parameters of the instance._

??

trws:
That makes sense to me.  It just occurred to me that you may want to soften the "SHALL complete and release resources" part, depending on how an initial program is defined for a flux instance that's meant to serve as a daemon for external access  by users.  At some level we'll probably have to have one where the initial program is just the brokers themselves, or where there is no initial program or some similar.

grondo:
Good point. It did feel like a nice simplification to say the instance "program" is complete when its own initial program completes, though.

We were heading down the road of having the "initial program" be a lot like SysV init for a flux instance, in that the initial program will initialize, load modules, and otherwise configure the instance for its purpose.  (In a way, the initial program _will_ be an extension of the broker/comms session)

To accomplish this we might have to have bindings or commands that allow the initial program to wait for certain events. Some examples (maybe bad ones -- just thinking out loud here)

 * Normal batch job
```
module load kvs
module load job,sched,....
exec $BATCH_SCRIPT_PATH
```

 * "No" initial program
```
wait-on-event "shutdown"
``` 

* Batch job with external services
```
module load kvs,job,....
exec 0 /path/to/nginx -c myconf.conf
wait-on-event "nginx-ready"
exec $BATCH_SCRIPT_PATH
```

grondo:
I pushed some of the changes suggested by @trws.
I held off on changes to the **Interface** section because I wasn't sure where you were going with the _execution_ vs _control_ interface, but I think that is a discussion we should still have.

trws:
You make some good points on the initial program, especially with an easy way to "daemonize" the instance it sounds like a practical decision.

As to the Interface section, I think all of the actions you listed are going to be necessary in some form, but I'm not sure what interfaces they will be part of or accessible from.  For example, is allocate meaningful if there is no scheduler involvement?  If not, is it really part of the program interface?  I really just may be over-thinking the whole thing on that end.

garlick:
Any objection if we merge as is and open up new PR next week if changes are needed?

grondo:
Is there maybe a way in the document to capture any unresolved items from this PR?

I have
 * Job/Program States as presented by @dongahn 
 * Interface section clarifications as proposed by @trws 
 * So far Scheduling section has been left intact, but at least @trws votes to remove it (I can't decide one way or the other)
 * Comments from @dongahn about _sync_, _bind_, and _contain_ have not yet been fully addressed

trws:
Maybe new issues referencing the PR?

garlick:
I think annotating the document would be fine at this stage.  Even just a TODO or FIXME annotation would be OK.

PR #22, opened 2015-05-27
rfc7: add std for variable names and typedefs

garlick:
There was some  discussion in PR flux-framework/flux-core#205 about names for function pointer typedefs.

I'm proposing that we add SHOULD-level guidelines for
* lower case C variable names
* upper case C preprocessor macros
* non-function typedefs: lower case ending in `_t`
* function typedefs: lower case ending in `_f` and defined as function pointers
* use typedef pointer to incomplete type to publicly define abstract data types

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,8 @@ HTML = \
 	spec_4.html \
 	spec_5.html \
 	spec_6.html \
+	spec_7.html \
+	spec_8.html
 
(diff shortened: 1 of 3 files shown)


garlick:
Oh, interesting.  I don't especially have an opinion, so I'll force an update of this PR without the break long lines patch, and my patch conforming to the style of the original doc.

PR #23, opened 2015-07-13
first draft of a best practices RFC on synchronization mechanisms in flux

trws:
This is currently a bullet-list of patterns and anti-patterns as we have touched on them.  It could probably benefit from references to issues where either the patterns or anti-patterns have shown up.  Treat this as a comment draft, and let me know what you think.

### DIFF
--- a/spec_9.adoc
+++ b/spec_9.adoc
@@ -1,0 +1,65 @@
+9/Distributed Communication and Synchronization Best Practices
+==============================================================
+
+
+* Name: github.com/flux-framework/rfc/spec_9.adoc
+* Editor: Tom Scogland <scogland1@llnl.gov>
+* State: raw
+
+== Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to
+be interpreted as described in http://tools.ietf.org/html/rfc2119[RFC 2119].
+
+
+== Goals
+
+To establish best practices, preferred patterns and anti-patterns for
+distributed services in the flux framework.  Several of the core services of
+flux, including comms, RPCs, and the KVS, provide services for distributed
+notification and synchronization, but not all of them are suitable for all
+cases.  For now, this is a listing of some common patterns and anti-patterns,
+but may shape into a more comprehensive guide as the most effective patterns
+are identified.
+
+
+== Anti-Patterns
+
+* **Watch chaining:**  The KVS provides the capability to watch keys for
+  changes in value, receiving a callback or value on change.  It is tempting
+  to chain these events together, adding a new watch inside the callback from
+  an event in order to catch the next action of a remote service.  This is
+  extremely likely to result in race-conditions unless additional external
+  synchronization is employed.  Flux services SHOULD NOT employ a pattern of
+  chained watches.
+
+* **KVS replacement updates:** When keys are used as triggers that are meant
+ to be watched, services SHOULD NOT use replacement on the value of each
+ update.  Race conditions or coalesced updates may cause updates to be missed.
+
+== Best-Practices
+
+=== General Guidelines
+
+* Services SHOULD NOT block the reactor any longer than necessary. Prefer to
+  place an asynchronous request or RPC and return to the reactor rather than
+  block for a reply, co-routines in message handlers are one way to accomplish
+  this.
+* For global notifications, prefer events.
+* For data which should be persisted and produce a notification, prefer to write to
+  the KVS in a way that supports watches.
+
+=== Preferred Patterns
+
+* **Centralize update triggers:** Services may notify completion or transition
+  events through watches on known points in the KVS. Services that use this
+  form of notification SHOULD group the points to be watched and/or retain the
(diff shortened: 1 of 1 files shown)


garlick:
By "replacement updates" below, do you mean a watch callback that does a put/commit on the key being watched?

> **KVS replacement updates:** When keys are used as triggers that are meant
>  to be watched, services SHOULD NOT use replacement on the value of each
> update.  Race conditions or coalesced updates may cause updates to be missed.

trws:
That's actually another good point, but I was actually trying to put a name to one of the issues JSC was having with wreck.  Specifically that the updates to the state replace the value of the key, and thus may lose important historical state that JSC intended to use.

garlick:
Ah.  Maybe this could be described as using KVS watch notifications for master/slave state machine replication.  The key should contain a log of states rather than the current state so that prior states can be replayed by a late joining slave?

garlick:
Incidentally, commits are only coalesced if they originate from different handles.   Successive commits on a single handle will never be coalesced.

trws:
Right, but if all states of a key's value are important then then a service should not assume that the watcher has already installed a watch when it begins to update the target.  The KVS would provide it all updates that happen strictly after the watch is placed, but that doesn't mean it would actually get them all.  If I implied that they might be coalesced, and thus fire no event in that case, it was unintentional.

garlick:
Sorry I commented on the anti-pattern without reading the patterns.  Looks like you covered my point later on.

On signaling a read from the KVS with an event:

> **KVS request and response:** Communication that needs to be logged and then
>   triggered may be written to the KVS, but the service SHOULD provide either a
>  single touch-point that it will watch in the KVS for requests or an event or
>  request name to trigger the read of the data.

There's a race in {put, commit, send_event} -> {recv_event, get} due to the "eventually consistent" KVS slave cache.  Instead, use {put, commit, kvs_get_version, send_event} -> {recv_event, kvs_wait_version, get}, where the event contains the version number.  With this synchronization, the get will return data from a cache with a version >= the version containing the put.

dongahn:
Yes, kvs_get|wait_version() will clearly establish "causal consistency" under our eventually consistency model and should be considered as a good practice. When the snapshot capability comes along, one should even be able to read the exact version contained in the send message, which I think should be encouraged as well. ..

dongahn:
BTW, this is really a good start and this will help KVS users (like me) a lot!

One thing that I am not sure is the appending as a preferred pattern: 

> Centralize update triggers: Services may notify completion or transition events through watches on known points in the KVS. Services that use this form of notification SHOULD group the points to be watched and/or retain the updates made by appending to rather than replacing values.

We talked about appending as opposed to updating before, but as I am thinking about this I'm more and more unsure if we can recommend this as a preferred pattern. 

For example, one of the anti-patterns says:

>
KVS replacement updates: When keys are used as triggers that are meant
to be watched, services SHOULD NOT use replacement on the value of each
update. Race conditions or coalesced updates may cause updates to be missed.

So, lwj.next-id would also be better to be rewritten either as a global notification or appending. I can see the use of global eventing, but I don't see the advantages of appending much. In particular, when we want to deal with 100K jobs it seems this list will grow unnecessarily large. I can see the use of appending when the number if triggers is manageably small, though. Perhaps, we can consider qualifying the recommendation a bit more?

BTW, do you have an example of the watch points grouping and how this can benefit? This might help clarify why this benefits...

>
group the points to be watched

grondo:
> So, lwj.next-id would also be better to be rewritten either as a global notification or appending. I can see the use of global eventing, but I don't see the advantages of appending much. In particular, when we want to deal with 100K jobs it seems this list will grow unnecessarily large. 

Yes, maybe we could qualify this by saying something like

"Value replacement *MAY* be recommended if the value itself may have unbounded growth and/or proceeds through exactly known state (e.g. monotonically increasing numbers, etc.)"

(Not that I'm recommending next-id as the way to allocate program identifiers for all time, but perhaps the point above still stands as @dongahn pointed out)

trws:
Good points all.  I've updated the RFC to address some of the concerns.

As to grouping keys, the specific reason there is that it would be much easier to watch updates on a large set of keys by having a directory above them to watch rather than trying to watch each key.  Sometimes watches on specific sub-keys will be the only way to go, but where possible I really think it would be better to fall back on events where that would otherwise be the case.

dongahn:
>
... but where possible I really think it would be better to fall back on events where that would otherwise be the case.

Yes, I thought you covered this well in the general guideline!

I understand why a watch on a directory where you group those sub-keys can give you much better results. In particular, I think this will be most useful when snapshot watch support is provided. As is, though, on a directory watch trigger, you can't atomically collect the state of the sub-keys inside this directory, I think this can lead to some other race issues. Perhaps, we can qualify this a bit in relation to upcoming snapshot capability?

trws:
We can.  In a more general sense, I think we need to make it clear that the watch mechanism should not be relied upon to receive all updates to given state.  As with file-system watches, they're more triggers that something has happened telling you to figure out what happened.  Clearly the KVS watches are more capable than standard file system ones, but the model is such that I tend to treat it as a side-channel or convenience rather than a way to implement reliable ordered communications.

dongahn:
Agreed.

PR #24, opened 2015-07-18
update rfc7 abstract type rule

garlick:
Although we still have prominent examples in our code (e.g. `flux_t`), I believe we concluded in PR #219 that abstract types should not obscure the fact that they are a pointer.


dongahn:
Yes, I personally like not hiding pointers better ;-). This looks good to be merged as far as I'm concerned.

Issue #25, opened 2015-07-22
Create RFC that specifies the form of a resource request and resource allocation


lipari:
In the simplest sense, when a job is allocated a resource, it is assigned that resource to use for a period of time.  A “job” is a synonym for a Flux program as defined in [RFC 8](https://github.com/flux-framework/rfc/blob/master/spec_8.adoc).

A scheduler’s role is to search, select, and allocate appropriate resources for each job over the course of time.  To do this, a scheduler implements a mechanism for annotating each resource with the job ID which it has either allocated or reserved to a job.

Subsequent resource searches for idle resources will thereby have a means to identify which resources that are in use, reserved, or available for new jobs.

Back in February 2015, we discussed resource allocation ideas on the flux-discuss list under the subject, "Resource Allocation Considerations".  A summary of those ideas was generated and this elicited more ideas and suggestions.

This issue identifies the need to capture in an RFC the ideas that were proposed in that thread.  [RFC 4](https://github.com/flux-framework/rfc/blob/master/spec_4.adoc) presents and defines the Flux resource model.  This issue identifies the need to build on that work in two ways:
* to specify the form of a request for resources
* to specify the rules for annotating resources with their associated jobs

Both of these specifications should avoid implementation details and instead provide guidance to be reflected in all implementations.

lipari:
As the person who submitted this issue and the person who modified [RFC 4](https://github.com/flux-framework/rfc/blob/master/spec_4.adoc) to address this issue, I am closing this issue.

PR #26, opened 2015-07-23
rfc: Add resource request and annotation guidance to RFC 4.

lipari:
This addresses rfc [Issue 25](https://github.com/flux-framework/rfc/issues/25) and organizes the ideas and recommendations generated in a Feb 2015 email discussion.

### DIFF
--- a/spec_4.adoc
+++ b/spec_4.adoc
@@ -18,10 +18,10 @@ be interpreted as described in http://tools.ietf.org/html/rfc2119[RFC 2119].
 
 As in traditional resource management software, the Flux framework
 requires a method for the description, configuration, tracking, and
-assignment or consumable and other resources in the system being
+assignment of consumable and other resources in the system being
 managed. The Flux Framework, however, requires a more generalized and
 flexible definition of resources and overall approach. This document
-describes the basic reosurce model used by Flux, and how it relates
+describes the basic resource model used by Flux, and how it relates
 to other resource management components. We term the model for
 describing resources in Flux _The Flux Resource Model_.
 
@@ -75,6 +75,14 @@ store basic resource _pool_ data (attributes of resource) as well
 as _composite_ data (hierarchical parent/child or ``has-a'' relationship
 between resources).  Data required to be stored, tracked, and queried
 is detailed in sections below.
+
+The Flux Resource Model SHALL support multi-granularity scheduling and
+management schemes. In such a scheme, the higher the Flux instance is
+in the Flux hierarchy, the coarser resource granularity it MAY be
+configured to operate at. For example, a higher-order Flux instance
+MAY be configured to operate at the racks and aggregates on their
+containing nodes while a lower-level instance MAY actually operate at
+the nodes and cores as the finest resource granularity.
 
 ==== Resource pool data:
 
@@ -131,11 +139,10 @@ Tag (K, [V]):: A method for tagging resource pools with
  arbitrary key/value pairs SHALL be provided. The value _V_ SHALL
  be optional.
 
-Aggregation:: A method for returning resource contents of composite object
- _in aggregate_ SHALL be provided. The aggregate method MAY be implemented
- by traversing all nodes of hierarchy rooted at the current object and
- accumulating the sum of available resources by type name. Resources with
- an available count of 0 SHALL be pruned from the traversal by default,
+Aggregation:: A method for returning resource contents of composite
+ object _in aggregate_ SHALL be provided. The aggregate method SHALL
+ return the sum of available resources by type name. Resources with an
+ available count of 0 SHALL be pruned from the results by default,
  since the composite model implies that all children of an unavailable
  resource are themselves not available.
 
@@ -171,7 +178,7 @@ Copy:: A method for copying a resource composite to a new instance SHALL
 
 Duplicate:: A method for duplicating an entire hierarchy SHALL be 
  provided. This method SHALL return a copy of of an existing hierarchy
- without any other unecessary changes.
+ without any other unnecessary changes.
 
 Merge:: A method for merging one Resource Pool into another SHALL be
  provided. The _merge_ method SHALL allow a Resource Pool at one URI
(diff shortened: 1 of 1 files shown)


grondo:
Thanks for doing this work. Sorry it took so long to get around to reviewing it!!


Quick comments
 * Formatting: Markdown allows you to associate a new paragraph with the preceding bullet item by indenting 4 spaces. There are couple places in this PR where that might be needed.
 * Formatting: should the important RFC words like "should" "shall" and "may" be capitalized as we've done in other RFCs? ("SHOULD" "SHALL" "MAY" "SHALL NOT" etc?)
 * I'm not sure on this one, but do things like "an implementation suggestion..." belong in an RFC, or should this be SHALL/SHALL NOT type stuff?
 * Specifically I'm not sure I agree on the use of the "size" parameter to indicate level of sharing of a generic resource, so I'd pull that line from the RFC (pending agreement from others)
 * Other prose in this PR may need to be reworded to fit RFC style, e.g.
        
        If a request contains elements that are not feasible or invalid, 
        then the job submission fails and the job request fails to enter the system.
    This could probably become a bullet below the following item that says "Invalid or infeasible requests SHALL terminate with an error and the request SHALL NOT enter the system" or similar.

    There may be other prose that could be reworded this way as well.

lipari:
@grondo, thank you for your comments.  I'll have an update to this PR soon.  One convention that I was attempting to impose may have been a bad idea.  I made the official SHALL's and MAY's bullet items (a convention established in RFC1).  I felt that some of those bullet items needed a context or introduction.  Hence, I prefixed some of the bullet items with paragraph(s) of text that did not contain any directives.  They were not meant to append to a prior bullet item as it may have looked.  For example:
```
Each resource can be allocated to one or more jobs. A shared or exclusive allocation is governed by two parameters. One is a characteristic of the resource itself.
* There shall be a means to stipulate whether a resource can be shared among multiple jobs.
The other parameter governing the level of sharing is the job resource request.
*  There shall be a means for a job resource request to stipulate whether the job requires exclusive use of a resource or whether it will accept a resource that is shared with other jobs.
```
Perhaps this style is too confusing and I should just stay with the bullet items alone.

grondo:
Ah, perhaps I read over the resulting rfc too quickly. I'd be interested to hear opinion of other project members. And I'll give the rfc another read tomorrow

garlick:
I would definitely capitalize all the RFC 2119 language  as we've done in all the other RFC's.  It makes the important stuff easier to spot.

I think I'd vote for turning prose plus bullet into a slightly more formal paragraph combining the two.  I sort of did that in RFC 3, although what you're doing here is I think harder to explain.

The RFC is a design document of sorts, and IMHO one indication that the design is good is if it can be made clear without a lot of explanatory prose.  When I worked on RFC 3 and RFC 5 for example, I had already prototyped the code but I iterated several times on the design when I attempted to document it because my earlier designs made for a complicated explanation.   As the design improved, I was able to delete explanations.  This is one benefit of going through the exercise of creating an RFC.  I'm not sure if that applies here exactly, but it felt like an epiphany to me at the time, so I thought I'd share.

lipari:
Reworked this PR to address the issues and recommendations identified above.  Thank you for your suggestions.

grondo:
Nice, thanks for fixing up prior usage of shall, may, should etc.

This looks good to me, you seem to have distilled the text nicely down to real requirements.

My only remaining comment is perhaps we should define what is meant by "composite resource granularity"? I think I know what is meant here based on earlier discussions, but when first reading the text I was confused on how granularity applies to a resource pool.

lipari:
@grondo, here is a rewording of the "composite resource granularity" paragraph, complete with embedded examples.  Let me know if this addresses the ambiguity you noted and whether I should leave the examples in place or delete them.
```
The granularity the composite data SHALL vary depending on the needs
of each Flux instance.  While higher order Flux instances MAY manage
composite resources to a limited granularity (e.g., seeing nodes as
the most granular composite), child instances MAY manage a more
detailed representation of the resource composite (e.g., managing
specific cores within each node).
```

grondo:
The example helped me grasp the specific meaning of *granularity* here, though now I suppose that the term itself is specific enough to not actually require the examples. I would like to defer to another project member's opinion for this case.

Thanks!

dongahn:
@lipari: Thank you for doing this. While you are at this, here are some feedback on the earlier sections: I will look at the request section later. 

1. The first sentence in the background : or -> of. 

2. The second sentence: reosurce -> resource

3. Drop "Conceptual Model" in a title: Conceptual Model: The Composite Resource Pool

4. URI (in “name:/path/to/resource” form) - just out of curiosity, do we have a resource representation (DAG?) where a graph node will not have a unique URI path? In such a case, will such a "hierarchy" require to store more than one path? 

5. 
----
Aggregation

<CUT>
The aggregate method MAY be implemented by traversing all nodes of hierarchy rooted at the current object and accumulating the sum of available resources by type name. 
----
This sentence is an implementation detail. For scalability, we talked about the user of the model (e.g., a scheduler) caching key aggregates on lower levels at higher-levels. 

6. 
---
Duplicate

    A method for duplicating an entire hierarchy SHALL be provided. This method SHALL return a copy of of an existing hierarchy without any other unecessary changes.
---

unecessary --> unnecessary

7. Since serialize operation is a primitive, don't we need deserialize to reconstruct a model from a serialized object?

lipari:
Just pushed an update to this PR containing the feedback of @grondo and @dongahn.  The only item of @dongahn which I did not act upon is item 3.  "Conceptual Model" seemed like an ok label to me.  The granularity discussion awaits any improvement @dongahn may suggest.

dongahn:
OK by me to keep "Conceptual Model:" My thinking was, since everything in that section should be about a "model" not "implementation" so I thought having that was redundant. But it's a minor point anyway. 

Regarding the granularity:

> The granularity the composite data SHALL vary depending on the needs of each Flux instance. While higher order Flux instances MAY manage composite resources to a limited granularity, child instances MAY manage a more detailed representation of the resource composite.

From Stephen's recent experience, we will probably end up needing to coarsen resource granularity at higher order scheduling (in addition to job aggregation) to improve scheduling scalabiliy further. But we won't know which specific schemes will work well until we have an exploration project.  Given that, we will probably want to make our description purposefully a high level and use "MAY" terms. Something like: 

> The Flux Resource Model SHALL support multi-granularity scheduling and management schemes. In such a scheme, the higher the Flux instance is in the Flux hierarchy, the coarser resource granularity it MAY be configured to operate at. For example, a higher-order Flux instance MAY be configured to operate at the racks and aggregates on their containing nodes while a lower-level instance MAY actually operate at the nodes and cores as the finest resource granularity.

lipari:
Sounds good to me.  Just updated PR with @dongahn's suggestion.

grondo:
This looks great to me and I give my ACK. If @dongahn clicks the merge button we can consider it a double ack.

I do have one more comment but I think this should perhaps be a new Issue.

In the recent presentation to software group here there was a note of some confusion in use of the term "pool" to describe a resource with a size component. In this RFC we term any generic resource as a "Composite Pool" because it may be composed of other resources (including itself) each of which may themselves be composites. In the aggregate, the individual resource may appear as a "pool" of generic resources.

I think the confusion came when we talked about the *representation* optimization where a set of indistinct resources are represented by a scalar size instead of a hierarchy of individual resources of size 1. While technically this is a pool, so is the node that contains the memory, and the cluster containing the node, etc. 

Do you think the terminology here is confusing enough that we should choose a different term for the individual resource of size > 1? Chris M. had suggested a "scalar" resource and I kind of liked that. 
(and in fact, all resources are scalar in that they have size >= 1 -- though there may be some future use for non-scalar resources, e.g. size of 0, perhaps for something like administrative containers or grouping of resources that wouldn't be considered "schedulable")

The discussion of granularity here made me think of this, and I kind of wonder if the aggregation of a composite resource should be more of a first class property here since technically a single resource plus its aggregation *is* the scalar representation. (If that makes any sense. And sorry for going off on a tangent here)

lipari:
There are three notions that need to be distinguished:
* A composite resource is comprised of one or more "child" resources, each of which could be comprised of "children" of their own.
* A scalar resource (to use Chris's term) is a resource made up of a number of indistinguishable units, any subset of which can be allocated to a job.  Scalar resources would have size > 1.  Another characteristic of a scalar resource is that it does not have child resources.  Something that did have a child resource would by definition be classified as a composite resource.
* An aggregate resource collapses a composite resource and all of its scalar resources into a sort of "summary" at the parent level of the parent resource and all its children and grandchildren.

I believe the trouble is that we have been using the term "pool" to represent a scalar resource as well as an aggregate resource.

I actually prefer the "pool" term to describe the scalar resource, and resource composite to describe a resource comprised of other resources, including resource pools.  An aggregate could then be seen as a kind of tag or label attached to a parent resource that summarizes all the child and grandchildren resources contained by that parent.

However, I acknowledge that this interpretation may be the minority opinion.  A quick search for the term "resource pool" comes up with a description of _hierarchical_ resources.

So, as a new commit to this PR, I'm willing to go back through rfc4 and disambiguate "pool" to refer to the resource composite and introduce a the "scalar" term to refer to the undifferentiated resources used to model things like memory.

dongahn:
I think it is actually a good idea to distinguish the two different cases. I am not sure if 'scalar resource' is the way to go, but I agree calling both cases as "pools" is confusing. 

If we want to distinguish two cases, I would see if a "scalar resource" sounds natural to the type of resources we want to model with this: memory, SSD capacity and bandwidth, switch bandwidth, ... what else.

I do think that we will also want to use the term 'aggregate' distinctively as well. 

The scalar number in a "scalar resource" represent itself whereas a "aggregate" represents what's under it. I'm unsure if we want to call this "aggregate resources" as opposed to just "aggregates."

grondo:
In this RFC, we are attempting to describe a model that can be mapped to *all* types of resources. Initially we called that model the "Composite Resource Pool" because this model can represent any kind of resource we could conceive of at the time. Perhaps the name is a bit confusing?

However, we shouldn't have to distinguish resource types that are different from a "Composite Pool".
All resources should fit this model or we've failed to adequately define the model.

If we assume this purpose then I think that differentiating "scalar" vs "pool" is probably not germane to this document. I meant only to point out the confusion between the document and the implementation, which seems to treat resources with size > 1 as a special kind of "pool".

Aggregate is a method on a resource, not a type of resource. A "composite" is by definition the aggregate of its children (recursively), so an aggregate resource could by a synonym, but we should probably just stick to aggregate as a method that provides a summary of the current composite.

lipari:
@grondo and I are in agreement that this PR can be merged pending @dongahn's review.  As it stands, RFC4's definition of "composite resource pool" can apply to a composite resource as well as a resource with size greater than 1.  Implementation decisions as to whether resources with size greater than 1 need its own name or are subject to unique limitations are outside the scope of this RFC at this time.

Issue #27, opened 2015-08-09
Add commit style to RFC 7


garlick:
@grondo listed some guidelines for commit formatting in flux-framework/flux-core#315

Perhaps RFC 7 should be renamed to _7/Flux Coding and Change Management Style Guide_ and be amended to include guidelines for well formed git commits?

grondo:
Good idea. Another title could be *Flux Contributing Guide* or similar? Standard practice to cover these topics seems to be  a `CONTRIBUTING` file in the top level of the src tree...

grondo:
Just noticed that if you place `CONTRIBUTING` or `CONTRIBUTING.md` in top level of your project, github will create a link to the file whenever someone creates a PR or Issue:

 https://github.com/blog/1184-contributing-guidelines

I'm thinking once we update RFC 7 we can make a standard `CONTRIBUTING.md` that links to this RFC and can be placed in the top level of any Flux Framework projects.

grondo:
Reading back through [RFC 1](https://github.com/flux-framework/rfc/blob/master/spec_1.adoc) I noticed some of this is covered in very high detail in the C4.1. Still ok to add the commit message guidelines and more detail on creating "patches" to RFC 7, or would you rather enhance C4.1?

garlick:
Lets amend C4.1 and keep RFC 7 focused on coding style then.  See PR #28.

grondo:
I think this issue can be closed. I submitted an example `CONTRIBUTING.md` to flux-core that references the updated RFCs here...

PR #28, opened 2015-08-12
add some patch requirements to C4.1

garlick:
Amend C4.1 per issue #27

Does this capture the main things we wanted?

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -53,7 +53,6 @@ link:spec_9{outfilesuffix}[9/Distributed Communication and Synchronization Best 
 Establishes best practices, preferred patterns and anti-patterns for
 distributed services in the flux framework.
 
-/Flux Task and Program Execution Services]::
 == Change Process
 
 The change process is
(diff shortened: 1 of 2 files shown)


grondo:
Ok, here's what I had staged against RFC 7. I think most of this is overlap, but there are some more specifics that you can take or leave.
```
 * Commits SHOULD only contain logically related changes. Where possible,
   a commit SHOULD contain only a single logical, atomic change.
 * Commits SHOULD NOT break the testsuite. Where this is impossible due
   to other rules, this rule MAY be relaxed.
 * Commit messages SHOULD be written in the imperative (Fixes or Fix).
 * Commit titles SHOULD be 50 characters or less where possible.
 * Commit titles MAY denote the section of code being changed with
   a tag followed by a colon, e.g. `name: short description`.
 * Commit titles SHOULD NOT include a period.
 * Commit message body SHOULD include a description of the change being
   made and its reason and/or purpose.
 * Commit message body SHOULD be wrapped at 72 characters or less. An
   exception SHALL be made "non prose" lines like list items, quoted text,
   or quotes from other commits.
 * Where applicable, commit message body SHOULD reference an Issue by
   number (e.g. "Fixes #33")
```

grondo:
This looks good to me if you are satisfied with the results.

garlick:
Yes I'm happy.

PR #29, opened 2015-09-15
initial jobspec rfc commit

trws:
First attempt at a job/resource specification RFC based on flux-framework/flux-core#354.  There is significant supplementary material to this RFC available as a [gist](https://gist.github.com/trws/64cf1fa659dda96a12e3).  That contains a version of all of our previous examples in yaml, along with the result of canonicalizing them with a python parser, also included, and outputting them in each of a number of different human readable serialization formats for comparison.

No guarantees that the python parsers are perfect, this has been in quite a bit of flux and still needs refinement.


grondo:
Thanks so much for doing this! I am reading through and collecting a list of comments, but this rfc proposal is an excellent help to get our discussion started and directed.

I'll throw out a few initial comments, I apologize if these are naive:

 1. From a user's perspective, I find the `range:` keyword confusing, especially for the common case where a user only specifies a `min` value. Would it be possible/advisable to have either an alternate key `count:` or something similar that either replaces `range`, is a synonym, or special keyword that does not allow a range?
 2. Should resource `name:` actually be called `type:`?
 3. I'm worried about the 20 copies of hostname example, which should be very simple but looking at both the CLI and yaml versions is pretty confusing. I think the resource request is ok `Node[5]` (though we might think about how to make the short form more readable, e.g. `<type>=[range]`), but the `-t 5` I think would throw people for a loop. The actual yaml example is even a little more confusing. 
 4. The links-in links-out thing is starting to make sense to me, though I worry a bit that users will never quite get it. Perhaps if we have `with:` without a `<` or `>` operator imply what you'd think, this problem will go away, and the common case would look a lot like a sensible hierarchical request?
 5. I actually expect that the "crazy" example from OAR will actually be quite common. Users at least will commonly want their job to, by default, run with resources all from the same cluster.
 6. The use of common resource like License seems a bit awkward in the gist examples. Why can't a rshard for a license (with 0 tasks slots) be appended to the shard list like in @hautreux's early examples?

grondo:
> I'm worried about the 20 copies of hostname example, ...

Actually I'm backing away from this statement a bit now. Maybe it is the commandline args that bother me, if it was
```
  --rshard="Node[4]" --tasks-per-shard=5 hostname
```
or preferably
```
 --rshard="Node=4" --tasks-per-shard=5 hostname
``` 
That makes sense. Also I think the yaml example would be less confusing if it were written without the "program defaults" style, e.g.

```
resources:
  - name: Node
    count: 4
    - tasks:
      command: hostname
      count: 5
```
(Sorry if I didn't get the yaml quite right but you probably get the idea how readable this is)

trws:
Thanks for the feedback @grondo.

1/2. The names for some of these certainly could use repair.  I went with range, from count, to express what it could take, but there's no reason we couldn't call it count and leave it with the same semantics.  I've been trying to avoid special-casing things wherever possible, but we could always do that if it causes too much confusion to leave it general.  Name is basically the same way, I had it as type originally but changed it for some reason after looking over the RDL RFC again.  I agree that should change back.
3.  Seeing your subsequent response, the yaml you gave is actually exactly right, modulo the name change (range->count), and long-form command line arguments certainly help clarify.  The short-form bracketed syntax for the range, while it differs from OAR's use of `=`, helps with the parsing and at least to me `thing[<num-or-slice>]` evokes "create an array of `thing` of size `num` possibly sliced or strided" where `thing=num` means "assign `thing` to value `num`.  It's an easy enough change if we want to go the other way, the only trick is that either the range syntax or some of the operators will need to change to disambiguate it for the parser.
4. I have no issue with making `with:` special cased that way.  We could even say that any key on a resource other than those that are pre-defined are implicitly outbound links of the given type, but I'm not sure if we want to go that far.
5. I agree they'll want to run on the same cluster, it just happened to be the longest most complicated example in the OAR example list.  It's rather simpler than some of the ones we've been trying, though it exposes some questions about expressing attribute requests, there is no way to put in an operator to ask for something like "attribute[x] < y" right now, so that's something to think about.
6. That's a good point.  For some reason I'd been thinking about it as "each task uses the license, and each must have a link to it" and writing it that way.  The way you put it, just associating it with the program, it could be done like this:
```yaml
resources:
 - IntelLicense
 - type: Core[10]
   tasks:
#Or
--rspec='Core[10]$,IntelLicense' #not sure I showed this, it takes a top-level list, or a yaml construct
```
Actually the simplest way to get a link from the same one to all of them would be to do this, but I thought it might not make immediate sense...
```yaml
type: IntelLicense
allocation: exclusive
<using: Core[10]$
```
or more explicitly
```yaml
type: IntelLicense
allocation: exclusive
links:
 - type: using
   direction: in
   targets:
    - type: Core
      count: 10
      tasks:
```

lipari:
With the publication of [Heterogeneous Resources and MPMD](http://slurm.schedmd.com/SLUG15/Heterogeneous_Resources_and_MPMD.pdf) to the SchedMD site, I will now include some comments w/r/t Flux's plan to support the scenarios described in the doc...

@dongahn:
If the lauching of the set of these heterogeneous tasks is not bulk synchronous but is time dependent, (which seems like the case), this can benefit from dynamic scheduling. And another fuctionality would be a workflow scheduling. 
[...]
BTW (I will have to look at Tom's job shape spec a bit closely but)  it seems one can specify such asymmetric resource requirements much more cleanly with it.

In any case, it may be good to have a proxy for this pattern similarly to minipipeline. As we have more emerging patterns captured in proxies, it seems we will have a better chance to evaluate the effectiveness of our early prototypes (cap, hierarchical dynamic scheduling and work flow scheduling in the future) and adjust them.

@trws:
Thanks for this Don. I'm glad to see this is a case people want, it makes motivating some of the ancillary side-effects of the syntax I've been working on easier. It would work a bit differently, in terms of how its specified, but has native support for it because of the way that tasks can be nested in multiple places under a programs resource list. If using the current draft, full long form, the example from their slide 6 would look like this:
`#implicit enclosing program
resources:
  - type: Node
    tags: Big
    tasks: ./controller
  - type: Node
    count: 1000
    with: GPU
    tasks: ./worker
  - type: Node
    tags: IO
    tasks: ./saver`
Command-line form would be something like this:

`-r "Node>-tags->Big" ./controller -- -r "Node[1000]>GPU" ./worker -- -r 'Node>-tags->IO' ./saver`

@grondo:
That seems pretty nice! I'm hoping Matthieu will comment on your pending
rfc to get the discussion bootstrapped again. We should add this example
to your list since CEA is clearly interested in it.

I'm still not completely familiar with this short form. Why
is ">-" required in "Node>-tags->Big", what is the difference between
">-" and "->" here?

Also, we might want to avoid use of "--" here to separate groups since
that almost always means "stop option processing", though I can see here
how it is quite convenient so I'm not totally opposed. (No matter what
it is going to be difficult if not impossible to support multiple task
specification on command line like this if the commands themselves also
have args... so maybe futile to try?)

@trws:
I agree it's a little odd, but I like it because it means we can just re-run the same command-line parser repeatedly to get each component. As the CEA folks did, some mpiexec/mpirun variants use colon, so we could certainly go that way. I'm not married to either one. The short-form syntax is more in flux than the rest anyway, we can always refine this in just about any way.

Oh, and the tags thing is because specifying a link type and distance works like this `>-type[range]->` where type and range are optional, and when both are missing the dashes are optional. There has to be some extra separator there to disambiguate between resources linked to one another and a link with a type.

hautreux:
The different things we discussed in issue #354 seems largely sufficient to provide what is necessary for what we called ARRM (Asymetric Resources Requirements Model == heterogeneous resources and MPMD) for Slurm.
I was hoping to find time to comment this afternoon but did not find enough time. I will try to comment longer this evening.

trws:
Thanks @hatreux, I'm going to try and pop in an update today to address some of @grondo's points, clarify the type decay for each parameter and I think simplify the overall structure some.



Sent with Good (www.good.com)

lipari:
@trws, You mentioned yesterday that you were favoring the term "group" in favor of "shard".  After that discussion, I realized I think the term "slot" best conveys the construct of a set of resources selected for a task.

trws:
That makes sense @lipari, I think I'll use slot for that purpose.  Group is meant to be an arbitrary grouping of resources at a point in the tree, which may have tasks or not though, perhaps use both and have Slot imply tasks where Group does not, but may have them added?

lipari:
Sounds good.

hautreux:
To start my bigger comment, I would first say that I agree with @grondo comments 1,2,5,6.

Concerning comment 4, I still 've not completely get the links stuff :) and that makes me think that I would have a hard time to explain that to my users. I still need to work on that.... I get the with> but not all the other logics exactly.

Concerning comment 3, I think that there is different examples with different ways to attach the tasks to in the CLI that disturb me a bit. The YAML representation are corrects yet maybe too much simplified in my mind to be clear. Syntactic sugar/shortcuts eventually lead to incomprehension for me. I would rather have more redundant information in the YAML but a clearer view of what I ask for. Concerning the CLI and the 20 hostname examples, I would be in favor of something like : 

```
CLI version:  -n 4 -r 'Node' -t 5 hostname
```

this better outlines the fact that I asked for 4 slots of one node each and that I want an oversubscription ratio of 5 per slot.

**-r 'Node[4]' -t 5 hostname** means for me that I want one slot of 4 nodes and that I want to launch 5 tasks on this slot (instead of the default of 1 task per slot) (something that you were doing with the distributed DB example).

the interest of introducing the **-n** to express the count of slots that we want is that we can then have the compatibility with historic syntax considering the Core as the default slot type. Thus, we could have : 

* to request 20 hostname on 20 cores : 
```
CLI version : -n 20 hostname
```
* to request 20 hostname on 4 cores, 5 tasks per core : 
```
CLI version : -n 4 -t 5 hostname
```

This could be used to describe multiple slot-groups too, like in MPMD : 

```
CLI version : -n 1 -r 'Node...' ./controller  : -n 1000 -r 'Node>GPU' ./worker : -n 10 -r 'Node...' ./saver
```

We could by the way map a single **-N x** to  **-n x -r 'Node'** to be even more compatible with past syntax and help users do the transition and do not see any difference for classic jobs.

To take into account that last example I would adapt the RFC glossary with something like : 

* Slot/Shard: the finest-grain element that must be allocated in respect of an associated count (default=1) when allocating the requested resources. Might be associated to a task description an an optional task count to specify the amount of task replicas to start per shard (default to one).

* slot-group/shard-group: a counted group of slots/shards that can be identified inside a job by a relative id/name. (something similar to the notion of pack-group in the MPMD pdf page 12, really useful when it is times to launch some apps in only some slot-groups and so on)

* Job : the highest level of organization in a jobspec, a list of one or more slot-groups.

* * *
Concerning the license example, I thought that the tasks information added to the resources requirement description was some kind of annotations that were not mandatory. In this case we could just skip the task description of the license slot and put that in its own slot-group. The core slot group could depend on the license slot-group using its relative id. I am wondering if finer grain dependencies (at core level) really brings something.

By the way, I thought that we could have jobspec without task information at all in some cases ? Is it still possible with the proposed RFC ? I am not sure  as the task logic seems to be a central piece there. Subsystems not requiring them, for exemple a subsystem in charge of nesting a flux instance inside a more global one (to create a multi-tenant sub-cluster) would not add such annotations and only focus on describing the required resources for the nested instance, would it ?
The task centric logic is what we have come up with for Slurm/ARRM (aka H.Res&MPMD), but our discussions made me recently think that having this kind of annotions was better. Maybe it is something that you have made evolved since our last exchanges on that ?

* * *

Concerning the Memory[4096]GB description, I would be in favor of putting the suffix inside the brackets as it can be seen as a human readable form of the numerical value and is better understandable. So more something like : 

```
Memory[4096GB]
```

* * *

Concerning the switch example, I would have nested the switch under the cluster resource and placed the nodes under it like in : 

```
slot-group:
    - relative_id : 0
    - name: Cluster
       with>:
        - name: Switch
           tags: Infiniband10G
           with>
            - name: Node
              with>:
              - name: Group
                count: 2
                task :
                   command : whatever
                with>:
                  - Core
                  - Memory[4GB]
walltime: 4h
```
```
CLI version : -n 2 -r 'Cluster>Switch(tags..)>Node>(Core,Memory[4GB])' 
```

That is all I have in mind right now, I need to focus on the links now :-)

hautreux:
I forgot that one : concerning the proposed range syntax, have you considered the following syntax : 

```
[min-]max[/[stride-op]stride-val]][,...]
```

It enables to have things like : 

```
4
1-100
1-10,20-30
1-10/2,2-11/2
9-300/^3
```

which seems more compatible with pdsh/clush and other cluster related tools to manipulate nodeset and idset IMHO. It could help when cuting/pasting ranges while manipulating them from CLI to CLI.

hautreux:
I missed the ''a resource become a slot when associated to a task'' in the RFC in my previous read, my wonderings about the importance of the notion task in the RFC has no sense, please ignore that in my previous comment. It perfectly makes sense now.

trws:
On 22 Sep 2015, at 14:42, hautreux wrote:

> To start my bigger comment, I would first say that I agree with 
> @grondo comments 1,2,5,6.
>
> Concerning comment 4, I still 've not completely get the links stuff 
> :) and that makes me think that I would have a hard time to explain 
> that to my users. I still need to work on that.... I get the with> but 
> not all the other logics exactly.
>
> Concerning comment 3, I think that there is different examples with 
> different ways to attach the tasks to in the CLI that disturb me a 
> bit. The YAML representation are corrects yet maybe too much 
> simplified in my mind to be clear. Syntactic sugar/shortcuts 
> eventually lead to incomprehension for me. I would rather have more 
> redundant information in the YAML but a clearer view of what I ask 
> for. Concerning the CLI and the 20 hostname examples, I would be in 
> favor of something like :
>
> ```
> CLI version:  -n 4 -r 'Node' -t 5 hostname
> ```
>
> this better outlines the fact that I asked for 4 slots of one node 
> each and that I want an oversubscription ratio of 5 per slot.
>
> **-r 'Node[4]' -t 5 hostname** means for me that I want one slot of 4 
> nodes and that I want to launch 5 tasks on this slot (instead of the 
> default of 1 task per slot) (something that you were doing with the 
> distributed DB example).
>
> the interest of introducing the **-n** to express the count of slots 
> that we want is that we can then have the compatibility with historic 
> syntax considering the Core as the default slot type. Thus, we could 
> have :
>
> * to request 20 hostname on 20 cores :
> ```
> CLI version : -n 20 hostname
> ```
> * to request 20 hostname on 4 cores, 5 tasks per core :
> ```
> CLI version : -n 4 -t 5 hostname
> ```
>
> This could be used to describe multiple slot-groups too, like in MPMD 
> :
>
> ```
> CLI version : -n 1 -r 'Node...' ./controller  : -n 1000 -r 'Node>GPU' 
> ./worker : -n 10 -r 'Node...' ./saver
> ```

You may be able to convince me on this, but I'm a bit hesitant about 
having `-n` be "number of slots" in a generic sense.  Part of the reason 
for it is that users can specify say `-r 'Node[10]>Core[12]' -n 5`.  
What does it mean in this case?  Without the `-n` it means you get 120 
tasks, each running on a core where they execute in groups of 12 on 10 
nodes.  With `-n` does it mean you only run 5 tasks on this allocation 
somehow, or does it mean you replicate the spec 5 times or..?  There is 
no way to direct what it targets in the resource spec is what bothers 
me, which makes its meaning in various corner cases pretty ambiguous.  
Then again, part of it might be that I find the current behavior 
ambiguous, does `-n 1 -N 50` give you 50 nodes or 50 cores for example?  
It's not actually explicit in  its current form.

>
> We could by the way map a single **-N x** to  **-n x -r 'Node'** to be 
> even more compatible with past syntax and help users do the transition 
> and do not see any difference for classic jobs.
>
> To take into account that last example I would adapt the RFC glossary 
> with something like :
>
> * Slot/Shard: the finest-grain element that must be allocated in 
> respect of an associated count (default=1) when allocating the 
> requested resources. Might be associated to a task description an an 
> optional task count to specify the amount of task replicas to start 
> per shard (default to one).
>
> * slot-group/shard-group: a counted group of slots/shards that can be 
> identified inside a job by a relative id/name. (something similar to 
> the notion of pack-group in the MPMD pdf page 12, really useful when 
> it is times to launch some apps in only some slot-groups and so on)
>
> * Job : the highest level of organization in a jobspec, a list of one 
> or more slot-groups.

These definitions do not immediately make sense to me.  The slot/shard, 
at least presently, must be associated with a task to be a slot, 
otherwise it's just a resource or resource group.  Job is definitely the 
top, but actually a jobspec is a list of programs, each of which has 
configuration parameters and a resource/task tree

>
> * * *
> Concerning the license example, I thought that the tasks information 
> added to the resources requirement description was some kind of 
> annotations that were not mandatory. In this case we could just skip 
> the task description of the license slot and put that in its own 
> slot-group. The core slot group could depend on the license slot-group 
> using its relative id. I am wondering if finer grain dependencies (at 
> core level) really brings something.
>
> By the way, I thought that we could have jobspec without task 
> information at all in some cases ? Is it still possible with the 
> proposed RFC ? I am not sure  as the task logic seems to be a central 
> piece there. Subsystems not requiring them, for exemple a subsystem in 
> charge of nesting a flux instance inside a more global one (to create 
> a multi-tenant sub-cluster) would not add such annotations and only 
> focus on describing the required resources for the nested instance, 
> would it ?
> The task centric logic is what we have come up with for Slurm/ARRM 
> (aka H.Res&MPMD), but our discussions made me recently think that 
> having this kind of annotions was better. Maybe it is something that 
> you have made evolved since our last exchanges on that ?

There is an option for making it implicit.  Currently all executable 
leaf nodes of a resource query tree are automatically made slots, by 
adding an implicit task to them, if no tasks are specified in the tree.  
Basically `Node[10]>Socket[5]>Core[5]` would create 250 tasks at the 
core level, with whatever parameters the user included, or if it's 
specified to create an instance it would create 10 brokers that "own" 
250 cores.

>
> * * *
>
> Concerning the Memory[4096]GB description, I would be in favor of 
> putting the suffix inside the brackets as it can be seen as a human 
> readable form of the numerical value and is better understandable. So 
> more something like :
>
> ```
> Memory[4096GB]
> ```

That's a good point.  Having a "units" parameter be the way someone 
specifies that a resource is a pool is still making me a bit uneasy, 
this is one corner I just don't like much, but that might help make it 
more readable.  I'll have to see what it would take to make sure that 
would parse unambiguously.

>
> * * *
>
> Concerning the switch example, I would have nested the switch under 
> the cluster resource and placed the nodes under it like in :
>
> ```
> slot-group:
>  - relative_id : 0
>  - name: Cluster
>     with>:
>      - name: Switch
>         tags: Infiniband10G
>         with>
>          - name: Node
>            with>:
>            - name: Group
>              count: 2
>              task :
>                 command : whatever
>              with>:
>                - Core
>                - Memory[4GB]
> walltime: 4h
> ```
> ```
> CLI version : -n 2 -r 'Cluster>Switch(tags..)>Node>(Core,Memory[4GB])'
> ```

This works if the Switch is a parent of the Node in the physical 
hierarchy, but it probably isn't.  Imagine a machine that is built like 
this:

```yaml
type: Cluster
with:
	- type: Rack
	  id: 0
	  with:
	    - type: Switch
		  id: s0
	    - type: Node
		  names: node[0-10]
		  <connected_to>: s0
	- type: Rack
	  with:
		- type: Node
		  names: node[0-10]
		  <connected_to>: s0
```

Now the with links can't be used that way, but the connected_to graph, 
representing the network can.  Does that make any sense?

trws:
On 22 Sep 2015, at 14:56, hautreux wrote:

> I forgot that one : concerning the proposed range syntax, have you 
> considered the following syntax :
>
> ```
> [min-]max[/[stride-op]stride-val]][,...]
> ```
>
> It enables to have things like :
>
> ```
> 4
> 1-100
> 1-10,20-30
> 1-10/2,2-11/2
> 9-300/^3
> ```
>
> which seems more compatible with pdsh/clush and other cluster related 
> tools to manipulate nodeset and idset IMHO. It could help when 
> cuting/pasting ranges while manipulating them from CLI to CLI.

We can try both, but I went with the [base:max:[op]stride] option to 
make it reminiscent of slicing syntax from python or go etc. and also to 
distance it some from hostlist syntax. The current version is using 
hostlist entities for expansion on `nodes:` or `ids:` **instead of** 
`count:`, it makes specifying machines a great deal easier, but means it 
may be important to make it clear which is which.

hautreux:
2015-09-23 0:54 GMT+02:00 Tom Scogland <notifications@github.com>:

> You may be able to convince me on this, but I'm a bit hesitant about
> having `-n` be "number of slots" in a generic sense. Part of the reason
> for it is that users can specify say `-r 'Node[10]>Core[12]' -n 5`.
> What does it mean in this case? Without the `-n` it means you get 120
> tasks, each running on a core where they execute in groups of 12 on 10
> nodes. With `-n` does it mean you only run 5 tasks on this allocation
> somehow, or does it mean you replicate the spec 5 times or..?


I would say :

* **-r 'Node[10]>Core[12]'** means one slot made of 10 nodes with 12 core
on each, 1 task for that slot
* **-r 'Node[10]>Core[12]' -t 120** means one slot made of 10 nodes with 12
cores on each, 120 tasks on that slot using an oversubscribe ratio of 120
* **-n 10 -r 'Node>Core[12]' -t 12** means 10 slot of 1 node with 12 cores,
12 tasks per slot using an oversubscribe ratio of 12
* **-n 120 -r 'Core'** means 120 slots of 1 core, 1 task per slot

In the last case, we lose the information that we want to group the tasks
by 12 on 10 nodes. In the case just before we lose the information that we
want 10 separate nodes. So this is not ideal I agree.

However, the initial proposal for the CLI is also ambiguous to me. if **-r
'Node[10]>Core[12]'** means **120 tasks , each running on a core**, how do
you ask for 10 tasks, each one of them running on a node with 12 cores
using the CLI ?




> There is
> no way to direct what it targets in the resource spec is what bothers
> me, which makes its meaning in various corner cases pretty ambiguous.
> Then again, part of it might be that I find the current behavior
> ambiguous, does `-n 1 -N 50` give you 50 nodes or 50 cores for example?
> It's not actually explicit in its current form.
>
-n1 -n50 is a syntax error. Slurm prints a warning and transforms that into
-n 50 -N 50 automatically.




> These definitions do not immediately make sense to me. The slot/shard,
> at least presently, must be associated with a task to be a slot,
> otherwise it's just a resource or resource group. Job is definitely the
> top, but actually a jobspec is a list of programs, each of which has
> configuration parameters and a resource/task tree
>

ok so at least a task must be associated to a jobspec, and by default it
would be a flux instance ?


>
> There is an option for making it implicit. Currently all executable
> leaf nodes of a resource query tree are automatically made slots, by
> adding an implicit task to them, if no tasks are specified in the tree.
> Basically `Node[10]>Socket[5]>Core[5]` would create 250 tasks at the
> core level, with whatever parameters the user included, or if it's
> specified to create an instance it would create 10 brokers that "own"
> 250 cores.
>

Ok, but I am still confused by the ambiguous usage of
`Node[10]>Socket[5]>Core[5]` that could mean different things depending on
what level is eventually associated to the slot. The YAML has the answer as
it is were the tasks will be associated but the CLI has no clue for that
and as the CLI is supposed to build the YAML, I think we miss something.


> >
> > * * *
> >
> > Concerning the Memory[4096]GB description, I would be in favor of
> > putting the suffix inside the brackets as it can be seen as a human
> > readable form of the numerical value and is better understandable. So
> > more something like :
> >
> > ```
> > Memory[4096GB]
> > ```
>
> That's a good point. Having a "units" parameter be the way someone
> specifies that a resource is a pool is still making me a bit uneasy,
> this is one corner I just don't like much, but that might help make it
> more readable. I'll have to see what it would take to make sure that
> would parse unambiguously.
>

ok


>
> >
> > * * *
> >
> > Concerning the switch example, I would have nested the switch under
> > the cluster resource and placed the nodes under it like in :
> >
> > ```
> > slot-group:
> > - relative_id : 0
> > - name: Cluster
> > with>:
> > - name: Switch
> > tags: Infiniband10G
> > with>
> > - name: Node
> > with>:
> > - name: Group
> > count: 2
> > task :
> > command : whatever
> > with>:
> > - Core
> > - Memory[4GB]
> > walltime: 4h
> > ```
> > ```
> > CLI version : -n 2 -r 'Cluster>Switch(tags..)>Node>(Core,Memory[4GB])'
> > ```
>
> This works if the Switch is a parent of the Node in the physical
> hierarchy, but it probably isn't. Imagine a machine that is built like
> this:
>
> ```yaml
> type: Cluster
> with:
> - type: Rack
> id: 0
> with:
> - type: Switch
> id: s0
> - type: Node
> names: node[0-10]
> <connected_to>: s0
> - type: Rack
> with:
> - type: Node
> names: node[0-10]
> <connected_to>: s0
> ```
>
> Now the with links can't be used that way, but the connected_to graph,
> representing the network can. Does that make any sense?


Yes that make sense, but that may be hard to explain in a CLI. Maybe a
hierarchical switch agregation logic expressed in the CLI could be mapped
to a set of connectedto links that should be search in the resources
description ? It could help to simplify the description of the requirements
while still using a graph backend for resources description ?

hautreux:
2015-09-23 1:07 GMT+02:00 Tom Scogland <notifications@github.com>:

>
>
> We can try both, but I went with the [base:max:[op]stride] option to
> make it reminiscent of slicing syntax from python or go etc. and also to
> distance it some from hostlist syntax. The current version is using
> hostlist entities for expansion on `nodes:` or `ids:` **instead of**
> `count:`, it makes specifying machines a great deal easier, but means it
> may be important to make it clear which is which.
>
>
ok, I would be in favor of one single way to express ranges for hosts: ids:
nodes: ranges:. That could help to read things.

trws:
On 23 Sep 2015, at 8:36, hautreux wrote:

> ok, I would be in favor of one single way to express ranges for hosts: 
> ids:
> nodes: ranges:. That could help to read things.

Much as I agree, they don't actually mean the same thing.

```
type: Node
names: hype[1,3-5]
```
means
```
- type: Node
   names: hype1
- type: Node
   names: hype3
- type: Node
   names: hype4
- type: Node
   names: hype5
```

It's a textual expansion operator, not a range, where:

```
Node[3:5]
```

means "match any number of nodes between 3 and 5" in a query context, 
and is either undefined or "replicate this element 5 times" in a 
resource definition context.

hautreux:
Thanks @twrs, I better see the distinction now. May make sense to
differentiate like you do.

Le mer. 23 sept. 2015 17:40, Tom Scogland <notifications@github.com> a
écrit :

trws:
On 23 Sep 2015, at 8:34, hautreux wrote:

>
> I would say :
>
> * **-r 'Node[10]>Core[12]'** means one slot made of 10 nodes with 12 
> core
> on each, 1 task for that slot
> * **-r 'Node[10]>Core[12]' -t 120** means one slot made of 10 nodes 
> with 12
> cores on each, 120 tasks on that slot using an oversubscribe ratio of 
> 120
> * **-n 10 -r 'Node>Core[12]' -t 12** means 10 slot of 1 node with 12 
> cores,
> 12 tasks per slot using an oversubscribe ratio of 12
> * **-n 120 -r 'Core'** means 120 slots of 1 core, 1 task per slot
>
> In the last case, we lose the information that we want to group the 
> tasks
> by 12 on 10 nodes. In the case just before we lose the information 
> that we
> want 10 separate nodes. So this is not ideal I agree.
>
> However, the initial proposal for the CLI is also ambiguous to me. if 
> **-r
> 'Node[10]>Core[12]'** means **120 tasks , each running on a core**, 
> how do
> you ask for 10 tasks, each one of them running on a node with 12 cores
> using the CLI ?

Interesting. You're thinking top down where I'm thinking bottom up.  
I'll be interested to hear what everyone else thinks here, since we 
could make either work I'm sure.  In the current model, when no tasks 
are specified, all executable leaf resources implicitly become slots.  
The result is that `Node` will do what you expect without specifying a 
number of tasks or anything.  Because of the task specifying a slot 
thing, it's actually pretty flexible.  To answer your question 
`Node[10]$>Core[12]`, where `$` is a task marker and partial task 
specification syntax though I keep meaning to make that `!`, means ten 
nodes each with one task and at least 12 cores.  Perhaps it makes more 
sense as yaml, where this would be what a user might put in a batch 
script or something:

```yaml
type: Node
count: 10
tasks: #makes this a slot level, inherits command from outside
with:
	type: Core
	count: 12
```

Doing it the other way means the number of tasks and number of slots has 
to be independent, so we would have to add a slot level to hold the task 
definition to make it work I think, though there may be an easier way I 
haven't thought of.

> -n1 -n50 is a syntax error. Slurm prints a warning and transforms that 
> into
> -n 50 -N 50 automatically.

Fair enough, it still may give you 50 nodes or 50 cores depending on 
cluster configuration however, unless `--exclusive` (sorry if I got that 
wrong) is specified?

> ok so at least a task must be associated to a jobspec, and by default 
> it
> would be a flux instance ?

Right, or a command specified as part of the command line in a CLI 
context.

>>
>> There is an option for making it implicit. Currently all executable
>> leaf nodes of a resource query tree are automatically made slots, by
>> adding an implicit task to them, if no tasks are specified in the 
>> tree.
>> Basically `Node[10]>Socket[5]>Core[5]` would create 250 tasks at the
>> core level, with whatever parameters the user included, or if it's
>> specified to create an instance it would create 10 brokers that "own"
>> 250 cores.
>>
>
> Ok, but I am still confused by the ambiguous usage of
> `Node[10]>Socket[5]>Core[5]` that could mean different things 
> depending on
> what level is eventually associated to the slot. The YAML has the 
> answer as
> it is were the tasks will be associated but the CLI has no clue for 
> that
> and as the CLI is supposed to build the YAML, I think we miss 
> something.

There is an implicit addition of the task to all compute-capable leaves 
when no tasks are specified, so it's always the bottom level unless a 
task is in there somewhere.  It doesn't have to be, but that seemed to 
hit the most cases without having to change it.  The full syntax for a 
CLI rspec is approximately the rspec item in this, where optional 
elements are surrounded by curly-braces ({}) because square-brackets are 
taken:

```
range = [<min>{:<max>{:{<op>}{value}}}]
task = {${<command>}{range}}
link-type = -{<name>}{range}-
link-target = resource | (rspec)
in-link = <{link-type<}link-target
out-link = >{link-type>}link-target
inout-link = <{link-type}>link-target
attribute = in-link | out-link | inout-link
attributes = attribute | (attribute{, attribute}*)
resource = <type>{range}{task}{attributes}
rspec = resource {, resource}*
```

Wow... that's actually a lot of stuff. Anyway, 
`Node[10]>Socket[5]>Core[5]` is the same as 
`Node[10]>Socket[5]>Core[5]$` by expansion.  If the user puts a task 
spec anywhere in there, it overrides that expansion and uses the level 
they select.  If someone were sufficiently insane they could actually do 
all of this:

```
Node[10](>Socket$"./writer">Memory[10]gb,>Socket[4]>Core[5]$"./worker"[4])
```

They would get 10 nodes, each devoting 1 socket and 10GB of memory to a 
writer process and four logical sockets, each of which must have 5 cores 
(the sockets are not exclusive, so some may be grouped together as long 
as the cores are separate) running four copies of a worker task on each 
core.  A single rspec can specify an entire program.  It doesn't have to 
stay that way, it might simplify things to make it less powerful and say 
you can only have one task specification per rspec or something, but 
this is how it's currently prototyped.

>>
>> That's a good point. Having a "units" parameter be the way someone
>> specifies that a resource is a pool is still making me a bit uneasy,
>> this is one corner I just don't like much, but that might help make 
>> it
>> more readable. I'll have to see what it would take to make sure that
>> would parse unambiguously.
>>
>
> ok

I think this can work, it might want for a separator or something to 
keep from accidentally conflicting with hex-value inputs though, unless 
we want to restrict to decimal.  I'll put this in while I get the query 
test together later today.

>>
>> This works if the Switch is a parent of the Node in the physical
>> hierarchy, but it probably isn't. Imagine a machine that is built 
>> like
>> this:
>>
>> ```yaml
>> type: Cluster
>> with:
>> - type: Rack
>> id: 0
>> with:
>> - type: Switch
>> id: s0
>> - type: Node
>> names: node[0-10]
>> <connected_to>: s0
>> - type: Rack
>> with:
>> - type: Node
>> names: node[0-10]
>> <connected_to>: s0
>> ```
>>
>> Now the with links can't be used that way, but the connected_to 
>> graph,
>> representing the network can. Does that make any sense?
>
>
> Yes that make sense, but that may be hard to explain in a CLI. Maybe a
> hierarchical switch agregation logic expressed in the CLI could be 
> mapped
> to a set of connectedto links that should be search in the resources
> description ? It could help to simplify the description of the 
> requirements
> while still using a graph backend for resources description ?
>

That sounds like a good idea, but I'm not sure what it would look like.  
Could you provide an example maybe?

Something came up about BlueGene IO nodes the other day that had me 
thinking about this.  The desire was to allocate some nodes on every IO 
node, where each node essentially always uses the IO node it's directly 
associated with.  I'm pretty sure we could get that with this syntax.  
Here the `*` is "any connection type", with a hop count of exactly 1.

```
IONode[10]>-*[1]->Node[10]
```

This would be 100 Nodes, 10 per IO node, but it doesn't actually ensure 
that all of the IO nodes are unique, because it's not exclusively 
allocating them.  Haven't thought of a way around that one yet.

Anyway, thinking about it a little more, we could get 90% of the way to 
your example like this:

```
CLI version : -n 2 -r 
'Cluster>Switch#InfiniBand,>-*[1]->Node>(Core,Memory[4GB])'
```

The switch will be part of the hardware of the cluster, so it can go in 
on with, and then a single-hop connection from the switch on any graph 
to get to the Node level.  How does that strike you?

hautreux:
thanks again @trws for your patience and explanations ! I have now seen the light...

I missed the '$' operator that enables to instruct the level at which the slot is specified in the **-r '....'** descriptions. Having that, all seems to eventually fit properly in my mind. We could still use the **-n x -r '...Something>...** as an equivalent to **-r '...Something[x]$>...'** to help with transition but I agree that it is not necessary at all.

Concerning the tagging of resources, what is the preferred way ? I have seen that you added a **?** operator for that purpose in the RFC but talked also in a comment about something like **-r 'Node>-tags->IO'**. I think it might be interesting to only have one way to manage any single thing at the end, or it will bring confusion, don't you ?

I am wondering if the notion of "Group" that you are using is not what is described as a composite resource in RFC4. If it is not the case (I am wondering if a composite resource is a resource node or a set of resource nodes readingthe RFC), the group notion should be added. If it is the case it may be interesting to use **Composite** insteadof **Group** to be coherent, isn't it ? At the end both RFC would probably have a lot in commons, don't they ?

I really like the Node[count] syntax versus the Node=count one. The array-like is as intuitive and much more readable when the command lines start becoming longer.

Should the scheduler return an **instanciated jobspec** having id-list specified according to the allocated resources based on the different provided counts/ranges ? That would make easier to then manage the start of the tasks, the affinity, ... Maybe it is something you already planned.

Concerning the tree to links mapping, now that I get (I think!) the links logic, I think that it would just be an other way to represent it with its own complications. It might not really worth the effort. At the end, as I said, having only one way to ask for every single thing could help the syntax to be usable.

I have a use case that is a real requirement on a mid-term basis for us. I think that it can be explained, I will try to explain it in another comment.

The only thing that is still a little bit confusing for me now seems to be the default-{tasks,resources,programs} and the way they interact with each other (with the &,*,<<,>> operators) and the way they are inherited/overriden. I will try to read again the RFC.

Thanks again !

trws:
On 23 Sep 2015, at 14:18, hautreux wrote:

> thanks again @trws for your patience and explanations ! I have now 
> seen the light...
>
> I missed the '$' operator that enables to instruct the level at which 
> the slot is specified in the **-r '....'** descriptions. Having that, 
> all seems to eventually fit properly in my mind. We could still use 
> the **-n x -r '...Something>...** as an equivalent to **-r 
> '...Something[x]$>...'** to help with transition but I agree that it 
> is not necessary at all.

Yeah, I think we'll want to offer a `-n` and `-N` style interface, or 
even those options, I'm just not quite sure how they should interact 
with an rspec.  Maybe if `-n` is specified the rspec is required to be 
linear? Not sure.
>
> Concerning the tagging of resources, what is the preferred way ? I 
> have seen that you added a **?** operator for that purpose in the RFC 
> but talked also in a comment about something like **-r 
> 'Node>-tags->IO'**. I think it might be interesting to only have one 
> way to manage any single thing at the end, or it will bring confusion, 
> don't you ?

I completely agree, I just keep changing my mind on what it should be.  
In the internal representation, I want to make it part of the graph, 
because that has some interesting and useful side-effects, including 
O(1) lookups of the entire list of all resources that are marked with a 
given tag, and as a result thought that having it be a link in the 
syntax would work.  I'm now thinking again that it would be pretty 
nasty, so having an attribute for it is back, and it's just a question 
of what the prefix is.  I'm now leaning toward using `#` to leverage the 
"hashtag" association, but any thoughts or comments on this would be 
appreciated.

> I am wondering if the notion of "Group" that you are using is not what 
> is described as a composite resource in RFC4. If it is not the case (I 
> am wondering if a composite resource is a resource node or a set of 
> resource nodes readingthe RFC), the group notion should be added. If 
> it is the case it may be interesting to use **Composite** insteadof 
> **Group** to be coherent, isn't it ? At the end both RFC would 
> probably have a lot in commons, don't they ?

They absolutely do.  I'm using Group, or Slot if one wants to imply a 
task association, right now, but composite would be another good label 
to use.  Generically though, a composite could be any resource spec with 
more than one resource in it, where a group is a non-hardware entity in 
the resource graph that serves to tie together resources that otherwise 
might not share a convenient common parent.  Really they're 
interchangeable, if it makes it easier to read I'd be for it.

> I really like the Node[count] syntax versus the Node=count one. The 
> array-like is as intuitive and much more readable when the command 
> lines start becoming longer.
>
> Should the scheduler return an **instanciated jobspec** having id-list 
> specified according to the allocated resources based on the different 
> provided counts/ranges ? That would make easier to then manage the 
> start of the tasks, the affinity, ... Maybe it is something you 
> already planned.

Absolutely, at least that was my plan.  There are three "contexts" of 
resource specs in my head, I'm just now realizing I never wrote this 
down, thanks for asking good questions. =)

* Specification: A resource spec representing a resource graph, may or 
may not have IDs but represents a known set of finite resources to be 
managed.  Tasks and groups are ignored in this context, and if 
concretized they may be lost.  Ranges should generally only contain 
"min" in this context, but if fully specified the maximum value is used.
* Query: Each level is interpreted as part of a query, walking an 
existing resource graph.  IDs are allowed, but constrain the match to 
only that ID, ranges are allowed and will be interpreted as the minimum 
unless a scheduler decides otherwise.
* Concretized/Instantiated: A Specification for which all nodes are 
tagged with their types and have had IDs assigned.  This is the portable 
wire-format for resources that a user shouldn't have to see, and will 
likely be a good bit larger than what they wrote.

> Concerning the tree to links mapping, now that I get (I think!) the 
> links logic, I think that it would just be an other way to represent 
> it with its own complications. It might not really worth the effort. 
> At the end, as I said, having only one way to ask for every single 
> thing could help the syntax to be usable.
>
> I have a use case that is a real requirement on a mid-term basis for 
> us. I think that it can be explained, I will try to explain it in 
> another comment.
>
> The only thing that is still a little bit confusing for me now seems 
> to be the default-{tasks,resources,programs} and the way they interact 
> with each other (with the &,*,<<,>> operators) and the way they are 
> inherited/overriden. I will try to read again the RFC.

That's completely understandable, it may be the most complicated part of 
the whole thing, because it's meant to do "the right thing" as much as 
possible without user intervention.  If anything in that seems off to 
you, definitely say so, there is a significant chance I'm missing some 
common case in there. I need to write this up more explicitly too, 
perhaps pseudo-code or similar for the RFC.

Thanks for all your feedback on this, it's been quite valuable.

hautreux:
The use case that I talked about previously is an extension of the MPMD examples that are illustrated in the PDF. The idea is to add to an homogeneous/heterogeneous SPMD/MPMD program a second program (or multiple) that would be ran simultaneously and interact with each other by the means of dedicated network connections.

The main goal is to provide an IO-Proxy program that could be used by an MPI program to access file systems through a caching logic. The IO-Proxy would act as a proxy, ran by the same user as the MPI program, and would be connected using a particular TCP port (to bootstrap an RDMA comm) allocated dynamically by the scheduler (as multiple IO-Proxy could be ran on a same shared IO node).

To do that, I think that something like the following yaml could be specified : 

```yaml
program: &ioproxy # if it is the way to give a name to that program to reference it in other programs
    - resources:
        - Name: Node
          Count: 1
          Tags: IO
          With:
               UnprivilegedTCPPort[1]
               Core[4]
               Memory[16G]
          Tasks:
                Command: io-proxy --threads=4
program: &user-mpi-app
    - resources:
        - Name: Node
          Count: 32
          Tags: Compute
          With:
               - Name: Group # or Composite ?
                 Count: 16
                 With:
                     Core[2]
                     Memory[4G]
                 Tasks:
                     Command: ./mpmd-part1 --with-io-proxy # or maybe --with-io-proxy=...something enabling to link to the TCP port allocated in the first program once allocated
        - Name: Node
          Count: 32
          Tags: Compute
          With:
               - Name: Group # or Composite ?
                 Count: 32
                 With:
                     Core
                     Memory[4G]
                 Tasks:
                     Command: ./mpmd-part2 --with-io-proxy # or maybe --with-io-proxy=...something enabling to link to the TCP port allocated in the first program once allocated
    - walltime: 4h
    - <<: *ioproxy # as a way to say that I require that and want it to be started before and stopped after ?
```
Does it sounds like what it would look like ?

As in your example for bluegene @trws, a way to get the IO nodes that are closer to the compute nodes would certainly benefit in being expressed in the requirement too.

grondo:
@hautreux, @trws -- this discussion has been really good for me too. 

I'm going to have to go back through this again before I fully understand it, but the latest example by @hautreux is really illustrative. It also seems to be an example of a special case for dependencies, in that the `mpmd-part[12]` codes are dependent on the `io-proxy` starting, not necessarily on its completion (which is a more typical dependency)

In SLURM this io-proxy functionality would probably typically be part of a job prolog, with teardown in the epilog? (Ignoring for now the fact it runs on disjoint set of resources). If we handle the do-something-before/after-user-program directly in the job spec, we may have just subsumed user-prolog/epilog functionality directly into the job spec syntax. (which seems kind of cool).

Note: perhaps getting ahead of myself below, feel free to ignore:
Eventually we'll probably want some way to `include` snippets of job spec, such that some things like the `io-proxy` service could be defined in a system file, and included in user requests (perhaps included with args like `include io-proxy (threads=4) # count defaults to 1`.

For command lines, we should allow some sort of parameter expansion, as I think @hautreux points out above. It could be as simple as allowing environment variables to be used, or perhaps something more advanced -- however, it is a good idea to me in general. This could be used to expand parameters in the include syntax.

trws:
I think all I would change would be to take the `program:` off and make 
them dashes to merge them into one "program" with multiple resource 
sets/tasks.  The way you're doing it is actually the right way, but we 
don't have the dependency model done yet to be able to say that you want 
to wait for the ioproxy to be set up before the other one is started.  
I'm leaning toward something like this for it, but keep re-working it 
because something feels off every time I write it up, maybe this is the 
one:

```
#general dependency link
dependencies:
	- direction: in | out | inout
	  queue: queue-id #this is the one I like, personally user-local deps
	  id: job-id
	  satisfied: success | runtime | exit
	  # success means only satisfy if the job didn't fail, clear the queue
	  # exit means it's satisfied regardless
	  # runtime means it can be cleared **during the run**
# for the simple case, use a link
<dependency: <queue or job id> # in
dependency>: <queue or job id> # out
<dependency>: <queue or job id> # inout
```

Given that, I would leave it as you have it but maybe add this to your 
spec @hatreaux:

```
#on ioproxy
dependencies:
	satisfied: runtime
	queue: io
	direction: out

#on user-mpi-app
<dependency: io
```

Then in the ioproxy task, once it's set up and ready, it would at some 
point run something like `flux dependency satisfy io`, at which point 
the MPI job would be free to run.  Knowing to tear down with 
dependencies is harder, I don't have a good way to do that, but this 
should be sufficient for most server alongside job type cases.

grondo:
BTW, another way to think about things like an `io-proxy` alongside job is as a flux service started by the initial program. Actually it feels a bit like the job spec here is taking on some of the role that we had talked about for the initial program. I kind of like that the job spec isn't procedural like a traditional "initial program" might be, so we might ponder whether this syntax could handle other things an initial program might do, for example

 - Set/change flux config/attr values
 - load/unload modules
 - start other flux "services" e.g. io-proxy, per-job fs
 - distribute files, fetch inputs, etc
 - and so on

Ideally some of these "services" could saved as standalone files and shared between users or via system-discoverable locations, add dynamically add cmd line options to flux-submit/run like SPANK-plugins, etc. The really neat thing is these service files could be as flexible as SPANK plugins, but if needed could define their own task-slot shapes, so flux automatically knows where to start processes or how many extra resources to add to the user's request, etc..

We could end up with something that is a combination of spank plugins and systemd unit files for starting distributed services on demand. (Eh, I should probably go home now)

trws:
It should be able to handle lots of things, like file distribution and service management, but probably not so much environment manipulation and module management, unless we make them nestable.  That's actually something Todd mentioned wanting to look into with us by the way, finding a way to come up with a module or other loader that would let a user specify the tools and env they want and leverage spack and the modules systems to provide it.  Help make running a job a little more portable.

hautreux:
@trws for his first comment since I have seen the light :) 

Maybe the '-n x' should just be considered as a multiplier, so wether or not the rspec is linear we would end up with x times the count(s) of tasks specified in the rspec associated to their x times specified slots. Most cases of '-n' usage would be linear I guess.

Concerning attributes, it may be interesting in that case to represent them using the hashtag (or an other char) in the spec, but internally in the resources desc build the graph of attributes to help in lookups.

Concerning Group/Composite, I find Composite interesting as once instantiated the jobspec becomes in some sense the resources spec of the nested allocator/flux-instance. So reusing the rfc4 terms when possible may make sense. At the end an hardware Composite versus a virtual Composite still aggregate resources that we may want to manage similarly, no ?

I like the specification/query/instantiation stages. Especially if the instantiated form (filter out of any unrelevant annotations like tasks infos & program infos) could be reused in a nested allocator/sched. However, I think that it could be great if the annotations of the jobspec could persist the instantiation stage.

I 'be still not found time to dig again into the default-* syntax/logic, but will ask for your help if necessary.

* * * 

@grondo, @trws concerning the io-proxy use case :

I have separated in 2 program specifications in the jobspec as I do not want all the tasks to be part of the same MPI application, and thus find interesting to consider that multiple programs inside a jobspec could be considered as different seq/parallel apps (lightweight jobs?) each started with its own PMI provider. Does it make sense ? At the end, programs are like tasks a kind of annotations across resources shars to specify tasks and apps as well as their dependencies.
We could add a Count field beside programs: to instruct to launch a particular number of similar program/app instead of just one (for use case similar to your capacitor ?)

In Slurm, I am thinking about adding an independent pack-group added to the mpmd job using a spank-plugin plus prolog/epilog to wait for the io-proxy and mount/unmount it. I clearly like the idea of the snippets in the jobspec. Adding tasks and programs prolog/epilog  in the tasks/programs definitions of jobspec could be great to (privileged and unprivileged when possible maybe...). That is very interesting to use that jobspec as a vehicle for informing multiple subsystems of what they should perform. Some way to sign the instantiated content may be necessary to avoid users tampering with annotations from privileged subsystems ?

You are right about the expansion I tried to exhibit. That would be very powerful for autonomous set of dependant programs to be bootstrapped. A kind of etcd-free bootstrap :) : instead of pushing+querying for info once started, tasks could get that from the instantiated jobspec (if accessible to them)

If it turns out that our ideas are not just pure fiction, it would be something very cool...

lipari:
In Monday’s meeting, I volunteered to review this PR.  I know this represents a monumental task to describe a Flux job specification in all its complexity.  But as such, my feeling is that it is overly complex.

I’m going to make a last ditch effort to propose a way to simplify the model.  I understand there has been a lot of thought that has gone into this discussion and that a lot of agreement has been forged.  So, if my ideas are off-base, I won’t be offended.  But here are my ideas for simplifying the proposal.

I’m going to restart the discussion with some simple axioms.  The first is that the scheduler schedules resources.  It is a glorified calendar manager.  It allocates one conference room to Bob from 1 to 2pm and the same room to Betty from 2 to 3.  It does not care what conversations go on in the room during that time.

The second axiom is the concept of a program specification as defined in [RFC8](https://github.com/flux-framework/rfc/blob/master/spec_8.adoc).  The program specification embodies the ability to specify a command, task slots, the ability to map tasks to task slots with arbitrary complexity, etc.

The third axiom is that there is a fundamental difference between a job request and a list of resources.  The user submits a job request for resources to the scheduler - describing all aspects of the resources required or desired.  The scheduler finds and allocates specific resources to the job.  From that point on, the job has a well-defined set of resources on which it can launch programs.

Absent from the scheduler’s purview is an awareness of the task specification.

I submit that the way to simplify the job request specification proposed in this PR is to extricate all task-related information from the job specification and move it to a (yet to be formally defined) program specification.  What remains of RFC 10 would still be complex.  We will still need the notion n numbers of resource shards and all the “has-a” and “uses” relationships that that entails.  But absent from this spec will be which tasks run on which resources.

Ultimately, all the scheduler needs to concern itself with is finding and allocation resources.  What runs on those resources is superfluous and irrelevant info to the scheduler.

Many of you have apparently “seen the light”, and if tasks are in inextricable component to a Flux job spec, then I guess I'll take another dive through the spec to find that light.  And then look for ways to make this more palatable for the user.

SteVwonder:
Overall, I like the current jobspec.  The format seems expressive without being unwieldy.

I generally support keeping the spec in yaml mainly because it would allow the spec to be parsed by almost any language (without requiring call out to a Lua program/library).

I think the only missing piece is the CLI.  There are a few examples, but no section discussing it.  At this stage, I think we should avoid trying to make a very extensible CLI and instead, just support very simple jobspecs from the command line, mainly for compatibility (-n, -N, --n-per-node).

trws:
Good point @SteVwonder.  I agree that keeping a first CLI simple is a good idea, then later we can look at doing something more dense like the old short-form syntax (once we've worked the kinks out of the canonical form that should be easier).

spec_10.markdown
```diff
@@ -42,48 +42,95 @@ with a *task*, thus making it a *slot* or *shard*.
...
-    range: 5
-    # range-type: per-slot # already the default
 resources:
-  - name: Node
-    range: 4 #leaf-level is a slot, inherits task
+  - type: node
+    count: 4
+    tasks:
+        command: hostname
+        count: 5 # replicate 5 times *per node*
```
SteVwonder: How would you specify 4 nodes with 5 tasks distributed across those nodes?

trws:
```yaml
type: group
tasks:
  command: hostname
  count: 5
with:
  type: node
  count: 4
```

Or, if that's the whole job:
```yaml
tasks:
  command: hostname
  count: 5
with:
  type: node
  count: 4
```

spec_10.markdown
```diff
@@ -225,74 +252,24 @@ A link MUST contain:
...
-      range: 1
-      range-type: per-shard #could also be total or others later
+      count: 1
+      range-type: per-slot #could also be total or others later
       affinity: bind
 default-resource: &def-res #!resource
-  - range: 1
+  - count: 1
     name: PU # Smallest allocatable unit
     allocate: exclusive #default when not specified is shared, innermost exclusive
```
SteVwonder: It might be useful to add this tag to the above 'General Structure' section.

trws: Yeah, you're probably right.  I had been contemplating just leaving this out, for simplicity's sake, but I don't think we can get away with it.  It has too much effect on everything else, I'll get that re-incorporated.

trws:
BTW @SteVwonder: WRT using yaml, I largely agree but for two considerations:

1.  If we ever need to add conditionals, or loops, or anything like that, lua would just give us all of that stuff out of the box.  For long-term compatibility and avoiding having to do custom extensions, it's really not beatable as a configuration format because of this.  That said, as you say, it requires interpretation of lua code, and isn't necessarily as easy to serialize or transform unless we're pretty draconian about it.  It's also more verbose and harder to make a short-form syntax for that would be valid as a long-form config.
2.  The state of the art for parsing yaml in C is... well, it's painful.  Getting a yaml document loaded is easy, but the most compliant C library doesn't really offer an interface for interacting with the resulting document as a reader, it expects you to use their event interface to fill your own data structures of some kind.  We can most definitely do that, but it amounts to something like 200-500 lines of code, and means we would have to establish our own data structures to fill.  Given what we're planning to do with it, we'd probably have to do that anyway, but it front-loads the work pretty heavily. 

P.S: I feel compelled to point out that the corresponding, though slightly more up-to-date and yaml-spec-compliant, C++ library version of an equivalent operation takes two lines, one of which is `#include <yaml-cpp>`...

SteVwonder:
@trws:

1. I am interested to hear what use-case you have in mind for conditionals and loops.  My gut instinct is to intentionally limit the base jobspec to a simple data language and then build an even more expressive layer on top using lua, python+cheetah, etc that ultimately "compiles down to" or "emits" the base jobspec data language.  The other cons (in addition to the language dependence) to using Lua (or any Turing-complete language for that matter) that I see are:
  1. Security: I get a little queasy thinking about executing arbitrary lua code given by a user.  I imagine there are serious security implications if there are malevolent users on the system. I'm not a security or Lua guy, so there might be safe guards in place that I am unaware of that negate this point.
  2. QoS: Even assuming that all the users are benevolent, if a user accidentally creates an infinite loop (or even just a long running loop), whatever thread is parsing the jobspec is going to be tied up.  With YAML, I would assuming the cost of parsing the jobspec would be a function of the size of the YAML string, so strings greater than a certain size could be rejected to ensure QoS.  For a lua script, this would be much harder to determine.
2. This is a good point. In comparison to parsing in C, how hard would it be to leverage your existing python parser?  Or implement the parsing in C++/Lua (and then expose the functionality through a C api)?  This kinda goes against what I was originally advocating for, but at least this way we aren't requiring a particular language to parse the jobspec.

PS - The security and QoS points are the same points I made against sending arbitrary Lua code to the KVS for execution.  So if we can solve those 2 problems in this context and make Lua work, it would be helpful in the context of the KVS as well.

SteVwonder:
FWIW, it seems as though PyYAML is as unsafe as pickle unless you use 'safe_load' which I assume limits the capabilities/ease-of-use of the parser.  Source: http://pyyaml.org/wiki/PyYAMLDocumentation#LoadingYAML

grondo:
> The other cons (in addition to the language dependence) to using Lua (or any Turing-complete language for that matter) that I see are:

I still have not had a chance to fully digest this yet, and I am by no means arguing for Lua here, but Lua itself is used as a data description language often enough that sandboxing code is pretty well understood (e.g. see the [Lua Live demo](http://www.lua.org/demo.html) which runs any script entered in a sandbox and appears to be secure enough).  However, another embedded Lua example,  Redis [EVAL](https://redislabs.com/blog/cve-2015-4335-dsa-3279-redis-lua-sandbox-escape) did have an exposed vulnerability awhile ago, so your point is certainly valid.

There are also methods to prevent bugs and malicious code from running infinite loops in sandboxed scripts. This would be quite important for the KVS use case as you said, but I'm almost positive this will have been solved by Redis or one of the many other server based systems using embedded Lua out there currently.

Another counterpoint is that I don't think YAML, JSON, XML etc are guaranteed "safe" either when given untrusted input (heck, even most low level functions aren't safe with untrusted input :wink:)

However, I do agree that the jobspec itself could be a pretty simple, if verbose and arcane markup or notation. It feels like we'll have to wrap up the parsing of the spec into a higher level API anyway, and that could be done in a library in C/C++/etc. with bindings for other high level languages like Lua/Python/golang etc.

One case I'm still pondering for the jobspec is a way to promote specification "reuse" among users. I really haven't thought this through well enough at all, so hopefully this doesn't cause more confusion. This might require some kind of syntactic sugar on top of the jobspec as laid out here, with "include", "use", or "import" directives, and perhaps some kind of inheritance system similar to resource composite idea from RDL discussions and prototype. I'm not sure how this would work with straight YAML, but it might be fine if an library is layered on top that parses canonical jobspec yaml to a configuration blob.

trws:
Good arguments.  For the record, I'm still largely for the yaml approach (honestly I just like the way it reads better), but playing devil's advocate...

I don't really have a particular use-case in mind for dynamic configuration, but I could see users wanting to use it so they could write a single input for use on multiple machines that results in different requests depending on where they're submitting from for example.  The other one that could be interesting would be to re-interpret the configuration after finalizing the resource set that it will actually receive to allow the task placement to change, but I don't know if we'd really want to do that.

As to the issues with pathological scripts, my plan for that had been to say that it is always interpreted in the client the user is running to submit a job to generate a static configuration object, then (if necessary) inside their actual allocation.  That way the user pays for it, and it can't stall the server or the launcher or anything else important.

We can certainly implement a C++/python/lua/whatever parser and manager and offer a C API with relative ease.  Honestly that's where I'm leaning more and more at the moment, particularly C++ to use yaml-cpp or go to use goyaml and provide a C api.  As to safety, yaml is safe *if and only if it is limited to producing fundamental types*.  Because it supports annotation of object types, and the pyyaml and some others support this, it can be used to create arbitrary objects in the parsing application unless that's disallowed.  Even these features can be used safely though, and I think the canonical representation would include such object annotations to help with parsing clarity and performance (using a whitelisted set of course).

You make a good point about reuse @grondo.  The yaml spec has support for "including" yaml objects inside other yaml objects, and actually for merging dictionaries (so you could have a default dict that someone uses then overrides parts of I think) but it doesn't have support for including other yaml documents or files.  We could look into having a directive that pulls in other content for inclusion, or we could say that the yaml features are enough and require it all to be in one logical document but allow backreferences and inheritance inside of that.

That said, this kind of thing is exactly why we might want to consider lua again, since it handles inclusion out of the box...

SteVwonder:
@grondo, taking a quick glance at the Redis EVAL [documentation](http://redis.io/commands/EVAL#atomicity-of-scripts) and [source code](https://github.com/antirez/redis/blob/3b644e82b0cc96113e29672d1551248c9e9db893/src/scripting.c) it looks as though they take the "we're all consenting adults here" approach.  In the documentation it warns about writing slow scripts since they are atomic, and in the source code, they have a timeout.  I'm not 100% sure, but it seems that once the timeout is reached, the Lua code continues to execute.  Redis just alerts the user and waits for a "SCRIPT KILL" command.  Combined with @trws's point about executing on the client side, this seems like a reasonable solution.  The user would only be hurting his/herself with a slow running script.

The point about conditionals for portability and imports for reuse are really good ones.  It seems that trying to do that directly in YAML is asking too much and would have to been done by a higher layer (assuming we use YAML as the lower layer).

@trws: quick clarification on your points about executing on the client side - what did you have in mind when you said "static configuration".  Did you mean some sort of data structure within Lua that is serialized out, or did you mean something like YAML/JSON?  If there is a way to fully execute the Lua client side and guarantee that the Lua object passed to the sched module is entirely static, that would be awesome!  But I think I'm overanalyzing what you said....

trws:
@SteVwonder, that's actually pretty much what I meant.  The lua code would be executed to generate a given data structure or representation that would then be static.  We could even say that whatever the user generates with would be serialized as the yaml we discuss in this PR, or a JSON equivalent or similar.  They would be asked to fill in a variable, or have their code return a specific object or string type that would have, or be trivially convertible to, the structure we want then carried around as that static representation, with the dynamic version only there for possible later re-evaluation.

trws:
Hrm... Maybe that's the answer to the question.  If we like the yaml syntax, say that's what we take, but maybe have an option for users to use lua to generate it as part of submitting a job?  Probably as part of a phase two, but I kinda like that as an option.

grondo:
I'm just feeling the need to point out that 99% of users will *not* be willing, capable, or care to use this kind of long form syntax we've been discussing here (I'm sure it has been implicit in our discussion, but I wanted to make it plain again) ;-)

Almost all users are going to want to use simple command line or very short form syntax. Even the Slurm `srun -B, --extra-node-info` option is rarely used afaics, over the simple `-N, -n, --cpus-per-task` options. Just a reminder! :-)

Therefore, as a first consideration we should keep in mind how easy it will be for us to generate the static job spec from these simpler options (probably trivial, but just in case)

I think it was mentioned before that a prototype which can emit and read in the job spec, and translate common `srun` `psub` `qsub` etc requests would be excellent as a demo for users and other stakeholders.

grondo:
> Hrm... Maybe that's the answer to the question. If we like the yaml syntax, say that's what we take, but maybe have an option for users to use lua to generate it as part of submitting a job? Probably as part of a phase two, but I kinda like that as an option.

@trws, great point and I think would be a good compromise between safety and a powerful interface.

SteVwonder:
At the sake of repeating what @trws and @grondo just said, it seems to me that the two suggestions (Lua & YAML) are converging to roughly the same point:

- The user writes their jobspec in a very expressive, high-level language.
- The jobspec is compiled/transformed client-side into a static configuration
-  The static config is serialized and sent to the sched module.

On one hand, both the high-level language and the low-level static configuration could be Lua.  On the other hand, the high-level language could be something we would have to make/define (could be in Lua/Python/etc), and the low-level language could be a strict data-language (YAML).

At this point, I think we can achieve the same level of expressivity and terseness with either solution.  I also think we could also make it portable in either case (using python+cheetah to output valid Lua isn't that hard).  Finally, I think you both have convinced me that QoS and security are manageable in either scenario. So maybe it just comes down to which is easier to implement?

trws:
Excellent point @grondo.  That's part of the reason older versions of this focused as much as they did on building a short-form syntax, but that ended up being enough of an additional complication that it seemed to stall things out.  If we like this as a canonical representation, then mapping simple (slurm/qsub-like) command line arguments to it, as well as having a CLI-friendly way to write it, is the next necessary step.

The only complication I see for the simple arguments is not in converting from them, which should be straightforward, but in what they mean when combined with a jobspec, and what they convert to.  `-N` is pretty straightforwardly a Node, but can it be combined with a spec? `-n`, if kept as number of tasks, could make sense but does it mean that's how many tasks there are that use the spec, or should the spec be replicated for each task or..?  This is what I meant by item 4 in my email about the update, I don't think working it out will be too bad, but nailing down the side-effects seems important.

grondo:
Some very quick and minor comments on the current version of the RFC in this PR:

 * Glossary:
   - **Resource** is defined as a 'single resource *vertex*' -- it might help to previously define what a vertex means in the context of this RFC. I'm not sure what, but perhaps in a **Goals** or other front-loaded section a general description of job spec as generic graph with "vertices" as resources and "edges" as relationships between those resources would help? (I'm not sure)
   - Under **Range-Type** the term "per-shard" is used, and its meaning is not entirely clear, though I'm sure we did discuss it at length at some point. Perhaps **shard** should be added to the glossary.

 * It would be interesting to relate this back to RFC 4 somehow, which defines the concept of a Resource Composite as the base resource type for Flux. It may not make sense to define job specs in terms of composites explicitly (and if we need a rewrite of RFC 4, that is completely fine), but it seems strange not to reference back to the Flux Resource Model somehow in this RFC. I'm not sure how or where to do it exactly, but thought I'd mention it here.

 * As mentioned in a recent meeting, it would be nice if `count` and `amount` could be made synonyms. However, this should not be done if it makes parsing expensive or unwieldy. But I'm not sure how someone authoring a job spec, or even a front end that "compiles to" job spec could necessarily know when to use `count` or `amount` with generic resources.

 * It feels like the meaning of "edges in" vs "edges out" could or should be explained a bit more explicitly somewhere.

 * For **Task** section, we might want to add that environment, current working directory, resource limits, etc MAY be added to the task spec or a default spec at the `program` level. Though perhaps it should be SHALL since these properties of tasks will have to come from somewhere -- traditionally it is the current environment of the submitter, unless  overridden on command line.

 * In *Resource specifications* section, there is vestigal use of `with>:` which might confuse some readers.
   - Also, in final example resource specification, it would be interesting to know how to specify different logical numbering of cores within sockets, e.g. `sockets = { "0,2,4,6", "1,3,5,7" }` in the RDL syntax.

SteVwonder:
Is **shard** synonymous with **slot**?  If so, my recommendation would be to pick one a stick with it.

trws:
It is an old name for slot, if you dig back far enough in this rather epic comment chain you'll find where we shift back and forth between the two.   I had meant to make it consistently slot, but clearly missed that one, thanks for the errata, notes and comments @grondo, @garlick and @SteVwonder, much appreciated!

garlick:
I think it might be helpful at this stage to rework this RFC in the form of our other RFCs, and maybe tighten the scope to something like "canonical job specification language".  If it would be helpful, I could volunteer to create an asciidoc skeleton for RFC as a separate PR, maybe just containing a brief introduction that defines the scope and context (that we could iterate on) and a rough outline.

Hopefully we can get consensus on that quickly, get it merged, and then fill in with additional PRs? Thoughts?

grondo:
@garlick, great suggestion! Let me know if I can help in any way

lipari:
Given that we're leveraging graph theory, I think we should make an effort to disambiguate the word, "node".  E.g., use "node" to refer to a compute node.  Use "vertex/vertices" to refer to node/nodes in a graph.

> As mentioned in a recent meeting, it would be nice if count and amount could be made synonyms.

I actually like the two different terms - specifically "amount" to request a quantity of a resource "pool" and count to request a discrete resource.  I like specifying an "amount" of memory, bandwidth, and power and a "count" of cores, nodes, and other discrete objects.  I realize this leaves the question of what to call an "amount" range.

dongahn:
Does it make sense to change walltime key to time key and the time key expresses both the walltime requirement and the start time requirement?

garlick:
Any objection if we close this PR since we've begun doing it in smaller chunks elsewhere?
(The epic discussion will of course be preserved through references in other PR's, as it should be!)

dongahn:
@garlick. This is fine by me. BTW, note that I posted one minor recently question right before your posting, as something like this will be needed to support a start time or reservation request. Do we want to consider this before closing this PR? Or do we want to discuss this as part of a smaller chunk? (Or maybe it doesn't belong here).

garlick:
@dongahn if you don't mind re-asking that question elsewhere I think we can close this.   In fact doing that now.

PR #32, opened 2015-12-15
rfc10: add spec for content storage service

garlick:
Here is an RFC for the protocol used by the content service described in PR flux-framework/flux-core#493

Differences from the PR as it exists at this moment:
* PR blobrefs are not yet sent as message payloads; we are hardwiring SHA1 raw digests at the moment
* Asking the enclosing instance to resolve unknown blobrefs is not yet implemented
* Foreign blobrefs not yet implemented

I was going to describe JSON structures for representing ordered lists and hashmaps of (blobref, type) tuples which could then be used to construct structured content like the KVS namespace, but that was not getting quite off the ground so I would like to propose getting this in now and deferring that until work on the KVS gets underway again.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,8 @@ HTML = \
 	spec_6.html \
 	spec_7.html \
 	spec_8.html \
-	spec_9.html
+	spec_9.html \
+	spec_10.html
 
 all: $(HTML)
 
--- a/README.adoc
+++ b/README.adoc
@@ -53,6 +53,10 @@ link:spec_9{outfilesuffix}[9/Distributed Communication and Synchronization Best 
 Establishes best practices, preferred patterns and anti-patterns for
 distributed services in the flux framework.
 
+link:spec_10{outfilesuffix}[10/Content Storage]::
+This specification describes the Flux content storage service
+and the messages used to access it.
+
 == Change Process
 
 The change process is
--- a/spec_10.adoc
+++ b/spec_10.adoc
@@ -1,0 +1,171 @@
+ifdef::env-github[:outfilesuffix: .adoc]
+
+10/Content Storage Service
+==========================
+
+This specification describes the Flux content storage service
+and the messages used to access it.
+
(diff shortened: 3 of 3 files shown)


garlick:
Oh yeah, another diff from the PR is there the primary service is "content-cache" and optional backing store is named "content".  Here I named the primary service "content" left off any description of the backing store as an implementation detail.

grondo:
Questions: (Sorry I'm almost positive some of these are naive)
 *  blobs have an RFC imposed size limit of 1MB? Do you envision a different store used for larger blobs, or a scheme for storing multi-part blobs? (e.g. input datafiles). Also, I suppose the limit could be there to discourage use of the content store for large files since it is in-memory?
 *  No "delete" operation for the content store? I know a blob is immutable, but does this imply it can't be unlinked from the store? (actually I think it is fine for no delete operation to exist, I'm more curious about the decision to leave it out, and perhaps thinking that a small explanation called out in the RFC could be helpful.)

Thanks!!

garlick:
Those are good questions. On the size limitation, yes I was thinking that giant blobs could be supported by concatenating smaller blobs using the JSON ordered list structure I mentioned.  I hope that's a viable approach.

In this type of a store I believe deletion implies usecounts, and I just didn't want to go there right now.  Maybe we won't ever need to go there?  I guess that is an open question, or possibly I'm living in a dream world :-)

grondo:
Well I personally think it is fine not to go there, but maybe a note in the RFC (or do you not even want to got *there*?) ;-)

garlick:
I added a little section on garbage collection
> References to content are unconstrained from the perspective of the content service, therefore content MUST persist for the lifetime of the instance.
>
> During instance shutdown, some content MAY be preserved by storing it in the enclosing instance when the instance is reaped. All other content SHALL be destroyed when the instance terminates.

Does that make sense?

grondo:
Yeah. Sounds great! Want to squash it all together?

garlick:
Ok, all squashed now.

PR #33, opened 2015-12-16
rfc10: add implementation overview, misc cleanup

garlick:
Describe the distributed cache implementation briefly at the beginning
of the implementation section.  The dropcache and flush operations
don't really make a lot of sense without this context.
Mention an OPTIONAL backing store, but punt on the details, calling
it out of scope for this RFC.

Define more closely what flush does, and make it mandatory.

Define more closely what dropcache does, make it mandatory,
and limit when ENOSYS can be returned.

Fix a couple spelling errors.


grondo:
Sorry I missed some of those typos in review!

PR #36, opened 2015-12-21
rfc10: fix blobref example

garlick:
I changed the description of blobrefs to be lower case per discussion, but forgot to update the example.


grondo:
Sorry I should have caught that one.

PR #37, opened 2015-12-30
rfc11: add spec for KVS tree objects

garlick:
This is not complete but I wanted to get it out there for any comments ASAP.

Currently the KVS uses an internal JSON representation for directories.   I needed to fix that in conjunction the content storage integration PR so that directories and values greater than 1MB can be represented.  At the same time I thought to document and provide some general tools for manipulating KVS namespaces so that there need not be only one, so that "snapshot references" could be passed around, etc..  Some other ideas are listed in the Goals section.

Anyway, just a sketch at this point.  Maybe others will have some interesting thoughts on where this could go.


garlick:
I rebased this on pr#43 (spellcheck) and added some words to the dictionary for this spec.

Even though this is very preliminary I would like to get this merged after the spellcheck PR, as it's a good reference for planned changes to the KVS.  It can evolve as the problem gets worked.

PR #38, opened 2016-02-02
rfc1: add a Release Process section

garlick:
As discussed in flux-framework/distribution#8, this PR adds a section called Release Process to RFC 1, and requires that framework projects use semantic versioning.  I also added a requirement for git annotated tags, and dist tarballs (if applicable).

### DIFF
--- a/spec_1.adoc
+++ b/spec_1.adoc
@@ -141,6 +141,20 @@ C4.1 is meant to provide a reusable optimal collaboration model for open source 
 
 * Maintainers MAY commit changes to non-source documentation directly to the project.
 
+* Autotools products, if applicable, SHOULD NOT be checked into the project
+  revision control system
+
+=== Release Process
+
+* Releases SHALL be tagged with git annotated tags.
+
+* Release names SHALL employ version numbers that follow the
+  Semantic Versioning 2.0.0 standard, (C.f. http://semver.org).
+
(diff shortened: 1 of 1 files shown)


garlick:
Were there any objections to this?  Anybody want to push the button?

grondo:
Oh sorry, totally missed this PR!

My only nit is that the last bullet under Release Process  "Autotools products SHALL NOT be checked into revision control system" should be moved up into Development Process section?

Otherwise seems extremely reasonable.

lipari:
My feeling is that the proposed mod may belong more with the distribution project than with rfc1.  rfc1 deals mostly with process and "rules of order".  The proposed change seems more detail focused, specifically how to build and package code.  All of the topics discussed in [Distribution Issue 8](https://github.com/flux-framework/distribution/issues/8), including the proposed mod, seem like they should be codified somewhere in the distribution project.

lipari:
P.S. My above comment is more of a subtle judgement call.  I'm perfectly fine if this PR is merged into rfc1 as it is.

garlick:
@grondo: just forced a push with the bit you identified moved to development process, which seems right to me.

@lipari, well, I think this has more to do with asking all framework projects to use similar practices for making tags and presenting releases.  Following this process doesn't necessarily imply that a project would be included in the distribution project (though it could be a requirement for inclusion) so it seems to me more appropriate to define here.

PR #39, opened 2016-02-26
RFC 3, 5: keepalives to communicate module state with broker

garlick:
Update RFC 5 with the goal of:
* begin a process of iterating to bring sched plugins in line with RFC 5 or vice versa as discussed in flux-framework/flux-sched#125
* return errors in module initialization to `flux module load` as described in flux-framework/flux-sched#13
* provide low-level, single broker basis for future distributed synchronization of module init as touched on in flux-framework/flux-core#585
* enhancing module debug capability

The basis for most of this is an idea that keepalive messages could be used to inform the broker of module state transitions such as initializing -> [ready <-> busy] -> finalizing -> exited.

### DIFF
--- a/spec_3.adoc
+++ b/spec_3.adoc
@@ -217,7 +217,7 @@ PROTO		= request / response / event / keepalive
 request		= signature version %x01 flags nodeid   matchtag
 response	= signature version %x02 flags errnum   matchtag
 event		= signature version %x04 flags sequence unused
-keepalive	= signature version %x08 flags unused   unused
+keepalive	= signature version %x08 flags errnum   status
 
 ; Constants
 signature	= %x8E			; magic cookie
--- a/spec_5.adoc
+++ b/spec_5.adoc
@@ -32,45 +32,146 @@ The goals of the Flux Module Extension Protocol are:
 * Module name should indicate what service/component the module extends.
 * Define mechanism to pass arguments to modules at insertion time.
 
+== Introduction ==
+
+In this description, "base component" is used to describe any Flux
+component that is being extended with modules, and "extension module"
+is used to describe the extension.  A Flux component may have both
+roles; For example, comms modules, introduced in RFC 3, are extension
+modules that extend the Flux broker base component.  When a comms
+module _itself_ has extension modules, it is also a base component.
+The design places no limit on the depth of the extension hierarchy;
+that is, extension modules can have extension modules which can have
+extension modules, etc..
+
+Extension modules may be implemented as "messaging actors", which
+execute in their own thread and communicate wtih the base component
+exclusively with messages, or as "plugins" that provide a set of
+well known function entry points that are called from the base
+component's thread of control.  Both types of extension module define
+common symbols and can be loaded/unloaded/listed with common management
+messages and user tools, thus reducing duplication of effort across Flux
+components.  A base component may support one or both types, according
+to its requirements.
+
+Extension modules are assigned names that indicate their position
+the extension hierarchy.   An extension module with a single word
+name like "foo" is a comms module;  its module extension names have
+"foo." as a prefix, e.g. "foo.bar";  _its_ module extension names
+of "foo.bar." as a prefix, e.g. "foo.bar.baz; etc..
+
+Base components with extension modules must handle requests to load,
+unload, and list their extension modules.  For example, a comms module
+named "foo" would implement request handlers for "foo.insmod", "foo.rmmod",
+and "foo.lsmod".  This allows a single tool to be used to manage
+extension modules for all Flux components that have them.
+
+Base components that are extended by messaging actor based extension
+modules should provide message routing services to their extension modules.
+For example, a request to "foo.bar.ping" would be delivered to "foo",
+which should have registered a request handler for "foo.bar.*".  That
+request should be routed by "foo" to "bar", which would send a ping
+response message back through the same routes taken by the request.
+
+Plugin style extension modules have no communications path to the base
+component and no thread of control, so they cannot have managed extension
+modules without a proxy arrangement with the base component.
+
+
 == Implementation
 
 === Extension Module Symbols
 
-A Flux extension module SHALL declare at least the following global symbols:
+A Flux extension module SHALL export the following global symbols:
 
 +const char *mod_name;+::
 A null-terminated C string defining the module name.
-The module name SHALL be structured as set of words delimited by periods,
-interpreted as zero or more service name words followed by exactly
-one module base name word.  If the service name is not defined,
-the service is Flux Comms Message Broker.
+The module name SHALL be structured as set of component names
+delimited by periods.
 
 +int mod_main (void *context, int argc, char **argv);+::
-A C function that SHALL be called by the service with argc, argv style
-module arguments and an opaque, service-dependent context at module
-load time.  The purpose of the function is service-specific.
+A C function that SHALL either be the entry point for a thread
+of control, or an initialization function.  This function SHALL
+return zero to indicate success or -1 to indicate failure.
+The POSIX `errno` thread-specific variable should be set to indicate the
+type of error on failure.
 
-A module loading service MAY call +dlopen()+ with _RTLD_LOCAL_ flag,
+A Flux extension module MAY export the following global symbols:
+
++const char *service_name;+::
+A null-terminated C string defining the service name provided by the
+module, if different from the module name.
+
+A base component MAY call +dlopen()+ with _RTLD_LOCAL_ flag,
 then access these symbols with +dlsym()+.
 
-A module loading service MAY create a new thread or process and
-pass control to +mod_main()+ to create a messaging _actor_.
+=== Loading an Extension Module ===
 
-A module loading service MAY use +mod_main()+ as an accessor to set
-options in a module context, and then use +dlsym()+ to access
-service-specific methods provided by the plugin in a more traditional
-extension model.
+A base component SHALL load an extension module either as a
+messaging actor or a plugin.
 
-=== Message Definitions
+A messaging actor style extension module SHALL be launched in its own
+thread or process with a communications socket connected to its base component.
+In the new thread or process, +mod_main()+ SHALL be called with optional
+arguments and a handle representing the communications socket.  When
++mod_main()+ returns, the module thread or process SHALL exit.
+While the module thread or process is executing, the base component provides
+message routing services on its behalf.
+
+A plugin style extension module's  +mod_main()+ SHALL be called as an
+initialization function in the base component's thread of control.
+The base component MAY use +dlsym()+ to access service-specific methods.
+
+=== Monitoring an Extension Module ===
+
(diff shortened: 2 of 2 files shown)


garlick:
I should add that one reason I proposed keepalives here as opposed to a single-ended request is that keepalives are just a PROTO block which is very low overhead to encode/decode.

Also, the other use of keepalives, monitoring the TBON, might be able to leverage the state, errnum fields added here for a better setup/teardown handshake in the future.

This allows us to replace the ad hoc zero length message currently sent when `mod_main()` exits with a real message type defined in RFC 3.  So that's progress too.

grondo:
This looks good to me. Wait to merge until the work in this area is finalized?

Once we're ready to merge you could squash the renaming into the original commit.

garlick:
If he doesn't mind I'd like @lipari to give this a read and see if it makes sense to him, with sched plugins in mind.

lipari:
@garlick, I have reviewed your proposed changes and they seem very well structured.  My only comment is that I failed to understand the reason for what appears to be a new level of abstraction.  After succinctly defining "flux broker" in rfc3, I do not see the reason for defining "base component" as a "flux broker" in rfc5.  I interpret "extension module" as a superset that includes "comms module".  The difference seems to be that a comms module will be reported in `flux module list` while any extension modules that a comms module loads will not.  Plus extension modules will support the "plugin" behavior while comms modules can only take on the "messaging actor" role.

Also, as far as what this new def means for the existing sched module/plugin architecture, I don't see any impediment to rewriting the sched module and two plugins to comply.

garlick:
Thanks @lipari.  To clarify:
* broker and comms module are instances (important ones) of base component and extension module
* a component can be both a base component and an extension module (sched comms module is both)
* an extension module that extends a comms module could be either type (sched comms module could have messaging actor extension modules)
* `flux module list` could list all extension modules (like sched's loaded plugins)

Maybe I need to do some editing if that didn't come across!

lipari:
Ok, that helped.  Now I get it.  I didn't think of the sched module as a base component before; I thought of it only as an extension module.

lipari:
> Maybe I need to do some editing if that didn't come across!

Instead of,
"Comms modules may in turn have their own extension modules...",
you might say,
"Comms modules may also be base components and load their own extension modules..."

garlick:
Just updated the intro and squashed it down.  Better?

garlick:
@trws: were you proposing in todays meeting that if the `mod_name` symbol is not found, we should fall back to some function, say `mod_get_name()` that would return a `char *`?  Just wanted to make sure I had it right before I put it in.

lipari:
> Just updated the intro and squashed it down. Better?

Much better!

trws:
@garlick: Yup, either that or the other way around, not sure if it matters really.  The main use-cases are: writing modules in languages that have problems making constant string literals (I.E. Go, really nice module language otherwise by the way), and dynamic module naming.  The latter means that the py_mod module could be used to load arbitrary Python modules and disambiguate them by loading names dynamically from the python.


grondo:
A comms module that handles loading interpreted modules (e.g. like the python module) could also act as a "base component" and load its own (interpreted) extension modules within its namespace. That might make sense if you want to load multiple extensions into a single interpreter state, instead of having 1:1 interpreter-to-service mapping.

I'm not even sure if interpreted extensions would fall under rfc 5, but I thought I'd just throw the idea out there.

grondo:
OK, I think you already handled the kind of thing I was talking about above as "plugin style extension modules".

garlick:
I think the pymod comms module could have either style of extension module, written in python, couldn't it?

grondo:
> I think the pymod comms module could have either style of extension
module, written in python, couldn't it?

Theoretically, sure. Though currently I think the python comms module acts
as a shim to load a single module written in python as a first-level comms
module.

Sorry, I didn't mean to dispute anything with my comment, but to make sure
interpreter-loaded plugins were not inadvertently proscribed by the
language in the rfc (right now there is language that extensions SHALL
export C symbols), and to note that there are multiple ways extensions
could load other extensions (but I noted you'd already covered that nicely
in the rfc language)

It also occurs to me that the only type of extension module that should be
*required* to follow rfc5 structure in terms of naming, symbols exported,
etc, would be a first-level comms module, loaded directly by a base
component using the flux C module loading mechanism. I guess that is
implied by the language in the document.

>

garlick:
Ideally this should be written so that its use is encouraged (if not required) throughout Flux for the benefits of
* code reuse (`flux module` command, libflux module API)
* uniform user experience

I guess the way it's currently structured with static naming does make it difficult to provide a shim like pymod for interpreted languages.  Hmm.  Pondering.

grondo:
Well a use case I was thinking of is a Lua plugin loader that might run as a comms module with service name `lua.` and exposes itself to its own plugins as a simple interface to a reactor loop. This kind of interface does not lend itself well to `flux module` style loading and listing, so it might use its own interface, and I assume would not fall under rfc5 if you let it.

Examples of tiny lua scripts that might be useful to insert on the fly might be

 * timer that kills something gracefully after a timeout
 * built-in io-watchdog like functionality for a session or an individual job
 * one-off distributed operations

Of course callbacks that trigger on msg topic strings could also offer new "services" to the session, though they'd have names like `lua.foo`.

This interface doesn't seem like something we should discourage.

Maybe we don't have to say anything here to allow non-rfc5 plugins to be loaded by extension modules and other entities, as the rfc doesn't strictly forbid this kind of thing?

The only drawback I see is that, currently, first-level services (e.g. service `foo`) can't ever be registered by anything but a comms running in its own thread of control (I could be wrong about this?), thus implying that for N services there will be N threads in the broker.

garlick:
> The only drawback I see is that, currently, first-level services (e.g. service foo) can't ever be registered by anything but a comms running in its own thread of control (I could be wrong about this?), thus implying that for N services there will be N threads in the broker.

Ah right, we should replace the static `char *mod_service` with a way to ask the broker to (un-)register service names, so `lua.foo` could register to receive `bar` requests via `lua`.

trws:
Yeah, having both mod_name and mod_service offer a dynamic mechanism would go a long way to making that kind of thing work nicely.  The current language seems to make these either not extension modules, or not compliant though.  Would it be sufficient to say that top-leve, or base or whatever wording, modules must conform to the API, but that a module defining its own list/insmod/rmmod is allowed to define its own API to its modules as long as it presents the necessary functionality on the flux side?

trws:
By the by, I should have noted this before, but it would be most useful if `get_mod_name` would have the same signature and take the same arguments as mod_main.  That way shim modules like pymod can serve up the name of their client, without necessarily requiring N:M module loading support.

garlick:
It's useful to have extension module .so's retain their static module name so that `flux module load` can search for them by name without relying on the filename.  It's fine if that's a function call as opposed to deferencing a static symbol, but IMHO it starts to go a different way than is proposed here if the names of extension modules have to be passed in as arguments to that function call.

As I understand it, the problem seems to be with service names.  The module name is implied to be a service name and originally that was all we had.  Currently we also have `mod_service`, a static service name that can be different than the module name, e.g. `content-sqlite` and `content-sophia` both provide the `content-backing` service (not simultaneously).   If the service name was dynamically registered by the extension module, then a shim module could register one or more services for each of its extension modules, and it wouldn't necessary have to be hierarcical, e.g. `pymod.foo` could register  plain `foo`.

For example, 
```C
int flux_service_register (flux_t h, const char *service_name);
// implicit at handle close also
int flux_service_unregister (flux_t h, const char *service_name);
```

Is that heading towards something usable?

trws:
I think so, but I'm not sure it completely addresses what I was hoping to do with pymod.  That said, it might allow something similar enough to be usable instead.

The issue with pymod is if it gets loaded twice, with two different sub-modules, those shouldn't conflict (in an ideal world).  One simple way to do that is to let it override its own name, either during the get_name call or during mod_main.  Another would be to implement the kind of sub-module interface that @grondo has been talking about, such that the sub-modules would be disambiguated.  Neither of these, it occurs to me now, actually require having the arguments ad mod_name time.

grondo:
To clarify what I think @trws is going for is that a pymod python module would be loaded like
```
flux module load pymod /path/to/script.py
```
The module/service "name" is encoded in the `script.py`, but pymod doesn't know which module to load until it has seen the arguments, therefore it can't present a "mod_name".

trws:
Thanks @grondo, that's pretty much exactly how it's set up.  It could use py_mod as its mod_name, but then as soon as it actually starts it would want to become "script_mod_name" or whatever.

garlick:
Spoke with @trws and @grondo and I think we agreed a base component that loads interpreted module extensions could do so hierarchically, e.g. pymod loads script.py then script.py (or pymod on its behalf) registers a service as described above.  If services can be registered at the top level, then services can be prototyped in an interpreted language and re-implemented as a C comms module without changing the service name.

One issue (noted by both @grondo and @trws and ignored my me until now :-)  is the C-centric module search won't really work for .lua or .py scripts.

@trws had a good idea:  add a `findmod` or similar method which allows a base component to provide an abstract search for extension modules it knows how to load.   We could have this method copy the module into the content store and return a reference.  `flux module load` would execute this on the local rank, then pass the reference to the insmod method on the target ranks.

@grondo wants to be sure we're not precluding other use of modules such as the lua plugins loaded by wrexecd.  I'll try to add language to make sure it is clear that RFC 5 applies only to module extensions that are sharing the protocol and tools developed for comms modules.

garlick:
Another topic that should be covered here is support for initialization and finalization synchronization, when an extension module is loaded or unloaded on multiple ranks at once.  For example, the KVS requires that (at least) rank 0 is loaded first so the initial root reference is created before upstream asks for it with a `getroot` RPC.  The live module does some reduction that requires its upstream to be loaded before downstream.

My thought was that extension modules could use a barrier over the number of ranks being loaded/unloaded concurrently.  The `insmod` message would need to grow a `count` parameter, which the base component would pass to the extension module's `mod_main()` as a standard argument like `--insmod-count=N`, which could then be used as a parameter to the optional barrier.

For finalization, the `rmmod` message would get a `count` parameter which the base component would pass to the extension module in the `shutdown` message.   In the shutdown case, the barrier comes at the end, before `mod_main()` exits so this implies a couple of things we don't have now:
* barrier interface with RPC continuations
* way to keep module status in FINALIZING with reactor active

Though really those are just implementation details.

garlick:
This PR has languished a while.  I was about to revisit RFC 5 to look at adding support for a better module option interface, and found it way out of date with respect to the code.  The changes proposed here would get it up to date, and are consistent with the way extension modules are now loaded by sched.

I'd like to propose we merge this, and defer the changes for interpreted modules under discussion (perhaps ask that @trws and @grondo update flux-framework/flux-core#663 or create a new issue as needed to cover the gaps).

trws:
I'm ok with that.

grondo:
Oh yeah, that's fine. Merging.

garlick:
Thanks!

PR #41, opened 2016-05-04
spec_4:  Disambiguate basename and name

lipari:
Implements the changes to the Flux Resource Model that were discussed
in: https://github.com/flux-framework/flux-sched/issues/157

### DIFF
--- a/spec_4.adoc
+++ b/spec_4.adoc
@@ -88,13 +88,25 @@ the nodes and cores as the finest resource granularity.
 
 * Type
 * UUID (Unique ID for this resource)
-* Name or Basename
+* Basename
+* Name
 * ID (optional numeric ID to be appended to basename to get name)
 * Properties (static properties associated with this instance)
 * Tags (dynamic list of tags)
 * Size (Total number of resources in this pool)
 * Allocation table (List of active allocations from this pool with metadata)
 * Hierarchy table (Hierarchies and topologies to which this resource belongs)
+
+The default value for `basename` is the `type`.  The default value for
+`name` is a concatenation of `basename` and `ID`, or just `basename`
+when the `ID` value is not assigned.
+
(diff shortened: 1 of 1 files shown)


grondo:
This looks fine to me. Anyone else want to comment? 

@trws did have the good idea to allow a resource `name` to be generated by an arbitrary parameter substitution/format specifier. If we state that idea in the RFC we could then say that by default the name format specifier is something like `"${basename}${id}` where `${id}` expands to empty string if not set, and that substitution SHALL support arbitrary resource attributes and properties (e.g. a strangely named system might have node "names" list `"node-${frameid}-${rack}"`)

lipari:
Just ran into a snag.  As proposed so far, if the basename is not specified, we assign it the type.  But what if the basename is not specified, but the name is?  A good example would be the cluster resource.  Its name might be cab.  Should the basename be "cluster" or "cab"?

grondo:
Perhaps resource `type` should be required? The type of a cluster resource should be "cluster", with name set to the cluster name. That way you could easily query for all "cluster" resources in a center, for example.

lipari:
@grondo, I agree with you w/r/t the type.  But I'm wondering what to set basename to.  IOW, can we support a lua config like this:
```
Hierarchy "default" {
    Resource{ "cluster", name = "cab",

```
where basename is not present?

grondo:
I think you already said basename defaults to the type, so in your example above, `basename = "cluster"`, though this would not really be used since `name` is already set to cab... The default of `basename = type` is really just a convenience for resource types like `socket` and `core` (?) Does that make sense?

lipari:
In that case, the latest push is ready for final review and merging.

grondo:
This seems like a good improvement over what was there! Merging

PR #42, opened 2016-06-06
RFC6: add matchtag group description, cleanup

garlick:
A few mods to RFC6 (Flux RPC protocol)
* Minor cleanup
* Allow an RPC to have zero or one response, rather than just one
* Describe "group RPC" and the allocation of the upper 12 matchtag bits as a group id, as implemented in flux-framework/flux-core#687

Left for another time:
* allowing for multiple responses per request
* more precise cancellation protocol (see flux-framework/flux-core#688)


grondo:
Nice clarifications. The wording here actually helped me a lot.

PR #45, opened 2016-06-20
rfc12: add security arch RFC

garlick:
As discussed in flux-framework/flux-core#699, this is a raw draft of a Flux security architecture document.

I think it accurately documents the single-user, non-bootstrap strategy we have now.  The intent here is to do that, so that we have a hopping off point for further design work - not to advance the design just yet.


spell.en.pws
```diff
@@ -147,3 +147,49 @@ Google
...
+Melnikov
+NaCl
+PMI
+SASL
+SELinux
+UID
+et
+al
+cgroups
+evesdropping
```
SteVwonder: This should be "eavesdropping"

garlick:
Oh thank you!

grondo:
This is really nicely done. I'm ready to merge if you are. 

I did happen to notice that `libk` snuck into the dictionary but doesn't seem to be necessary.

grondo:
> I did happen to notice that libk snuck into the dictionary but doesn't seem to be necessary.

oh, maybe because of `libk5crypto`?  Either way I don't think it matters too much.

garlick:
Thanks!  I don't think it hurts to merge this.  It's just a starting point.

garlick:
Let me squash this down to one commit though.

PR #46, opened 2016-06-21
rfc13: document PMI-1 subset of API and wire proto

garlick:
This pulls together some bits spelunked from the MPICH source code , its git history, SLURM source code, testing with mpi tracing a couple of years ago, and the pmi.h header file.


grondo:
This represents a huge amount of work, thanks! I glanced through it and it all looks sane.

PR #47, opened 2016-06-22
Further work on RFC 13

garlick:
This PR includes some cleanup and expansion of the PMI-1 RFC.

I tried to emphasize the concept of PMI as a tool for decoupling process managers and MPI runtimes, since it seems that this has been weakly achieved at best, despite a pretty good design IMHO.

I went out on a limb and suggested that a new environment variable PMI_LIBRARY be set to the path to a process manager's library, similer to Intel's I_MPI_PMI_LIBRARY environment variable.

### DIFF
--- a/spec_13.adoc
+++ b/spec_13.adoc
@@ -40,7 +40,7 @@ file descriptor to communicate with the resource manager.
 This specification describes a subset of the PMI-1 API and wire protocol,
 culled of:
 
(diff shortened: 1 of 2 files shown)


grondo:
Nice! I noticed you've found a nice standard way to cite references. Not important but should we document that somewhere?

PR #48, opened 2016-06-23
RFC 13

garlick:
OK, more cleanup in this PMI-1 RFC. 

This update tightens up the description of PMI 1.0 versus 1.1, which it turns out can be only a wire protocol change; the PMI API can be the same across both versions to keep things simple for implementations.

This adds an MPI Quirks section where I thought we could try to collect things like environment varaibles and build-time configuration runes necessary to get specific implementations to use PMI-1, preferably configured in a way that doesn't couple the MPI to a specific process manager.


garlick:
All done updating for now - please merge if it looks OK.  Thanks.

grondo:
Looks good after a quick read. Lots of great info here, thanks!

PR #49, opened 2016-06-25
rfc13: express wire proto in ABNF

garlick:
This pr converts the kind of lame wire proto doc to ABNF, and changed a bit after cross-checking with the MPICH sources.


grondo:
Is this one ready?

garlick:
OK.

PR #51, opened 2016-07-11
rfc14: add skeletal canonical job spec

garlick:
As discussed in pr #29 and in our meeting today, here's a skeletal jobspec RFC redone in asciidoc, with a new number assigned.  I made @trws the editor.

I added an overview section which might be a little bit off the main focus of this document.  Happy to iterate to get it right.  Same with the goals section (I undoubtedly missed some or added some non-goals so please scrutinize).

The only bit I brought over from pr #29 was the terminology.  That wasn't due to a conscious exclusion of the other material - just trying to stick to the plan of submitting a template plus intro and leaving the hard work to other pr's.

### DIFF
--- a/spell.en.pws
+++ b/spell.en.pws
@@ -272,3 +272,15 @@ nnodes
 ppn
 tuples
 interoperability
+jobspecs
+jobspec
+workflow
+acyclic
+reusability
+hostlist
+expressivity
+enqueued
+Ingerson
+Kiki
+YAML
+mpirun
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,8 @@ HTML = \
 	spec_10.html \
 	spec_11.html \
 	spec_12.html \
-	spec_13.html
+	spec_13.html \
+	spec_14.html
 
 all: $(HTML)
 
--- a/README.adoc
+++ b/README.adoc
@@ -78,6 +78,13 @@ standardized by the MPI Forum and has been only lightly documented.
 This RFC is an attempt to document PMI-1 to guide developers of resource
 managers that must support current and legacy MPI implementations.
 
+link:spec_14{outfilesuffix}[14/Canonical Job Specification]::
+A domain specific language based on YAML is defined to express the
+resource requirements and other attributes of one or more programs
+submitted to a Flux instance for execution.  This RFC describes the
+canonical form of the jobspec language, which represents a request to
+run exactly one program.
+
 == Change Process
 
 The change process is
--- a/spec_14.adoc
+++ b/spec_14.adoc
@@ -1,0 +1,109 @@
+ifdef::env-github[:outfilesuffix: .adoc]
+
+14/Canonical Job Specification
+==============================
+
+A domain specific language based on YAML is defined to express the
+resource requirements and other attributes of one or more programs
+submitted to a Flux instance for execution.  This RFC describes the
+canonical jobspec form, which represents a request to run exactly
+one program.
+
+
+* Name: github.com/flux-framework/rfc/spec_14.adoc
+* Editor: Tom Scogland <scogland1@llnl.gov>
+* State: raw
+
+== Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to
+be interpreted as described in http://tools.ietf.org/html/rfc2119[RFC 2119].
+
+== Related Standards
+
+* link:spec_4{outfilesuffix}[4/Flux Resource Model]
+* link:spec_12{outfilesuffix}[12/Flux Task and Program Execution Services]
+
+== Goals
+
+* Express the resource requirements of a program to the scheduler.
+* Allow graph-oriented resource requirements to be expressed.
+* Express program attributes such as arguments, run time, and
+task layout, to be considered by the program execution service (RFC 12)
+* Express dependencies relative to other programs executing within
+the same Flux instance.
+* Emphasize expressivity over simplicity, as this canonical form
+may be generated from other user-friendly forms or interfaces.
+* Facilitate reproducible runs.
+* Promote sharing and reuse of jobspec.
+
+== Overview
+
+This RFC describes the canonical form of "jobspec", a domain specific
+language based on YAML[1].  The canonical jobspec SHALL consist of
+a single YAML document representing a reusable request to run
+exactly one program.  Hereafter, "jobspec" refers to the canonical
+form, and "non-canonical jobspec" refers to the non-canonical form.
+
+Non-canonical jobspec SHALL be decomposed into jobspec before
+it is enqueued for the scheduler and program execution service.
+
+User facing tools MAY generate jobspec from non-canonical jobspec,
+or other sources.  Such tools MAY:
+
+* generate a batch of dependent jobspecs representing a scientific workflow
+* generate a stream of jobspecs representing a steered parameter study
+* convert simulation parameters into jobspec containing computed
+resource requirements, etc.
+* convert command line arguments to jobspec, e.g. "flux mpirun"
+
+=== Jobspec and Program Life Cycle
+
+The jobspec SHALL be submitted to a job submission service.  Malformed
+jobspec SHALL be immediately rejected by the job submission service.
+A stack of plugins SHALL test jobspec against site or user defined
+criteria, and on failure, MAY reject the jobspec, or MAY warn the user
+and continue on.  The job submission service SHALL enqueue the jobspec
+for consideration by the scheduler.
+
+The scheduler SHALL consider each enqueued jobspec in the context of its
+dependencies and the pool of available resources.  When the scheduler
+chooses to execute a job, it allocates resources, associates them
+with the jobspec, and notifies the program execution service to start
+the program(s).
+
+The program execution service, described in RFC 12, launches the program(s).
+Task slots, containment, and task layout SHALL be created within the
+allocated resources as described by the jobspec, or if that is not
+possible, the job SHALL enter a failed state and resources SHALL
+be returned to the scheduler.
+
+Once a job is retired, the jobspec SHALL be retained as part of
+its provenance record.
+
+=== Resource Matching
+
+Resources are represented as hierarchies or graphs, as described in RFC 4.
+
+FIXME: describe how Flux hierarchical resource representation affects
+jobspec design.
+
+=== Terminology
+
+FIXME: Fill in
+
+== Jobspec Language Definition
+
+FIXME: Fill in
+
+== Basic Use Cases
+
+To implement basic resource manager functionality, the following use
+cases SHALL be supported by the jobspec:
+
+FIXME: Fill in
+
+== References
+
+* [1]http://yaml.org/spec/current.html#representation[YAML Ain't Markup Language (YAML™) Version 1.1], O. Ben-Kiki, C. Evans, B. Ingerson, 2004.


grondo:
Thanks @garlick! This is extremely helpful as we can get this initial draft merged and then iterate on the final details at leisure. Here's some points that jumped out for discussion on my first perusal. These aren't necessarily at the level of full comments, but just some points we might want to explore before merging this initial document:

 * I'm not sure about their formal definitions, but the term "specification language" jumped out at me as possibly redundant or perhaps even incorrect? Do you think it is descriptive enough (and correct) to call the thing described in the rfc as just the "job specification" (jobspec), and drop the "language" modifier? i.e. instead of `"A formal job specification language is defined..."` just `"A formal job specification is defined..."`

 * Original doc mentioned a formal jobspec is "logically a list of programs" -- do we need to call that out in the summary and overview, i.e. instead of "... attributes of a program ...", "... attributes of a program or set of programs ...". It did seem like a good idea that a single jobspec could describe a set of resources on which multiple programs may be run (eventually parametric/array job syntax could be supported for instance) It occurs to me that I may be mixing up my program/job terminology here though, so please correct me if a program can be defined as a set of programs.

 * In Goals, my inclination would be to emphasize the resource shape features of jobspec before the program runtime requirements etc. Though, in the big picture program options are of a similar importance to the "resource request" syntax, it feels like the RFC itself will focus more on resource request/specification features of the jobspec. Perhaps move 3rd bullet to second position, place program options like runtime, env, etc all under one subbullet (?) Am I even making any sense?

 * Probably not in this PR, but later at least I'd like to see a quick description or summary of how resources fit the graph model before launching into the Glossary which uses the term "Resource Vertex" somewhat out of context (for me)

garlick:
> the term "specification language" jumped out at me as possibly redundant or perhaps even incorrect

Well, we are talking about a "specification specification" so I don't think we can just drop "language" without replacing it with something.  I'd like to find a way to drop "formal" also, if possible.  Hmm.  Not coming up with any great ideas at the moment, although it does seem like we can make that phrase a little more concrete by adding "YAML-based" if we are agreed that YAML will be the base language.

> It did seem like a good idea that a single jobspec could describe a set of resources on which multiple programs may be run

RFC 8 says:
> **job**  A synonym for program.

although I guess that doesn't necessarily mean one job to a jobspec.

Hmm.  It seems like we want to be careful not to inadvertently create a dead zone like slurm's "scheduling within an allocation" here.  If I want to run 100 instances of the same job in a parameter  study, or a DAG of jobs in a workflow, I had assumed that some other tool would create a batch of jobspecs and they would somehow reference each other, for example as dependencies.

However it seems like there might be cases where programs must be more tightly coupled, like program + tool, or programs related within an MPI comm world?

>  emphasize the resource shape features of jobspec before the program runtime requirements etc.

OK

>   I'd like to see a quick description or summary of how resources fit the graph model before launching into the Glossary

Yeah I was thinking the same thing but I'm hoping someone else can provide it, as this is where I tend to get lost.

garlick:
Would it be accurate to say the jobspec is domain specific language, defined as a dialect of YAML?

https://en.wikipedia.org/wiki/Dialect_(computing)

grondo:
> Well, we are talking about a "specification specification" so I don't think we can just drop "language" without replacing it with something. 

Ok, though it sounds a bit strange to me, I'm fine with with "specification language" -- searching around YAML is defined as a language enough to where I think I was probably wrong to call this one out..Sorry

grondo:
> Would it be accurate to say the jobspec is domain specific language, defined as a dialect of YAML?

Yeah, that sounds fine to me. Sorry for the confusion.

garlick:
Couple comments on the way out the door after a brief conversation with @grondo
* Maybe only programs that run concurrently should be specified within the same jobspec?
* Jobspec should be composable, e.g. one spec for database, one for MPI program, another jobspec to compose the two so they run together?

It seems like composability should be a goal.  YAML doesn't directly support includes so we will have to add this on some way.  [here's a discussion on the topic](
http://stackoverflow.com/questions/528281/how-can-i-include-an-yaml-file-inside-another)

spell.en.pws
```diff
@@ -272,3 +272,12 @@ nnodes
 ppn
 tuples
 interoperability
+jobspec
+reusability
+reproduceable
```
SteVwonder: This should be 'reproducible'

SteVwonder:
@garlick, this looks great!

My only suggestion is to reorder the terms under 'Terminology'.  It seems weird to start with 'Group' since it references 'Resource'.  I think forward references should be avoided when possible.

spell.en.pws
garlick: Hah that's twice you've caught me being lazy.  I should be embarassed 😂

SteVwonder:
In #29, it looks like we settled on using 'slot' rather than 'shard'?  It seems that 2 usages of 'shard' creeped in (lines 129 & 139).

garlick:
It seems convenient to define a job as program or set of concurrently executed, cooperating programs.  Is that reasonable?  That would be an update to RFC 8.

grondo:
> It seems convenient to define a job as program or set of concurrently executed, cooperating programs. Is that reasonable? That would be an update to RFC 8.

That seems like a clever redefinition of job. It somewhat solves the question of "what is the resource requirement for a job?" Though, there seems to still be a couple tricky cases like, in practical terms, what we will call the Flux equivalent of the "script o' sruns" batch script submission most people use?

I think it will be natural for people to call it a "job", and I suppose that's technically correct since the Flux instance is a program/job that runs a set of programs. Is that too confusing?

Another minor question, do the programs within a job have to all start and end at the same time? I'm thinking of the multi-level scheduling example where a parent instance and one or more levels of child instances are all cooperating, is the whole thing technically a job? I guess that might be splitting hairs...

garlick:
> what we will call the Flux equivalent of the "script o' sruns" batch script submission most people use?

It seems like another case of a tool that generates a batch o' jobspecs.  Instead of
```
$ sbatch runscript.sh
```
we would have
```
$ flux submit flux start runscript.sh
```
or possibly the flux start could be dropped in favor of a flux submit option?

grondo:
> It seems like another case of a tool that generates a batch o' jobspecs. Instead of

Sorry, I wasn't being clear. I'm still focused on terminology and traditionally this batch of programs would be called a "job". Howerver, technically in our case it is now a program or job (the flux instance launched by `flux start`) that runs a set of other jobs. I'm worried this will be confusing when communicating with users ("which job in your job had a problem?")

I could be making something out of nothing.

garlick:
I guess the jobid would be how the "which job" question is answered?   Uh oh.

If "job" is one or more programs, and we don't require those concurrent programs to be launched in their own instance, then we don't have any special identifier to track the "job" as opposed to the programs contained in it.

Maybe redefining "job" is a bad idea after all, and instead we should say one jobspec == one job == one program, and then figure out what dependencies look like between jobspecs, and handle cooperating, concurrently executing programs as a type of dependency?

Actually I guess I'm on a (related) tangent and not specifically addressing your question.  More thought required.

grondo:
> Maybe redefining "job" is a bad idea after all, and instead we should say one jobspec == one job == one program, and then figure out what dependencies look like between jobspecs, and handle cooperating, concurrently executing programs as a type of dependency?

Yeah, it seemed @trws had a very nice plan for dependencies with Makefile like rules. If a jobspec can only define one job then the analogy here is a jobspec is a source file, running and/or completed programs are the targets. Perhaps rules could define "targets" that are built from multiple job specs? Or I think Tom's idea was a system where a program can emit a named target, and other rules can require that target as a prereq? If we allowed appending a state to the generated target name, perhaps that would allow running cooperating jobs loosely in parallel, e.g.

```make
preprocess: preprocessor.yml
    flux submit -f preprocessor.yml  # Or this is the implied rule, or should that be flux run?

monitor: monitor.yml preprocess.starting # Start monitor once preprocess is starting
job: job.yml preprocess monitor.running  # Start job once preprocess completes and monitor is running 
```

Sorry to get off in the weeds here, but it makes me a bit uncomfortable to split the dependencies from the jobspec, so I'd like to understand how that is going to work before committing to "one jobspec, one program"

What makes me really uncomfortable about the above is how a set of cooperating jobspecs that are required to be run at the same time (multi-physics example, distributed app example) will be presented to the scheduler so the scheduler can collapse/union their resource requirements and schedule as a unit.

trws:
Ok, I'm clearly coming to this late (apologies, my parents got into town last night so I'm just getting back to email now).  It looks to me like a few different things are getting conflated here, so let me lay out how I conceptualized the parts of this that seem to be under discussion and we can go from there.

A job-spec specifies one or more programs as a YAML *stream*.  Any given YAML document in that stream contains a set of one or more potentially related programs, each program is a job, each program contains one or more tasks.   A program must schedule, and start, all of its tasks simultaneously under the same walltime, etc.  Between programs, dependencies can be used to enforce constraints like "program A must not start until program B is complete" or "program C must not start until Program B has notified the system that it has satisfied an internal dependency".  In other words, program == job.  The reason I like the ability to have multiple programs in a single spec, is the ability to do things like this:

```yaml
programs:
  - count: 500
    task:
      command: do-something-parametrized-by-jobid-from-env
    resources: Node
```

It also means you can use YAML backreferences to build a set of programs that are different but all need similar resources within yaml's constraints.  The reason I say it's a YAML stream, rather than a YAML document, is that if a tool is to generate a number of jobs, it is likely easier for it to generate a set of documents as a stream rather than a single combined document.  That would look like this:

```yaml
---
# program 1
task: do-thing-1
resources: Node
---
# program 2
depends:
  - id: mydep2
    direction: inout
task: do-thing-2
resources:
  - type: Node
    with:
      type: Core
      count: 17
---
# program 3
depends: mydep2 #implicit inout
task: do-thing-3
resources: Core
---
```


The original writeup had an additional task alias called "instance" that allowed nesting an instance and initial program/spec to be provided to that instance on startup.  This is not in the current version, but would not necessarily be a bad place to go.  Some of the examples that have come up, where you want to run two programs together and for some reason don't want them to be tasks under a common program but rather programs under an instance might look like this:

```yaml
resources:
  type: Node
  count: 75
instance:
  spec:
    programs:
      - task: my-db
        resources:
          type: Node
          count: 15
        depends:
          id: my-db-ready
          type: runtime
          direction:out
      - task: my-job-that-needs-a-db
        resources:
          type: Node
          count: 60
        depends:
          id: my-db-ready
          direction: in
```

Probably don't want to go there right now, but that was the idea for dealing with some of these conceptual issues.  Thoughts?

SteVwonder:
@trws: can you verify that I have interpreted the dependency syntax correctly?

Depends 'inout' means that all tasks with this dependency must be started simultaneously
Depends 'in' means that all tasks with this dependency must be started after the task that "provides" the dependency (has the 'out' depend) has completed.
Type 'runtime' on an 'out' dependency means that it is satisfied after the tasks starts.

At the risk of getting into the weeds (please feel free to say if so) my question is what to do in the case where the user specifies multiple tasks with the same dependency?  Do tasks with an 'in' dependency start after 1 task with the 'out' dependency starts, does there need to be a 1to1 mapping of ins to outs, or do all 'outs' need to start before any of the 'ins' can start?

grondo:
Thanks @trws,

Describing a jobspec as a stream of programs is helpful! That is something that should be emphasized in the RFC -- as it is I had missed it.

To our tangential discussion earlier in this PR, dependencies would be resolved local to the jobspec correct? Is there any facility to resolve dependencies between jobspecs?

@garlick had expressed interest in getting this skeleton PR merged immediately before we put too much more discussion here, and that seems like a good idea.

I'd just like to take one more pass on the Glossary, but any objections to merging the skeleton as is (maybe we remove any description of jobspec == one job from the document too), after any other comments directly related to the content already in the document as presented in this PR?

garlick:
Yeah let me make a pass trying to not say anything incompatible with @trws's conception.  Thanks Tom that was fairly clear and makes sense I think.

trws:
Generic thing, the dependency IDs I think should be global to a user by ID rather than just in a jobspec.  We could adjust this, but it could complicate matters for users to submit things individually on the command line.

@SteVwonder:
* in: do not start this program until all programs that specify "out" with this ID are done
* out: notify flux on completion of this program to schedule those programs that have an input dependency on this ID
* inout: both in and out, no two programs that specify the same ID inout can run concurrently
* runtime: allow the application to satisfy this dependency explicitly, is not satisfied until that happens, is set to "failed" or similar if it is not set before the application finishes.

Dependencies are program-only, if we try and do it per-task we'll go nuts I think. =)

@grondo, I think dependencies should be resolved globally, but with easier usage within a single jobspec.  A user depending on their own job would just specify an id, to depend on someone else's job it would be something like 'user/id' or similar, I haven't thought that through anywhere near as far but it simplifies life to do it that way.  This all means that the order the jobs are specified in, both in a jobspec and between them, is relevant for the order in which they are run.  Having that be implicit makes it much easier to specify a dag, but means there's a little more work in the implementation (not much).  It ends up working as though each id is a FIFO queue of programs to run, so that kvs queue-type-thing that we talked about would make this rather easier to do as a sidenote.

I'm good with merging this basically as is as long as we clean up my mistaken use of shard and the couple little bits that have already been identified, we can build up the rest subsequently.

grondo:
> Generic thing, the dependency IDs I think should be global to a user by ID rather than just in a jobspec.

> @grondo, I think dependencies should be resolved globally, but with easier usage within a single jobspec. 

I'm assuming by "easier usage within a single jobspec" you mean a user or tool generating a jobspec can use symbolic identifiers instead of global IDs, correct? This means that something will have to come along after jobspec submission and change symbolic IDs to global IDs, or append/prepend a ID to make them so.

Without this there would be no way to submit the same jobspec more than once at the same time, or even guarantee no ID collisions within the user namespace.

trws:
Right, that's exactly it.  Probably a different attribute name for whether you want a local or global ID or something like that, but there should be a way to make it scoped, preferably other than using an instance (which would certainly do it, but has other consequences).

garlick:
Once a jobspec is submitted, does it degenerate into its component programs in the queue, or is it tracked as a whole?

It seems like when multiple programs are involved, an instance is a convenient way to manage them as a unit.   It scales better that way too, if multiple complex jobspecs are submitted to a shared instance.

trws:
It decomposes, except for those programs that are "hidden" for lack of a better word inside an instance.


garlick:
Does it decompose to jobspec or something else?

trws:
Jobspec with a single program. In the end, a canonical job spec is a single program, we could even restrict it to that for right now if you want. The multiple program thing is a specification convenience.


garlick:
If all jobspec has to be decomposable to single programs, should we define "canonical" jobspec as the decomposed form?  That would simplify the task at hand.

trws:
That makes perfect sense to me if it works for others.


garlick:
I headed down that path a bit but can always pop that last commit off if it's not agreeable.

I'm wondering if the terminology should be removed from this PR (in the interest of getting something in and not having the discussion go off in too many directions before we get there)

trws:
I like the updates @garlick.  If you think it makes more sense to remove 
the other terminology for now, go for it.  I'll do a rework against 
what's accepted here as soon as possible after that goes in.

grondo:
I too like the direction this is going... sorry I was very little help :-(

garlick:
If we're agreed on what's here (not a lot really!) then I will squash down to one commit.

May I suggest for the next PR we get the slurm use cases documented with canonical jobspec examples?  That shouldn't be too hard to agree on :-)

garlick:
Right!   Squashed.

grondo:
Ok, this looks good to me. The one thing that gives me pause is:

> Non-canonical jobspec SHALL be decomposed into jobspec before it is submitted to a Flux instance.

This seems restrictive to me as it does not allow *any* non-canonical jobspec to be submitted directly to some service within an instance (not necessarily the scheduler). If you already discussed this with @trws I'm fine with keeping it for now.

Just give me yea or nay and then I'll merge this so we can move forward

trws:
It depends on how you look at it.  I'm expecting a service or library in 
flux to produce the generic jobspec to produce the canonical form, if 
you think that precludes this usage, it would be good to loosen that, 
but I could see it going either way based on interpretation.

garlick:
Maybe if I said "Non-canonical jobspec SHALL be decomposed into jobspec before it is enqueued for the scheduler and program execution serivce"?   That at least leaves the door open for a service acting as a filter converting non-canonical to canonical...

trws:
Works for me.

grondo:
Great, I'm ready to merge this -- if you want to squash your last commit go ahead, or it can stand as far as I'm concerned.

garlick:
OK squashed.

grondo:
Done! Nice work @garlick!

PR #52, opened 2016-07-13
RFC 14: Begin adding jobspec use cases

grondo:
I've started adding to the *Use Cases* Section of RFC 14. So far only the most basic use cases are added, not yet including all the examples from @trws's original document.

I'm still going to add more use cases from the more esoteric Slurm `-B` or `--ntasks-per-*` command lines. Then we have to explore the `--gres` (generic resources), `--multi-prog` examples, and finally at some point `--array` (job arrays)

However, I wanted to get this PR up for discussion, and because I did run into a couple places where I wasn't sure how the jobspec should exactly be specified (See UC4 and UC6).


spec_14.adoc
```diff
@@ -102,7 +102,167 @@ FIXME: Fill in
...
+Use Case 4:: Run N tasks, Require M cores per task
++
+Specific Example:: Run 10 copies of `myapp`, require 2 cores per copy,
+for a total of 20 cores
++
+Existing Equivalent:: `srun -n10 -c 2 myapp`
++
+[source,yaml]
+# FIXME: VALIDATE
+type: group
```
trws: Using a group here is valid, but not required:

```yaml
tasks:
  command: myapp
resources:
  type: core
  count: 2
```

spec_14.adoc
```diff
@@ -102,7 +102,167 @@ FIXME: Fill in
...
++
+Specific Example:: Request 4 nodes with at least 2 sockets each,
+and 4 cores per socket
++
+Existing Equivalent::
+`srun -N4 --sockets-per-node=2 --cores-per-socket=4` or
+`srun -N4 -B '2:4:*'`
++
+[source,yaml]
+# FIXME: VALIDATE
```
trws: Looks exactly right to me.

spec_14.adoc
```diff
@@ -102,7 +102,167 @@ FIXME: Fill in
...
+        count: 2
+        with:
+            type: core
+            count: 4
+
+'''
+Use Case 7:: Run different binaries with differing resource
+requirements as single program
++
+Specific Example:: 11 tasks, one node, first 10 using one core and 4G of RAM for
```
trws: Is this example meant to run 10 'read-db' processes overscheduled on the same core, or one each on its own core? Currently it's correct, but will run all 10 in the same single container of a single core with 4gb of ram total.

trws:
Thanks @grondo! This looks like a really good start.  It would be good to work over some of the -B and --gres options, and also some of the placement options, since I'm not sure I have a handle on how those fit in yet.  The multi-prog options, unless that translates to multi-task in our nomenclature,  should probably wait for speccing the non-canonical spec though, as they'll be pre-canonicalization by definition.

spec_14.adoc
grondo: This one is cut-and-paste from your original RFC. I'm guessing what we'd want is the first 10 (`read-db` tasks) to be in a container with 10 cores and 4G of RAM, and the final task (single `db` task) to be in a container with 6 cores and 24G of RAM.

trws: Huh, amazing how one misses these things in their original context.  Yeah, it does what's expected there, guess I was just trying to figure out if anyone would ever do that...  It's a small twist to make it that each read-db gets its own container of that size, but maybe it's good to have the oversubscription in there.

spec_14.adoc
grondo: So:
```yaml
type: group
tasks:
    command: myapp
with:
    type: core
    count: 2
count: 10
```

That seems to make sense, though I'm getting confused about the `resources:` tag. When is it required? Only when the tasks are separated outside of the resources: block, otherwise it is implied?

Also, is `group:` strictly required here?

garlick:
Oh you got yaml pretty printing to work in asciidoc.  It didn't work on my desktop for html so I assumed incorrectly that it wouldn't render on github either.  Cool!

spec_14.adoc
grondo: I'm getting myself confused on this one. Should the `read-db` section be reorganized to be
```yaml
tasks:
    command: read-db
    with:
        type: core
     count: 10
     with:
        type: memory
        amount: 4g
          
```

To run 10 copies of `read-db` across 10 cores with 4G *total* RAM?

spec_14.adoc
trws: My current "schema" as it were, would pedantically write that as:

```yaml
#this is the "program" level
walltime: <default>
... other program stuff
resources:
  type: group
  tasks:
      command: myapp
  with:
      type: core
      count: 2
  count: 10
```

The group is necessary to replicate both the tasks and the cores together.

I hadn't really thought about this, but we could make "task" a resource type with some special keys and solve that restriction.  That would make it nest more easily with other resources, but might have some other ramifications that aren't occuring to me.

Could look something like this:

```yaml
resources:
  type: task
  command: myapp
  count: 10
  with:
    type: core
    count: 2
```

Thoughts?

grondo:
> The multi-prog options, unless that translates to multi-task in our nomenclature

Yeah, I think Slurm multi-prog is simply a way to run "a different executable per rank" in a single job step, much like our `read-db` `db` example.

Note I'm still confused on how exactly to specify the grouping in a jobspec. An example would be run an executable across N cores where each copy requires 1G of RAM (container is 1 core, 1G RAM for each task), and run N copies of an executable, where each executable requires 1 core, but run with  4GB *total* RAM. (Just an example, same goes for cores, or other resources). I hope I'm being clear.

spec_14.adoc
grondo: >  It's a small twist to make it that each read-db gets its own container of that size, but maybe it's good to have the oversubscription in there.

I'd say for this example, we leave out the oversubscription, and add a specific use case for oversubscription. My feeling is this is a good example of non-oversubscribed multiple executable program, and that the oversubscription is too much for this example.

trws:
I haven't tried many of those examples, but for something that will get 4GB total ram with one task per core, 10 cores, inside the same container, I would expect this to work:

```yaml
resources:
  type: Node
  with:
    - type: Memory
      amount: 4GB
    - type: core
      tasks: 
        command: app
```

I admit this feels a bit odd, because the memory is actually on a branch rather than a direct parent, but this will only require that there be at least 4gb total ram for all tasks combined.  If I went for the more general case craziness, it would probably have a link from each task back to the same memory block as a requirement, but I see no reason to require that really.

spec_14.adoc
trws: Ok, so it would be:

```yaml
resources:
    - type: node
      with:
        - type: group # Note, special resource type "group" to add a level
	  count: 10
          tasks: 
              command: read-db
          with: 
            - core
            - type: Memory
              amount: 4GB
        - type: group
          tasks: 
              command: db
          with:
            - type: Core
              count: 6
            - type: Memory
              amount: 24GB
```

Or with tasks as resources, which I'm finding I kinda like:

```yaml
resources:
    - type: node
      with:
        - type: task # Note, special resource type "task" to add a level
	  count: 10
	  command: read-db
          with: 
            - core
            - type: Memory
              amount: 4GB
        - type: task
	  command: db
          with:
            - type: Core
              count: 6
            - type: Memory
              amount: 24GB
```

grondo:
Thanks. To me that does look like a requirement for 4GB per node. Is there a reason we couldn't leave out the Node parent? Also, it seems ambiguous now if I add `count: 10` whether this means the jobspec is requesting 4GB or 40GB. I'm probably missing something simple here.

trws:
The node parent isn't actually required.  Updated, and added count.

trws:
As things are now, this would do the same thing:

```yaml
tasks: 
  command: app
resources:
  - type: Memory
    amount: 4GB
  - type: core
    count: 10
```

spell.en.pws
```diff
@@ -284,3 +284,20 @@ Ingerson
...
+Infiniband
+bicore
+cpu
+hostname
+ib
+memnode
+myapp
+nbcore
+nnodes
+nodoes
```
SteVwonder: Should just be nodes.

Line 120:
> Specific Example:: Run `hostname` 20 times on 4 nodoes, 5 per node

spec_14.adoc
SteVwonder: @trws: for your examples, are the indentations correct?  Inthe second example, I would have assumed that the count and commands would have been under the task type, like so:

```yaml
resources:
  - type: node
    with:
      - type: task # Note, special resource type "task" to add a level
        count: 10
        command: read-db
        with:
          - core
          - type: Memory
            amount: 4GB
      - type: task
        command: db
        with:
          - type: Core
            count: 6
          - type: Memory
            amount: 24GB
```

trws: Somehow I ended up with tab characters in there, that didn't behave well.  Really not sure how that happened, but thanks @stevwonder that was definitely messy.

spell.en.pws
grondo: Heh, thanks!

spec_14.adoc
grondo: As much as my mind rebelled against it at first, I agree tasks as resources is actually a bit more intuitive. This might show how little I understand the current jobspec and yaml, but is it possible to have `group:` and maybe `task:` be synonyms for `type: group` and `type: task`? Actually perhaps that doesn't matter much for the canonical jobspec we're discussing here.

grondo:
>As things are now, this would do the same thing:
> ```yaml
tasks: 
  command: app
resources:
  - type: Memory
    amount: 4GB
  - type: core
    count: 10
```

Hey that is pretty straightforward!

Just to make sure I'm getting it, if the memory were a child of `type: core

File: spec_14.adoc (hunk 3 above)
Comment (trws) [inline spec_14.adoc]: Yeah, at some level they really are either resources or properties of 
resources anyway, so it kinda works.  I like the idea of having an easy 
way to specify it like that, but am trying to think of how that would 
look.  At some point I moved from tasks as resources to task as a 
property, and it's messing with me a bit.  We could say that it turns 
into this (so that it wouldn't require a with then a key with no value):

```yaml
with:
   type: task
```

Then it could be written as:
```yaml
resources:
     - type: node
       task:
         - count: 10
           command: read-db
           with:
             - core
             - type: Memory
               amount: 4GB
         - command: db
           with:
             - type: Core
               count: 6
             - type: Memory
               amount: 24GB
```

I'm no longer sure why I changed it into its own thing that was outside 
the
graph in some ways, but this is feeling better than having the group 
type for
this purpose.  If we have this I don't see that we necessarily even need 
group
anymore.

grondo:
Ok, one more case that is tripping me up (I feel like I'm failing an IQ test here).

Same as the above simple example:
```yaml
tasks: 
  command: app
resources:
  - type: Memory
    amount: 4GB
  - type: core
    count: 10
```

However, now we want to run 10 copies of app across exactly 2 nodes, with at least 4GB per node. The Slurm equivalent is simply
```
srun -N2 -n10 --mem=4096 app
```

I guess the straightforward jobspec would put tasks at the lowest level and require the user to specify tasks per node, e.g.

```yaml
resources:
    - type: Node
    count: 2
    with:
         - type: Memory
         amount: 4GB
         task:
             command: app
             count: 5
```

But now say the user wants to run the 10 tasks across 10 nodes. They would have to update both the `task:` count and the `type: node` count, whereas in Slurm they are used to just doing:
```
srun -N10 -n10 --mem=4096 app
```
Is there some way to specify the tasks at the top with a total count, but require a given shape to each task slot (I suppose in this case a core on a node, where the node has at least 4g)? Now that I wrote that down it kind of seems like this is what "edges out" are for?

grondo:
Oh, duh, or will this work?
```yaml
tasks:
    command: app
    count: 10
resources:
    - type: node
    count: 2
    with:
        type: memory
        amount: 4G
```

trws:
> Hey that is pretty straightforward!
> 
> Just to make sure I'm getting it, if the memory were a child of type: core, then each core and thus each task would require 4GB (e.g. 4g+core would be the task slot, and thus the container)?

That ends up being a kinda odd edge case, because memory is logically at the node level so you're referencing something higher in the hardware tree.  It also means the bottom resource, the memory, is not an executable resource, it can't actually run the tasks, so yes the core would still be the selected task-slot level, and the each task would get one core and its own 4gb of memory.

The ordering in the hierarchy is part of why the cores with memory examples tend to be on a level with one another.  If we treat the tasks as resources, it could also be like this:

```yaml
resources:
  task:
    with:
      - type: memory
        amount: 4gb
      - core        
```

trws:
> Oh, duh, or will this work?
> 
> tasks:
>     command: app
>     count: 10
> resources:
>     - type: node
>     count: 2
>     with:
>         type: memory
>         amount: 4G

That should work.  Also, this is one of the reasons it may be useful to be able to mark where tasks are executed without including them in the hierarchy so users can decouple the task count from the expansion of the tree if they need to.  That's something I don't have syntax nailed down for, but we talked about using yaml references or IDs or something for it.

grondo:
> memory is logically at the node level so you're referencing something higher in the hardware tree

Ah, ok -- starting to make sense. Is `- core` just expanded to `type: core` or was that a typo.

If memory is specified as a child of a core, will that be a fatal error returned by the system?

trws:
Sorry, I was using some of my old syntax without thinking about it.  Just specifying a string where a resource is expected, instead of a dict, was interpreted as 'type: <string>' in my parser before.

I would say yes.  There is actually a way to say you need a core with a certain amount of memory, but it would have to use an inward edge rather than an outward one to find it.  If that's unpalatable, we could say that with means in either direction, but it would make the matching logic slower.

grondo:
Great thanks!

Trying to keep up with comments I've pushed some commits that make fixes I think we've all agreed upon. I'm trying to keep it to one commit per comment, and referencing the commenter in the commit message for now.

grondo:
Ok, another couple updates pushed
 * I added a couple example use cases from the OAR documentation
 * I expanded some "Existing Equivalent" examples from `srun` only based examples to include `qsub` or `oarsub` equivalents. I haven't used PBS or OAR, so there is some risk I got the examples wrong.

@garlick was also suggesting `mpirun` examples but since `mpirun` is a launcher only I couldn't really map its usage onto most of these cases (standard MPICH versions seem to basically support `-np` and `-nodes`) So if someone can submit mpirun use cases we should tackle, post them here and I can include them.

garlick:
Yeah probably ignore mpirun for now, since we're mainly sorting out the resource requests.  I think not being too slurm centric is good here though.

trws:
@grondo, the node/core requests don't really translate, but I would pull in the mpirun.hydra examples when we start talking about actually mapping tasks to specific resource instances (probably not right now), but you can see some good examples [here](https://wiki.mpich.org/mpich/index.php/Using_the_Hydra_Process_Manager) of what the `-bind-to` argument can do, which is a pretty general process->something mapping system.

lipari:
I believe there is a typo in Use Case 9.  Shouldn't the memory size be 2GB instead of 4GB?  And if we are to assume that the core count of 10 leads to 10 tasks, how would one specify each task getting 2GB vs all 10 tasks sharing 2GB?

trws:
Yeah, that description would be:

```yaml
tasks:
  command: app
resources:
  type: group
  count: 10
  with:
    - type: memory
      amount: 2GB
    - type: core
```

The version above gives each task 2gb, the below constrains all 10 tasks to
use a single block of 2gb total:

```yaml
tasks:
  command: app
resources:
  - type: memory
    amount: 2GB
  - type: core
    count: 10
```

trws:
I should perhaps say that it requires that there be at least 2gb available for each task, I'm not sure if that would constrain the task to fit in that amount of memory or not, but we probably want it to.

grondo:
> The version above gives each task 2gb, the below constrains all 10 tasks to
use a single block of 2gb total:
>```yaml
tasks:
  command: app
resources:
  - type: memory
    amount: 2GB
  - type: core
    count: 10
```

Ah, Ok, I had thought this created 10 task slots each with memory and a core, since `memory` and `core` were at the same level. Your use of group above does make sense though. I'll try to fix up the Use Case 9. Thanks @trws and @lipari.

Side note, since memory is higher in hierarchy than cores, could you specify this as

```yaml
resources:
    - type: memory
      amount: 2GB
      with:
          type: core
```

Or is that not correct because typically a `core` is not a direct descendant of `memory`?

garlick:
Sorry for the lame question but should "resources" be optional in the canonical form?  Does that kind thing (implied options) weaken validation?

grondo:
Also, this might be a case where using `task:` instead of `group` makes things more intuitive, though my brain still gets queasy looking at `resources: task:`...

```yaml
resources:
    task:
        command: app
        count: 10
        with:
            - type: memory
            amount: 2G
            - type: core
```

trws:
> since memory and core were at the same level

They are, but they're actually separate objects in a list together, so the count only applies to the core object.

That version should also work as long as we make memory the first thing inside a node before you reach the socket level.  We actually haven't specified what the convention is for that, but one perfectly valid one would be:

```yaml
type: Node
with:
  type: memory
  amount: ...
  with:
    type: core
    count: ...
```

That's a good point @garlick, by the time it's canonicalized, the program will have a "resources" key.  That is assuming at least we decide to stick with special-casing program and/or task rather than treating them as standard resources themselves.

trws:
Actually, the convention for memory is a bit tricky if we want to be able to represent the existence of numa-nodes.  May have to think about that one a bit...  It's almost a segmented pool, which is something I'm not sure we have a representation for.

grondo:
> That version should also work as long as we make memory the first thing inside a node before you reach the socket level. We actually haven't specified what the convention is for that, but one perfectly valid one would be:

No, I think memory should be at the Socket level in at least most cases. That is one of the benefits of hierarchical resource model...

grondo:
> It's almost a segmented pool, which is something I'm not sure we have a representation for.

Yes you bring up a good point, the memory is attached at the Socket and thus is contained within it, however, you can indeed reach the memory from cores on other sockets... Eesh.

trws:
Right @grondo, that's what I was just thinking too.  Maybe it just works if we let the quantity of a thing in a pool be summed from several child resources the way counts are...

SteVwonder:
> Also, this might be a case where using task: instead of group makes things more intuitive, though my brain still gets queasy looking at resources: task:...
```yaml
resources:
    task:
        command: app
        count: 10
        with:
            - type: memory
            amount: 2G
            - type: core
```

@grondo:
I kinda like the "-type: task" as opposed to "task:" when nested under the "resources:" tag.  I think it is slightly nicer to look at in that particular case (totally subjective though).
e.g.:

```yaml
resources:
    - type: task
      command: app
      count: 10
      with:
            - type: memory
              amount: 2G
            - type: core
```

SteVwonder:
Being super nitpicky...

1) some of the examples use the '-' marks, e.g.:
```yaml
resources:
    - type: Switch
    count: 3
    with:
        - type: Core
```
while others don't, e.g. (use-case 6):
```yaml
resources:
    type: node
    count: 4
    with:
        type: socket
        count: 2
        with:
            type: core
            count: 4
```
I'm pointing this out purely from a consistency standpoint, but it does appear that the '-' has a syntactic meaning in YAML. I'm unsure if it actually affects the correctness in these examples.

2) The ordering of the use-cases might be "smoother" if it starts out with all of the "task-free"/"resource-only" jobspecs (i.e., 1, 5, 6, 11, 12) and then move onto the use-cases that include tasks.

Sorry for the nitpicks and nothing really of substance.

SteVwonder:
Of slightly more substance...

I did run a `grep srun` on the moab scripts from cab over the past 13 months.  There are ~500K instances of sruns during that time period.  I haven't attempted to parse out the set of all arguments used, but I did a quick analysis of some common arguments listed in this thread and the srun man page:

1. `--multi-prog` occurs ~2.6K times
1. `--gres` occurs 0 times
1. `-B` occurs 5 times
  1. Verified that all 5 instances were passed to SLURM (and not the user's app)
1. `--sockets-per-node` occurs 18 times
1. `--cores-per-socket` occurs 0 times
1. `--threads-per-core` occurs 0 times
1. `--ntasks-per-`:
  1. `core` occurs 0 times
  1. `node` occurs ~7K times
  1. `socket` occurs 12 times
1. `--cpu_bind` occurs 32 times
  - Unrelated: I wonder why is this underscored as opposed to hyphenated like the other arguments?
1. `--cpus-per-task` occurs ~2.9K times
  1. The '-c' version is also used quite frequently but it's a touch difficult to disambiguate between a -c passed to srun and a -c passed to a user's app
1. `-distribution` / `-m <block|cyclic|arbitrary|plane=<options>[:block|cyclic]>` occurs ~2.2K times
1. `--overcommit` occurs ~1.3K times
  1. This does not include `-O`
1. `--partition` occurs ~1.4K times
  1. This does not include `-p`
1. `--switches` occurs 182 times
1. `--nodelist` occurs 95 times
  1. This does not include `-w`
1. `--mpi` occurs ~2.6K times
1. `--array`occurs 0 times
  1. I assume this is an sbatch argument, but I checked for it anyways

I can analyze additional options and/or clusters if there is interest.

grondo:
@SteVwonder, thanks for all the comments! Very helpful actually.

> 1) some of the examples use the '-' marks, e.g.:

You are correct of course! Sorry I was being lazy and not really paying attention to whether we were writing lists or dictionaries, or even paying attention to whether I was writing valid YAML. It turns out my brain has a really hard time reading yaml.

I will go back and try to fix up these examples. It certainly seems like the value of the `resources` and `with` (and `edge`?) keys should be a strict list (list of resource type definitions), so the tick `"-"` should always be required.

I will also put the examples at least through a YAML validator. Some of the current examples don't even parse. :disappointed:

lipari:
>  --array occurs 0 times
>    I assume this is an sbatch argument, but I checked for it anyways

It is indeed an sbatch argument and only offered in versions of Slurm newer than are on LC Linux clusters.  I mention this because, contrary to what the statistics may imply, a Flux analog to job arrays will be important going forward.

grondo:
Can we agree, in the abstract, on the jobspec language definition in terms something like the following?

 * basic jobspec consists of a null document or yaml dictionary, null document is replaced with defaults
 * Currently supported dictionary keys include `tasks` and `resources`
 * `resources` key is a strict list of dictionaries, each element is required to have a `type` key, other supported keys include `count/amount`, `attributes` (?), and `with`. Specific `type` values may have other supported keys, e.g. `type: task` supports `command`
 * The special `with` key is a strict list of dictionaries similar to `resources` at the given nested level.

Actually, if any of that is correct, it occurs to me that `with:` and `resources:` are actually synonyms. It might be easier to define the language recursively if they were interchangeable... (Though I'm sure I missed something there, I still don't think I've had my "a-ha" moment)

grondo:
> I mention this because, contrary to what the statistics may imply, a Flux analog to job arrays will be important going forward.

@lipari I totally agree. My hope is that Flux will have something more powerful and flexible than job arrays, but which could implement job arrays as a subset.

Also, I'm sure you have some interesting use cases we could tackle here. Feel free to post them and we'll work through it!

garlick:
@grondo those constraints would really help to focus things for me.  What you have there could be reworded in RFC 2119 lingo and committed as the first paragraph under _Jobspec Language Definition_ .if others are on board with the content, which FWIW looks good to me including the simplification of with + resources.

grondo:
@trws, @SteVwonder: I find going back through the examples in the use cases I'm not clear on the `tasks` key. If we treat tasks as resources, does this key go away and require use of `- type: task`?

At the top level, is the `tasks` key similar to `resources` in that it should require a strict list?

grondo:
Just working through Use Case 2, I am kind of assuming there is no functional difference between
```yaml
resources:
    - type: node
      count: 4
      tasks:
        - command: hostname
          count: 5
```
and
```yaml
resources:
     - type: node
       count: 4
       with:
          - type: task
            command: hostname
            count: 5
```

i.e. `tasks: [ {dict1}, {dict2}, ...]` at any given level expands to
```yaml
with:
  - type: task
    dict1
  - type: task
    dict2
   ...
```

If so, then in the *canonical* jobspec should we require only the 2nd form?

If there is a difference please let me know and I apologize.

trws:
I've been a bit lax about this unfortunately, because when I put the original together it had a specification that worked like this:

*  The top level represents a program, which can be represented as:
  * If the document is null, it is replaced with the defaults
  * if the document consists of a single string object, that string is interpreted as a short-form syntax string or stand-alone resource type, other keys replaced with defaults
  * If the document consists of a list, that list is interpreted as a list of resources
  * if the document is a hash, then that hash must conform to the format of a program and include the resource, task and/or walltime keys
* Each resource list is interpreted as follows:
  * if it's a string, interpret as a list with one dict with the key type set to that string
  * if it's a list, interpret each entry as a resource
  * if it's a dict, it's a list with one dict that must conform to a resource

Because of the decay rules, I've been a bit lax in some places in how I write these things.

Ironing out the exact rules for the canonical version would be a good thing, I'd like to chat with you @grondo and anyone else interested about this so we can iron it out in person a bit.  The rules you set out sound reasonable, but how task placement and mapping interact is part of the question about where "task" items are in the hierarchy, and before we mandate placement it would help to talk that out for a few minutes.  Are folks around today?

grondo:
@trws meeting this afternoon sounds great!

lipari:
I'll join the meeting once you have a time/place.

lipari:
Two YAML items:
* The YAML reference listed in the PR should be updated to the latest version:  http://yaml.org/spec/1.2/spec.html
* The first YAML validator I found is http://www.yamllint.com/

garlick:
Sorry for the outdated YAML citation (good catch @lipari).

Regarding job arrays - canonical jobspec is limited to one program, so I assume job arrays are out of scope here, and any non-canonical jobspec or other representation of multiple programs gets turned into multiple canonical jobspecs for the queue.

grondo:
Ok, I've updated the Use Cases. I thought @SteVWonder's suggestion to reorder was a good one, so I've split the Use Cases into two sections. Section 1 has resource only jobspec, and Section 2 includes the Use Cases with task specification. Use cases are now numbered per section (1.1, 2.1, etc)

Also, I think I ran all the examples through the YAML parser at least. Now they at least parse, though I think a few of them are not quite right yet.

grondo:
I'm going back through Use Cases given some of the decisions that were made yesterday. I'm going to attempt to redo the canonical jobspec, however right away I've come up with a couple questions.

 * There is now no  canonical "resource only" request jobspec. At least one task must be specified IIUC. I assume in most cases the default task spec would be one `flux start` per `node`. Given this, what will the canonical job spec look like for 10 cores?
```yaml
resources:
  - type: core
    count: 4
  - type: node   # What does this mean 4 cores with at least one node?
    count: 1
    slot-id: node
tasks:
  - command: [ "flux", "start" ]
    slot-id: node
```
Does that even work? It would be much easier to "append" default task like this if the slot could be specified under `with:` (or maybe synonym `per:`?) in the tasks section. I'm sorry I'm probably just missing something simple.
```yaml
resources:
  - type: core
    count: 4
tasks: # Default system tasks section
  - command: [ "flux", "start" ]
    with:
      - type: node
        count: 1
```

* Before we moved tasks to the bottom of the jobspec, it was intuitive that a count of tasks at the top was mapped (in an unspecified way) over the set of resources that match the `resources` section. Now that `tasks` must reference `slot-ids` how is a similar result acheived? e.g. the new equivalent to Use Case 2.6:
```yaml
tasks:
    - command: app
      count: 10
resources:
    - type: node
      count: 2
      with:
        - type: memory
          amount: 4G
```
If you identify a node as a slot
```yaml
    - type: node
      count: 2
      slot-id: node # Actually is this "slot" 1 or 2 nodes?
```
and reference it in the tasks section
```yaml
tasks:
    - command: app
      count: 10
      slot-id: node
```
Does this say each task runs on the slot with 2 nodes (or 1 node if that is what `slot-id` indicated), and if there are more tasks than slots, then tasks are distributed more than one task to a slot?

trws:
Good points @grondo, and thanks for pushing on this the way you have 
been.  A few things have been hitting me as well, and it might be good 
to hold off on this until we can chat again Monday.  I'll try and have 
the python visualization/experimentation thing back together for this 
format by then so we can *see* what we're working with.  It's so hard to 
conceptualize this without the concrete implementation, or at least 
visualization, behind it that I think we (or at least I) keep stepping 
into potholes accidentally. ><

On 15 Jul 2016, at 10:47, Mark Grondona wrote:

> I'm going back through Use Cases given some of the decisions that were 
> made yesterday. I'm going to attempt to redo the canonical jobspec, 
> however right away I've come up with a couple questions.
>
>  * There is now no  canonical "resource only" request jobspec. At 
> least one task must be specified IIUC. I assume in most cases the 
> default task spec would be one `flux start` per `node`. Given this, 
> what will the canonical job spec look like for 10 cores?
> ```yaml
> resources:
>   - type: core
>     count: 4
>   - type: node   # What does this mean 4 cores with at least one node?
>     count: 1
>     slot-id: node
> tasks:
>   - command: [ "flux", "start" ]
>     slot-id: node
> ```

This is a good point, and ties back to when we had the "resource spec" 
that became the jobspec stuff.  A jobspec, I think, needs to have a 
task, but the format and data structures are just as good for 
representing arbitrary resource sets, just without one or more tasks in 
the mix.  For this example, I'd say you'd get one flux start on the one 
node contained to four cores, but it's a peculiar case to reason about 
isn't it?  Maybe having the tasks only, at least logically, in the tree 
but allowing an absolute count (rather than an expanded count) would be 
more sensible?

> Does that even work? It would be much easier to "append" default task 
> like this if the slot could be specified under `with:` (or maybe 
> synonym `per:`?) in the tasks section. I'm sorry I'm probably just 
> missing something simple.
> ```yaml
> resources:
>   - type: core
>     count: 4
> tasks: # Default system tasks section
>   - command: [ "flux", "start" ]
>     with:
>       - type: node
>         count: 1
> ```

You're not, and this might really be the easier way to go.

>
> * Before we moved tasks to the bottom of the jobspec, it was intuitive 
> that a count of tasks at the top was mapped (in an unspecified way) 
> over the set of resources that match the `resources` section. Now that 
> `tasks` must reference `slot-ids` how is a similar result acheived? 
> e.g. the new equivalent to Use Case 2.6:
> ```yaml
> tasks:
>     - command: app
>       count: 10
> resources:
>     - type: node
>       count: 2
>       with:
>         - type: memory
>           amount: 4G
> ```
> If you identify a node as a slot
> ```yaml
>     - type: node
>       count: 2
>       slot-id: node # Actually is this "slot" 1 or 2 nodes?
> ```
> and reference it in the tasks section
> ```yaml
> tasks:
>     - command: app
>       count: 10
>       slot-id: node
> ```
> Does this say each task runs on the slot with 2 nodes (or 1 node if 
> that is what `slot-id` indicated), and if there are more tasks than 
> slots, then tasks are distributed more than one task to a slot?

A single node would be the slot, and there are two slots, given the 
definition there.  You're right about the distribution.  The other thing 
here is there is a property of a resource, the executable property for 
lack of a better term, that defines whether a task can actually run it 
or not, that is necessary to make this work by avoiding the memory 
level.  It really feels like we need a better way to control whether one 
gets an absolute number of tasks of a given type, or a number 
determined-from or affected-by the resources.

grondo:
Thanks, @trws. I think a meeting where we go through the existing examples and specifically nail down each one will be good. Then we can use what we decide there to write up the Language Definition section. Honestly, this is one time where it feels like the instant feedback in a meeting will get us to converge on a solution much faster than the Github issues format ;-)

grondo:
FYI - I pushed updated use case examples to match the latest proposal in #53.

I kept the "resource only" requests for now, even though they violate the latest design. For flux, I think launching a new instance is the equivalent to `oarsub` `qsub` `salloc`, etc., so I could adapt those examples to run `flux start` once per `type: node` as the required task.

However, I'm kind of wondering how we'll do that in examples like *Use Case 1.4*, with a request for non-uniform nodes.. I guess we could do something like
```yaml
resources:
    - type: cluster
      count: 1
      with:
      - type: slot
        label: A
        with:
          - type: node
            count: 1
            with:
              - type: memory
                amount: 4GB
              - type: ib10g
        - type: switch
          with:
            - type: slot
              label: B
              count: 2
              with:
                - type: node
                  count: 1
                  with:
                    - type: core
                      count: 1
walltime: 4h
tasks:
  - command: [ flux,  start ]
    slot: A
    count:
      per_slot: 1
  - command: [ flux, start ]
    slot: B
    count: 
       per_slot: 1
```


it would clearly be nice to either be able to use "type slots" or something directly (somehow state that 1 copy of `flux start` should be run on each node that appears in the result), or at least to be able to reference a list of slots in the tasks section (e.g. `slot: [ A, B ]`).

In the non-canonical jobspec we'll want to allow resource only requests and magically turn it in to this, and if the user doesn't specify they need "type: node" somewhere in the request graph, that might be difficult without support for something like the first case above.

trws:
Well... I'm tempted to say "let's just do it as is and see if it's 
sufficiently annoying" but the one thing this reminds me of is we had a 
thought at one point to have a way to specify the slot by "level" or 
resource type, kinda the way hydra specifies the mapping level.  That 
would be a really clean way to express this kind of thing, as the 
default would just be something like `{level: node}` or maybe `{slot: 
{level: node}}`.

Thoughts?

grondo:
@trws, something like that was exactly what I was thinking. From the parser, graph-generator perspective, will that work out pretty well?

I'd say we make slot a dict then, with supported keys `label` to key off a labeled slot in the jobspect, or `level` as you say above, and for `salloc`/`oarsub` type requests, the default task block could simply be:
```yaml
tasks:
  - command: [ "flux", "start" ]
    slot:
      level: node
    count:
      per_slot: 1
```

That seems pretty nice to me.

grondo:
A related question, could this then simplify many of the "degenerate" slot cases (using a simple resource as a slot, e.g. "core" or "node")? i.e. from *Use Case 2.1* we have 
```yaml
resources:
  - type: slot
    label: default
    count: 4
    with:
    - type: node
      count: 1

tasks:
  command: hostname
  slot: default
  count:
    per_slot: 5
```

Could this reduce to:
```yaml
resources:
  - type: node
    count: 4
tasks:
  command: hostname
  slot:
    level: node
  count:
    per_slot: 5
```

spec_14.adoc
```diff
@@ -232,12 +232,18 @@ or
...
+    count: 4
+    with:
     - type: node
-      count: 4
-      with:
-        - type: task
-          command: hostname
-          count: 5
+      count: 1
+
```
lipari: This errant newline breaks the yaml formatting

spec_14.adoc
```diff
@@ -248,13 +254,16 @@ default distribution
...
 tasks:
-    - command: hostname
-      count: 5
-with:
-    - type: node
-      count: 4
-
+  - command: hostname
+    count:
+      total: 4
```
lipari: Needs a trailing blank line to end the yaml block

spec_14.adoc
```diff
@@ -264,14 +273,18 @@ for a total of 20 cores
...
     - command: myapp
-      count: 10
-      with:
-          - type: core
-            count: 2
-
-
+      slot: default
+      count:
+	per_slot: 1
```
lipari: needs indentation

spec_14.adoc
```diff
@@ -285,20 +298,30 @@ Existing Equivalent:: None Known
...
-        command: read-db
         with:
           - type: core
           - type: memory
             amount: 4GB
-      - type: task
-        command: db
+      - type: slot
+        label: db
+	count: 1
```
lipari: needs indentation

trws:
Yeah, I think it would simplify a lot of those. From the reader and matched perspectives it's no harder at all really. The question is whether it would be harder to interpret from a job execution perspective. I don't really think it would be, since it could just turn into a match against the assigned resources selecting resources at that level then schedule over that, but that's the only thing I think it would take.

Sent from my iPhone

On Jul 26, 2016, at 4:30 PM, Mark Grondona <notifications@github.com<mailto:notifications@github.com>> wrote:


A related question, could this then simplify many of the "degenerate" slot cases (using a simple resource as a slot, e.g. "core" or "node")? i.e. from Use Case 2.1 we have

resources:
  - type: slot
    label: default
    count: 4
    with:
    - type: node
      count: 1

tasks:
  command: hostname
  slot: default
  count:
    per_slot: 5

Could this reduce to:

resources:
  - type: node
    count: 4
tasks:
  command: hostname
  slot:
    level: node
  count:
    per_slot: 5

-

grondo:
Thanks for line notes @lipari! I'll try to more carefully check changes before committing next time.

lipari:
First off, shouldn't the task total count in Case 2.2 be 5?

Also, an earlier discussion proved helpful in demonstrating exclusivity w/o the exclusive key.  I propose squeezing in a new Case 1.4 (and bumping the current 1.4 to 1.5) to demonstrate the differences:

Case 1.4 A request for an exclusive allocation of 4 nodes that have at least two sockets and 4 cores per socket:

```yaml
- type: slot
  with
  - type: node
    count: 4
    with:
      - type: socket
        count: 2
        with:
          - type: core
            count: 4
```

grondo:
> First off, shouldn't the task total count in Case 2.2 be 5?

Yes, I just caught that too.

Great example recommendation, I'll throw that one in as you suggest

grondo:
Ok, @lipari. I made the updates you suggest (quickly so look out for typos).

I also update the "resource only" use case section with a preamble as to what that might mean for the jobspec language being described. I also added `version:` and `walltime:` to all examples.

spec_14.adoc
```diff
@@ -178,7 +178,29 @@ resources:
             count: 4
 
 '''
-Use Case 1.4:: Complex example from OAR
+Use case 1.4:: Exclusively allocate nodes, while constraining cores and
+sockets.
++
+Specific Example: Request an *exclusive* allocation of 4 nodes that have at
```
lipari: (correcting my own edit...)  another ':' is required after Specific Example:

grondo: Oh sorry I think that was my fault

spec_14.adoc
```diff
@@ -178,7 +178,29 @@ resources:
             count: 4
 
 '''
-Use Case 1.4:: Complex example from OAR
+Use case 1.4:: Exclusively allocate nodes, while constraining cores and
```
lipari: OCD people would love to see the C in case capitalized to match all the other occurrences of "Use Case".

grondo: Good catch, thanks!

grondo:
I've rebased this one on top of current master. I kept the commits separate for now until we are happy with the result, so we can see the progress and changes clearly.

SteVwonder:
These examples look great!  The flow is really nice, and I think the slot format turned out quite well.

Just to make sure I have the correct understanding... For use cases 1.3 & 1.4 (shared/exclusive allocation of 4 nodes), since the implicit task is set to run at the node level, 4 instances of flux will be started (1 per node) in both use cases, correct?  In other words,the introduction of the slot does not change the execution level (for that implicit task).

grondo:
@SteVwonder, thanks for taking a look!

> For use cases 1.3 & 1.4 (shared/exclusive allocation of 4 nodes), since the implicit task is set to run at the node level, 4 instances of flux will be started (1 per node) in both use cases, correct? In other words,the introduction of the slot does not change the execution level (for that implicit task).

That is my understanding -- All of the Section 1 use cases should be read as if the assumed `tasks` section mentioned above is appended to the jobspec. We could explicitly append that section if it would be more clear, though I was worried it would add too much detail.

I suppose this implies that all labeled slots are not required to be referenced by a task in the `tasks` section. I wonder if this will lead to confusion...

SteVwonder:
@grondo: I think the implicit appending is fine.
Your point about labeled slots not having a task which references them is a good point.  Maybe that should be a warning that is presented to the user when the verifier parses the jobspec before sending it off to the scheduler?

grondo:
Still TODO here is add some kind of UQ test case (though I'm not certain that makes sense for canonical jobspec which is not aware of other jobs), and some kind of storage job Use Case.

spec_14.adoc
```diff
@@ -401,7 +401,361 @@ walltime {
...
+walltime: 1h
+resources:
+  - type: slot
+    label: 2GB-node
+    count: 2
+    with:
+    - type: node
+      with:
+        - type: memory
+          amount: 4G
```
SteVwonder: For consistency, this should be `4GB`.

spec_14.adoc
```diff
@@ -401,7 +401,361 @@ walltime {
...
+tasks:
+  - command: app
+    slot: 2G-node
```
SteVwonder: For consistency, this should be `2GB`.

SteVwonder:
For a simple storage use-case, maybe something similar to the request for a specific amount of memory (with the exact location of the disk resources (i.e., burst buffer (BB) & parallel file system (PFS)) depending on their locations within the resource hierarchy).  In this example, I assume that the PFS is the root of the resource tree and the on-node BBs are located under the compute nodes, as siblings to the on-node memory.

``` yaml
version: 1
walltime: 1h
resources:
  - type: parallel-file-system
    amount: 1TB
    with:
    - type: node
      with:
        - type: memory
          amount: 4GB
        - type: burst-buffer
          amount: 200GB
```

Some questions I had while writing this example:

1. Can types have spaces in the name (e.g., parallel file system)?
2. How would you simultaneously request capacity and bandwidth on a PFS/BB?
3. If a center has multiple Lustre file systems and multiple GPFS file systems, how would the user specify that they want one type of PFS over the other (e.g., prefer Luster over GPFS) without limiting the jobspec to a single PFS or explicitly listing all of the Lustre filesystems as potential matches (e.g, how to avoid writing something like `lscratcha OR lscratchb OR lscratchc OR lscratchd OR lscratche OR lscratchf`)?

SteVwonder:
Also, just a note on a more advanced storage use-case (this doesn't need to be solved immediately, but I figured I would bring it up since it will be an important use-case in the future for not only storage but also network bandwidth and power).  After speaking with @dongahn the other day about burst buffer systems, it would be nice to have the ability to specify a job that has X GBs of space in the BB for staging-in data from the parallel filesystem and then X+Y GBs of space in the BB for actually running the user's app (since additional space will be needed for new checkpoints, etc).  When I thought about how to write this in the current jobspec, three ideas came to mind.

1. This information could be provided in the `properties.user` dictionary of a jobspec as two separate values, one for stage-in and one for execution.  A "BB-aware" scheduler could explicitly look for these values in the `properties.user` dictionary and then schedule the job according to some special BB logic. To me, this is a simple, straightforward way to do it.
2. The initial jobspec could request only X GBs of space, and then once the stage-in is complete, the job/user app could request additional space on the BB as needed via the dynamic scheduling interface.  The main problem here is that the job is never guaranteed to actually receive any additional capacity.
3. The jobspec could add support for time as a dimension in the `count`/`amount` fields.  Then malleable/evolving schedulers could take advantage of this additional knowledge, and rigid schedulers could just use the max of the `count`/`amount` fields.

dongahn:
@SteVwonder makes good points about storage use cases.

Given the importance of the burst buffer resources for future systems, I propose our canonical jobspec covers some of the basic burst buffer use cases. Here are a few criteria I think we should consider. Our canonical jobspec should be able to

1. express burst buffer requirements for at least two different emerging implementations and various storage access modes of them: 
    - Node-local NVRAM (our Sierra) implementation with node visibility
    - Intermediate  BB nodes ([Cray's DataWarp](http://docs.cray.com/books/S-2558-5204/S-2558-5204.pdf)) with stripped access mode (job-level BB filesystem mount and visibility)
    - Intermediate  BB nodes (Cray's DataWarp) with private access mode (node-level mount/visibility)

2. express requirements for BB capacity, file sets that need to be staged in and out, and bandwidth

3. express the capacity requirements in a multi-staged fashion, the information essential for the scheduler to overlap the job's stage-in phase with the currently running jobs on the target nodes

dongahn:
It seems we can do item 1 quite easily with the current jobspec 

For node-local implementation and BB nodes implementation in private mode:
```yml
resource:
  - type: node
    count:
        min: 4
    with:
      - type: bb
```

For BB nodes implementation in striped mode (all of the allocated nodes commonly access the BB file system instance):
```yml
resource:
  - type: node
    count:
        min: 4
  - type: bb
```

dongahn:
Capacity, staging file sets, and bandwidth requirements can use more discussions. First to describe the staging requirements, my first attempt is to treat the staging file sets as the resources and hang them under the burst buffer resource

```yml
resource:
  - type: node
    count:
        min: 4
    with:
      - type: bb
        with:
          - type: stage-in
            files: 
              - /lscratchb/user1/cr2.$rank
              - /lscratchb/user1/cr3.$rank
          - type: state-out
            files: 
              - /lscratchb/user1/cr100.$rank
              - /lscratchb/user1/cr101.$rank
```

$rank or some built-in variables would be needed for the spec to express per-node files to stage. I used `files` to as the key to describe the staging files as a list of files. But one can also introduce `dirs` and `flistfile` as alternative ways to describe the staging files. 

For BB nodes implementation in striped mode, all of the files that the job as a whole can be described in the same manner.

dongahn:
First to describe the capacity requirements, we at least need to specify two quantifies: the storage requirement for the stage-in phase and the running phase. The former is typically smaller in size than the latter, and hence this should help the scheduler to overlap the stage-in phase with the currently running jobs' computation or stage-out phase. My first attempt, which I don't yet like:


```yml
resource:
  - type: node
    count:
        min: 4
    with:
      - type: bb
        count:
            min: 20GB
        with:
          - type: stage-in
             count:
                min: 10GB
             files:
                -  /lscratchb/user1/cr2.$rank
```

I can imagine we can also:

```yml
resource:
  - type: node
    count:
        min: 4
    with:
      - type: bb
        count:
            min: 10GB
            max: 20GB
        with:
          - type: stage-in
             files:
                -  /lscratchb/user1/cr2.$rank
``` 

But my mind doesn't easily translate this as 10GB for stage-in and 20GB for running.

dongahn:
For bandwidth requirements, I think for now we only need to be able to specify the bandwidth between the BB layer and the PFS. (We treat CN to BB as a highway)

But even then, I'm not sure how to express this yet. I have to think my trouble is because I don't fully understand how general graph works in our jobspec yet. Perhaps we can use this as an example to show how to specify resource requirement in a general graph. 

In any case, I have to think that IO-bandwidth awareness can help increase resource utilization when we schedule jobs to overlap jobs' stage-in operations with the existing jobs. BB-PFS bandwidth awareness can allow our scheduler to delay stage-in operations as late as possible and help avoid early node locking of jobs.

dongahn:
Oh BTW. To support stage-in and -out state of the job, at some point we may also need to revise the  job states of our scheduler and [task/program execution service rfc](https://github.com/flux-framework/rfc/blob/master/spec_8.adoc)

grondo:
@SteVwonder, Thanks, good catches! I think I've addressed them in the latest commit.

@dongahn, @SteVwonder, The storage and bandwidth use cases are sounding like they might generate some discussion. Is there some simple cases you'd specifically like to include in this PR? Or would it be ok to finalize the kind of basic Use Cases here, then perhaps tackle any addenda or modifications to the jobspec definition *plus* use cases in a separate PR, which could have its own chain of discussion?

dongahn:
@grondo: how about completing this PR with non storage use cases to make things a bit more tractible. And open up a new one with a goal to generate storage use case discussions + extend the jobspec? To begin with,  we can move the existing use case discussion there then?

dongahn:
@grondo: 

> Still TODO here is add some kind of UQ test case (though I'm not certain that makes sense for canonical jobspec which is not aware of other jobs)

Thank you for the reminder.

Yeah it seems UQ support abstractions should be one level above the jobspec. But I am still curious if there is anything we can or should call out as to how jobspec can help UQ style high throughput scheduling. 

Is reusability of a jobspec is a feature that can be exploited for optimization? 

If there is slight program data differences between two jobspecs, how do we reuse an existing jobspec? 

If there are some unique features, dd we want to advise how the jobspec reader/builder should be implemented to provide optimization for UQ?

One idea might be to have a couple of *extended* use cases and describe how a upper level simple jobarray and/or UQ support (this could be capacitor) will generate a manageable number of jobspec objects and how our resource/job subsystem can use the deduplicated objects to optimize scheduling and execution?

I haven't given much thoughts here yet and this rfc may not end up being a good place to have this. But I thought we should put some thoughta there since this is an important initial use cases.

lipari:
I'm going to suggest that rather than attempting to pave the route to a UQ-style job spec now, we hold off and wait to see what UQ elements become typical or standard.  We could surmise about representing a range of input parameters, output parameters, complex dependencies, job to resource packing, timeouts, and other contingencies etc., now and try to capture a model for everything.  Or, we could just go with a single job spec, and wait until we see what patterns of UQ actually arise and become important and essential.

garlick:
Would it make sense to merge in what's been done here and open a new PR for whatever the next step is?

grondo:
Yes, I think that is what @dongahn suggested and I agree with that as well. As long as there are no glaring errors here I think this could be merged and the higher level discussion moved to a new issue/PR.

garlick:
In sec 1 intro there are a few typos
* "in terms a canonical" (missing of)
* "start and instance" (an?)
* probably new paragraph before "Thus..." is unnecessary
* can walltime statement be dropped since it appears that is explicit in all examples?

garlick:
One very minor suggestion that might improve readability is if the "existing equivalent" command lines are broken into one line per example, without the "or" conjunction.

If it were possible to consistently label them with "slurm:", "oar:", etc. that might be nice but it also might just add noise.  Not sure about that one.

grondo:
OK, good suggestion! I'll do that. Once we are happy with this PR I'll squash into a single "Add Use Cases" commit.

grondo:
> probably new paragraph before "Thus..." is unnecessary

There is a line between the preceding paragraph and the "Thus..." sentence, where are you seeing no new paragraph? Did I mess up the asciidoc formatting somehow?

trws:
Thanks everyone for pushing on this.   I agree with many of the comments, and think this should merge as soon as the editorial comments are ironed out.  As to the storage use-case, I really want that to be a bit separate because I think that will be the first place where we really want to leverage a non-hardware-tree resource.  I say that because I think a user should be able to express the amount of burst buffer that each task needs and have it able to match with any of the three examples Dong mentioned on the system side.  I'm not sure we can do that with storage expressed at the full job level, but going from per-task to job level should work out pretty cleanly.


---
Sent from Boxer | http://getboxer.com<http://bxr.io/PBI3C>

On August 8, 2016 at 10:11:20 AM PDT, Mark Grondona <notifications@github.com> wrote:

probably new paragraph before "Thus..." is unnecessary

There is a line between the preceding paragraph and the "Thus..." sentence, where are you seeing no new paragraph? Did I mess up the asciidoc formatting somehow?

-

garlick:
>> probably new paragraph before "Thus..." is unnecessary
> There is a line between the preceding paragraph and the "Thus..." sentence, where are you seeing no > new paragraph? Did I mess up the asciidoc formatting somehow?

Sorry - I was suggesting that we drop the paragraph break before "Thus..." e.g. make that sentence part of the previous paragraph.  If that seems wrong, happy to let that one go.

grondo:
> Sorry - I was suggesting that we drop the paragraph break before "Thus..." e.g. make that sentence part of the previous paragraph. If that seems wrong, happy to let that one go.

Oh, my! I'm sorry I totally misread that, not sure how.

I'm fine with that change.

grondo:
Ok @garlick, I changed *Existing Equivalents* section of each use case to a table, with the name of the software project referenced in the first column. I also added a title for the jobspec yaml itself, which seemed to make each section more clear.

garlick:
Thanks, that helps I think!

One nit: "jobspec yaml" header in 1.5 is not formatted properly.

grondo:
Thanks! Should be fixed (I hope!) in the latest.

garlick:
Should the section 2 header be located after the horizantal line instead of before, and should it have a brief introduction like section 1?  (Sorry I didn't catch that when I scanned through before - I don't mean to "nickle and dime" you to death :-)

grondo:
Sorry, I've reformatted this so many times I can't see things like that anymore. Will fix!

garlick:
Shall we merge now and toss the sec 2 header fix into the next PR?

grondo:
Oh, sorry I have the formatting fix in but was trying to think of what to say for the Sec 2 intro.

I think we want to squash all the noise down before merging. Let me get the suggested extra horizontal rule in there and then squash down to 1 or 2 commits

grondo:
Ok, I added a short blurb to the front of Section 2 as a placeholder. If it is ok I'll squash now

garlick:
Looks good to me!

grondo:
Ok, all squashed together!

PR #53, opened 2016-07-15
Initial stab jobspec language definition

grondo:
Based on our meeting yesterday I took an initial stab at fleshing out some of the *Jobspec Language Definition* section. I'm not too happy with what's here, but maybe this can generate some discussion and we can get an initial version merged today?

### DIFF
--- a/spec_14.adoc
+++ b/spec_14.adoc
@@ -95,7 +95,306 @@ FIXME: Fill in
 
 == Jobspec Language Definition
 
-FIXME: Fill in
+A canonical jobspec YAML document SHALL consist of a dictionary
+defining the resources, tasks and other attributes of a single
+program. The dictionary MUST contain the keys `resources`, `tasks`,
+`walltime`, `attrs`, and `version`.
+
+Each of the listed jobspec keys SHALL meet the form and requirements
+listed in detail in the sections below. For reference, a ruleset for
+compliant canonical jobspec is provided using *JSON Content Rules*[2]
+in the *Content Rules* at the end of this section.
+
+=== Resources
+
+The value of the `resources` key SHALL be a strict list which MUST
+define at least one resource. Each list element SHALL represent a
+*resource vertex*  or resource descriptor object as a dictionary
+(described below).  The list of resources defined under the `resources`
+key SHALL represent a composite resource request for the program
+defined in the jobspec.
+
+A resource vertex SHALL contain the following keys:
+
+ *type*::
+ The `type` key for a resource SHALL indicate the type of resource to
+ be matched. Some type names MAY be reserved for use in the jobspec
+ language itself. Currently reserved types include `slot`, used to
+ define *task slots*, and `group` used to add an unnamed level of
+ grouping to a resource request. Reserved types are described in the
+ *Reserved Resource Types* section below.
+
+ *count*::
+ The `count` key SHALL indicate the desired number or range of
+ resources matching the current vertex. The value of `count` SHALL be a
+ dictionary which SHALL contain the following keys:
++
+[horizontal]
+   *min*::: The minimum required count or amount of this resource
+
+   *max*::: The maximum required count or amount of this resource
+
+   *operator*::: An operator applied between `min` and `max` which
+   returns the next acceptable value
+
+   *operand*::: The operand used in conjunction with `operator`
+   to get the next value between `min` and `max`.
+
+A resource vertex MAY additionally contain one or more of the
+following keys
+
+ *unit*::
+ The `unit` key, if supplied, SHALL have a string value indicating
+ the chosen units applied to the `count` value or values.
+
+ *exclusive*::
+ The `exclusive` key SHALL be a boolean  indicating, when true,  that
+ the current resource is requested to be allocated exclusively to
+ the current program. If unset, the default value for `exclusive` SHALL
+ be `false` for vertices that are not within a task slot. The default
+ value for `exclusive` SHALL be `true` for task slots (`type: slot`)
+ and their associated resources.
+
+ *with*::
+ The `with` key SHALL indicate an edge of type `out` from this resource
+ vertex to another resource. Therefore, the value of the `with` key
+ SHALL be a dictionary conforming to the resource vertex specification.
+
+ *edge*::
+ **XXX**: need specification for other "edge match descriptors"
+
+ *id*::
+ The value of the `id` key SHALL be a string indicating a set of
+ matching resource identifiers.
+
+ *uuid*::
+ The value of the `uuid` key SHALL be a string indicating a set of
+ matching resource UUIDs.
+
+ *tags*::
+ The value of the `tags` key SHALL be a dictionary with supported key
+ `list` which SHALL be a list of string tags required. Other keys MAY
+ be reserved for future or site-specific extensions.
+
+ *attributes*::
+ The value of the `attributes` key SHALL be a dictionary with supported
+ key `list` which SHALL be a list of dictionaries with required attributes
+ as keys, and their required values as the key values. Other keys in
+ the attributes dictionary MAY be reserved for future or site-specific
+ extensions.
+
+
+==== Reserved Resource Types
+
+*group*:: A resource type of `type: group` SHALL indicate an anonymous
+grouping of resources in the resource request.
+
+*slot*:: A resource type of `type: slot` SHALL indicate a grouping
+of resources into a named *task slot*. A `slot` SHALL be a valid
+resource spec including a `label` key, the value of which may be used
+to reference the named task slot during tasks definition. The `label`
+provided SHALL be local to the namespace of the current jobspec.
++
+A task slot SHALL have at least one edge specified using `with:`, and
+the resources associated with a slot SHALL be exclusively allocated
+to the program described in the jobspec.
+
+=== Tasks
+
+The value of the `tasks` key SHALL be a strict list which MUST
+define at least one task. Each list element SHALL be a dictionary
+representing a task or tasks to run as part of the program. A task
+descriptor SHALL contain the following keys:
+
+ *command*::
+ The value of the `command` key SHALL be a string OR list representing
+ an executable and its arguments.
+
+ *slot*::
+ The value of the `slot` key SHALL be used to indicate the *task slot*
+ on which this task or tasks shall be contained and executed. The
+ number of tasks executed per task slot SHALL be a function of the
+ number of resource slots and total number of tasks requested to execute.
++
+ The value of the `slot` key SHALL be a dictionary with supported key
+ of either `label` or `level`. The `level` key SHALL reference a `label`
+ from the `resources` list, indicating an explicitly created and named
+ *task slot*. The `level` key SHALL reference a resource name, such
+ as `core` or `node`, indicating an implicitly created *task slot* on
+ which to map the defined tasks.
+
+ *count*::
+ The value of the `count` key SHALL be a dictionary supporting at least
+ the keys `per_slot` and `total`, with other keys reserved for future
+ or site-specific extensions.
++
+[horizontal]
+  *per_slot*:::
+  The value of `per_slot` SHALL be a number indicating the number
+  of tasks to execute per task slot allocated to the program.
+
+  *total*:::
+  The value of the `total` field SHALL indicate the total number of
+  tasks to be run across all task slots, possibly oversubscribed.
+
+ *attrs*::
+ The `attrs` key SHALL be a free-form dictionary of keys which may
+ be used for platform independent or optional extensions.
+
+ *distribution*::
+ The value of the `distribution` key SHALL be a string, which MAY
+ be used as input to the launcher's algorithm for task placement and
+ layout among task slots.
+
+=== Walltime
+The value of the `walltime` key SHALL be a dictionary with currently
+supported keys `duration` or `range`. Other keys MAY be reserved for
+future or site specific extensions. It SHALL be considered an error
+if multiple, conflicting keys are set in the `walltime` dictionary.
+
+Supported keys for `walltime` are described in detail below.
+
+ *duration*:: The `duration` key SHALL have a string value interpreted
+ as a single duration walltime estimate/limit for the program described
+ by the current jobspec.
+
+ *range*:: The `range` key SHALL be a dictionary with duration keys
+ `min` and `max` indicating the minimum and maximum walltime
+ estimate/limit for the program in the current jobspec.
+
+=== Attributes
+The value of the `attrs` key SHALL be a dictionary of generic
+program attributes, system or user defined, which MAY affect program
+execution, scheduling, task placement, etc. Some key names in the
+`attrs` dictionary MAY be reserved.
+
+
+=== Example Jobspec
+
+Under the description above, the following is an example of a fully compliant
+version 1 jobspec. The example below declares a request for 4 "nodes"
+each of which with 1 task slot consisting of 2 cores each, for a total
+of 4 task slots. A single copy of the command `app` will be run on each
+task slot for a total of 4 tasks.
+
+[source,yaml]
+----
+version: 1
+resources:
+  - type: node
+    count: 4
+    with:
+      - type: slot
+        count: 1
+        label: default
+        with:
+          - type: core
+            count: 2
+tasks:
+  - command: app
+    slot:
+      label: default
+    count:
+        per_slot: 1
+walltime:
+  duration: 1h
+----
+
+A simpler example using implicit *task slot* definition to run 4 tasks
+across 4 nodes
+
+[source,yaml]
+----
+version: 1
+resources:
+  - type: node
+    count: 4
+tasks:
+  - command: hostname
+    slot:
+      level: node
+    count:
+      per_slot: 1
+----
+
+=== Content Rules
+
+A jobspec conforming to version 1 of the language definition SHALL
+adhere to the following ruleset, described using JSON Content Rules[2]
+draft version 0.6.
+
+----
+# jcr-version 0.6
+
+{
+   "resources" : [ +vertex ],
+   "tasks" : tasks,
+   "walltime" : walltime,
+   "attrs" :  { /.*/ : any },
+   "version" : 1,
+}
+
+vertex_common {
+    "count" : resource_count,
+    ?"exclusive" : boolean,
+    ?"with" : [ +vertex ],
+    ?"edge" : TBD,
+}
+
+group_vertex {
+    "type" : "group",
+    vertex_common,
+}
+
+slot_vertex {
+    "type"  : "slot",
+    "label" : string,
+    vertex_common,
+}
+
+resource_vertex {
+    "type" : ( :string, + @{reject} ( "group", "slot")),
+    vertex_common
+    ?"id" : string,
+    ?"uuid" : string,
+    ?"tags" : resource_tags,
+    ?"attributes" : resource_attributes,
+    ?"unit" : string,
+}
+
+vertex ( group_vertex | slot_vertex | resource_vertex )
+
+resource_count {
+    "min" 1..,
+    "max" 1..,
+    "operator" : ( :"+" | :"*" | : "^" ),
+    "operand" : 1..,
+}
+
+resource_tags {
+    ?"list" : [ +string ],
+}
+
+resource_attributes {
+    ?"list" : TBD,
+}
+
+tasks {
+    "command" : [ +string ],
+    "slot" : { "label" : string  | "level": string },
+    "count" : { "per_slot" : 1.. | "total" : 1.. },
+    "distribution" : string,
+    ?"attrs" : { /.*/ : any },
+}
+
+walltime {
+    "duration" : string |
+    "range" { "min" : 0.., "max" : 0.. },
+}
+
+----
+
+
 
 == Basic Use Cases
 
@@ -107,3 +406,4 @@ FIXME: Fill in
 == References
 
 * [1]http://yaml.org/spec/current.html#representation[YAML Ain't Markup Language (YAML™) Version 1.1], O. Ben-Kiki, C. Evans, B. Ingerson, 2004.
+* [2]http://www.ietf.org/id/draft-newton-json-content-rules-06.txt[JSON Content Rules (jcr-version 0.6), A. Newton, P. Cordell, 2016]
--- a/spell.en.pws
+++ b/spell.en.pws
@@ -284,3 +284,12 @@ Ingerson
 Kiki
 YAML
 mpirun
+Walltime
+walltime
+attrs
+UUIDs
+jcr
+ruleset
+yaml
+vertices
+hostname


grondo:
It occurs to me that `count` for a task cannot be required. There needs to be some way to specify in canonical job spec that the number of tasks actually executed is a function of the number of task slots actually allocated when a range is given.

trws:
* resource spec
  * each resource vertex shall contain:
    * id: integer vertex id
    * uuid: globally unique resource id
    * name: string, <type>-<id> or user-specified pattern
    * logical_id: id in local neighborhood
    * system_id: id in local neighborhood
    * type
    * amount:
    * unit:
    * a list of edges where edge has:
        * direction: in, out or inout
        * from-id: id or uuid
        * to-id: id or uuid
        * type: link-type, or graph type
        * attributes:
    * executable: boolean
    * attributes
* resource query: abstract description of a resource spec
  * each resource descriptor shall contain:
    * type
    * sharing: exclusive, shared
    * count:
        * min
        * max
        * operator
        * operand
    * unit
    * a list of edge matchers
* program spec
  * at least one resource, conforming to resource spec
  * at least one task with:
    * command: string or list
    * count_per_slot OR total: one or other, not both
    * total
    * distribution: policy?

grondo:
Ok, I pushed a new *Jobspec Language Definition* section which I hope captures the result of yesterday's meeting, and @trws comments above. This version is still pretty rough, but I'd like to narrow down on something acceptable to merge soon.

@trws, given the above, I'm still assuming a conformant jobspec "SHALL" be a dictionary consisting of the keys `resources`, `tasks`, `walltime`, and `attrs`. It seems like that still must hold, though I can't remember if that is required in your test parser/digester.

SteVwonder:
@grondo: looks good!

My only confusion is with the `id` key.  My understanding is that this key means two different things when in a query vs in a resource description.  In a query, it is used as a label that can be referenced later on in the query (i.e., when defining where a task should execute).  In a resource description, it is used as a unique identifier for a given resource and is used as a reference by Flux modules (e.g., resrc, wreck, etc) within that Flux instance.  Is this correct?

grondo:
> My understanding is that this key means two different things when in a query vs in a resource description. In a query, it is used as a label that can be referenced later on in the query (i.e., when defining where a task should execute).

Yeah, jobspec only deals with query so I think we're only talking about the use as a label. (The other use of `id` is detailed in RFC4 I think). Maybe it would be better to rename `id` to `label` for the jobspec?

trws:
I like that to help clarify the difference.  There may be a time when using an id or a UUID in a query is appropriate, if someone is requesting a specific resource by id or something.  I like how this is looking, and given our discussions the other day I agree that requiring resources and tasks as keys at the program level is the way to go.  Programs, tasks, and resources are distinct types now, and the "slot" resource type serves as the task target in the resource specification.

My parser does not currently work that way, but it will make life noticeably easier once I make it work that way.

lipari:
@grondo, the resource `id` term is overloaded.  As it is written in your PR, it refers to a resource `slot_id` and this differs from the resource `ID` attribute described in [RFC 4](https://github.com/flux-framework/rfc/blob/master/spec_4.adoc).  The task `slot` would reference the resource `slot_id`.  That's what I recall anyway.

lipari:
Sorry... I glossed over the earlier comments before lunch.  I don't mean to rehash what's been said.

grondo:
> Programs, tasks, and resources are distinct types now, and the "slot" resource type serves as the task target in the resource specification.

Hm, I perhaps wasn't clear on this particular point from yesterday. Are you saying that a "slot" is now a full type, e.g. `type: slot`, and not just defined by giving a resource vertex a "label", e.g.
```yaml
resources:
 - type: core
   label: default
   #  other keys left out for brevity
tasks:
  - command: myapp
    slot: default
```
vs
```yaml
resources:
  - type: slot
    label: default
    with:
        - type: core
          # ...
tasks:
  - command: myapp
    slot: default
```

Either way is fine with me actually, though the `type: slot` syntax is more verbose, it does seem more clear, and we'd need to group resources in many real-world scenarios in any case, e.g. you probably don't want a core without memory, so the canonical request for a core (using the group syntax) would likely turn into 
```yaml
resources:
  - type: group
    label: default
    count: { min: 1, max: 1, operator: "+", operand: 1 }
    with:
      - type: core
        count: { min: 1, max: 1, operator: "+", operand: 1 }
        sharing: exclusive
      - type: memory
        count: { min: Xmb_per_core, max: Xmb_per_core, operator: "+", operand: 1 }
        units: "MB",
        sharing: exclusive
tasks:
  - command: myapp
    slot: default
    count_per_slot: 1
    distribution: default
    attrs: {}
```

garlick:
Formatting comment: consider using capitalized subheadings for top level keys.  The indentation is a bit subtle in the github rendering with the current approach.

grondo:
> Formatting comment: consider using capitalized subheadings for top level keys. 

Thanks, I assume you are talking about the RFC not the yaml snippets? I think I noticed that about the document as well, however I was hesitant to change the case of top level jobspec keys just for the RFC. I don't actually like the way the whole thing in the RFC is presented and would be open to comments on a different way to present the constraints... a table perhaps?

trws:
Table, nested bullets so we have bullet style, or similar might be practical.  I'm thinking we might add an actual schema based on something like the [Rx yaml/json schema setup](http://rx.codesimply.com/) also, since that makes it trivial to verify the structure, if not the semantic content, of a spec.

grondo:
Unfortunately it appears Github rendering does not indent definition lists. I've changed formatting as @garlick suggested, but I'm afraid the indentation is still a bit subtle, but hopefully enough of an improvement that we can parse the content and review that part of the doc.

Sadly, when I was formatting I mixed in the changes from `id` -> `label`, so that change snuck in there as well.

However, @trws, the current language in this doc allows any resource vertex to be labeled as a slot with the `label` key, and I'm not entirely sure that is what you wanted, see [comment](https://github.com/flux-framework/rfc/pull/53#issuecomment-233764024) above.

I like the idea of formalizing a schema here as well -- If we do it as JSON, we could use the JSON content rules we've used elsewhere (as explained to me by @garlick).

garlick:
That looks quite a bit more readable to me :-)

Since a min/max count of a resource can be requested, does wallclock need to be expressible as a function of the actual quantity of resources allocated?  (maybe using a label like with slots?)

grondo:
> Since a min/max count of a resource can be requested, does wallclock need to be expressible as a function of the actual quantity of resources allocated? (maybe using a label like with slots?) 

Great point! Perhaps this was already discussed but perhaps it would be sufficient for now to say walltime is also a dict with at least one supported key, where the first key we support is `duration` or something similar. Extra keys could be reserved for future use (for instance a function, or a list of labels, etc.)

Your comment also reminds me that we should ensure we have some way in the future for users to be able to submit a list of resource descriptors that represent a set of alternatives (`||` vs `&&`). I thought perhaps @trws had an idea of how to do that with edges, but I don't quite fully grok how that would work. If we need to add an extra level to the jobspec, now is when we should do it. Like `walltime` we can define parts of the spec as a dict with only one supported key, with other keys reserved for future expansion.

dongahn:
> Great point! Perhaps this was already discussed but perhaps it would be sufficient for now to say walltime is also a dict with at least one supported key, where the first key we support is duration or something similar. Extra keys could be reserved for future use (for instance a function, or a list of labels, etc.)

I agree, this is a good point. This should be important for expressing moldable jobs. 

A simple way to express a `walltime` as a function of allocated quality may be to use a simple resource scaling factor? 

Specifically, users can either specify a constant `walltime` regardless of resource quantities allocated or as a function of the scaling factor: 

`walltime` request for maximum resource x scaling (where scaling is defined to be maximum / allocated resource quantity) x C (adjustment constant)

dongahn:
BTW, I just noticed this in one of the resource vertex fields: `system_id: id in local neighborhood`

I think this meant to be a system-defined id? 

It would also be good if we can pencil in the purpose of having two id space as well.

grondo:
@dongahn, I don't think we should conflate the "resource spec" or resource description which is currently defined in RFC4, and the jobspec "resource query spec" which is what we are defining in this RFC.

This RFC now uses "label" instead of "id" to denote task slot labels which can then be referred to in other parts of the jobspec, most notably task specifications.

As @trws points out, we will need to support `id` `logical_id` `tag` etc matching in the resource query spec, and that isn't there yet, but generic resources will have generic properties and attributes so those might fall under a `properties` key or other extensible component, and each property or attribute may have potentially different rules for matching. (e.g. id could take a nodeset style list, `tags` might be a syntax tree or string representing a query `"(griffin or bear) and goshawk and not slow"`). Maybe we can use the same rules for the various potential missing resource components?

grondo:
@dongahn: Sorry I just realized you were correcting a typo above, and not necessarily talking about adding `system_id` or `logical_id` to this RFC. Apologies! and sorry for my tangential diatribe!

dongahn:
@grondo: yes the last comment was a typo correction. Sorry, I should have inlined my comment with the original posting.

lipari:
Three misc comments:
* walltime, like a count, could be a range
* The concept of a task slot of resources needs better definition in the Resources section.  Specifically we need to state whether all the resources described in the resource section is one task slot, or whether it is a aggregation of multiple task slots, or whether multiple task slots can be defined under a single `resources` key.
* We *could* theoretically support "count_per_slot AND total" tasks.  Do we want to?

dongahn:
@lipari, good points. I think 'walltime' as a range support the same concept as the scale fator-based spec I sugested, if i understand you right. One is just more explicit than the other. 

So as far as we are clear on what a walltime range means in the spec and covers the scale factor case, I think this is a good idea.

trws:
Using a range for walltime, that matches or exceeds the number of levels allowed by the corresponding resource ranges, as @lipari suggests was also the idea that Suraj, @tpatki and I settled on when hashing over the walltime issue last summer.  I think that's the easiest way to apply it in the short term, and we can always extend it later.

As to supporting both per and total, we certainly can, but we need to be clear about which overrides the other.  I would probably say that total is the maximum, regardless of "per_slot," but either way it would need to be explicit.

lipari:
@dongahn, I see two approaches to consider and I was inserting the first:
* Give me `X` resources for anywhere between `m` to `n` minutes.
* Give me `X` resources for `n` minutes or `(X / 4)` resources for `(n * 4)` minutes

trws:
As @grondo points out, RFC4 defines the actual resource hierarchy spec, and we have diverged in a couple of places.  It might be good to reconcile some of these, since this is effectively defining the find and/or match components of RFC4 in terms of an RFC4 resource graph.  The resource spec syntax, or what goes into the "resources" key, could also be considered a valid format for a serialization of at least an abstract resource graph as defined in RFC4, though it might need a couple of tweaks.

The only things that jump out at me:

* 4 requires basename, name and id such that name is basename-id
* tags exist on pools and resources, these are valueless attributes, so they could be modeled, but currently are not in any explicit way
* properties == attributes, we can just change the name of the key or say they're the same and this one works
* Size and allocation table: size we talked about, or count/amount/unit for requests, we have also discussed having the allocations represented by incoming edges that record the amount consumed by the edge, so as not to maintain another list, but this doesn't map 1-1 just yet
* Hierarchy table, this is encoded in incoming/outgoing edges and edge types in the language we've been using for this RFC, again it's the same information but we might want to normalize the language

grondo:
> Give me X resources for n minutes or (X / 4) resources for (n * 4) minutes

I think this was the case @garlick had mentioned above. I had suggested, for extensibility, we promote `walltime` key to dictionary with each key supporting a different method for communicating time limit, e.g. strict `duration`, `range` similar to resource count range, or `function` as a function of task slot count or other generic field of actual resources allocated. If there is some future method for communicating walltime someone can just add a new key, instead of modifying the definition of the `walltime` value.

trws:
I like the idea of making it a dict so it's easier to extend later. Especially to support a user-defined function or expression embedded in it.

dongahn:
@grondo: making it extensible makes sense to me. I think duration, range and function already covers a wide range of cases.

grondo:
> 4 requires basename, name and id such that name is basename-id

Actually, it should not require basename and id, basename is is optional and defaults to the type name (e.g. "core" for a "core", but a node can have a basename of "hype" for instance), name defaults to `"${basename}${id}"` but can be overidden with an explicit name.

We should have a way in the jobspec language to request (and exclude!) a set of resource by id or name.

> properties == attributes, we can just change the name of the key or say they're the same and this one works

The important distinction in RFC4 is that properties are a shared attribute of a common resource type, inherited by all instances, and attributes/tags are specific to an instance of a resource type. For the query language that distinction probably does not matter. (Perhaps that is what you were saying above.)

In general, I like the direction @trws is going with normalizing the RFC4 resource hierarchy terms and the jobspec resource spec. This is actually getting quite close to one of our original goals of making the language used for resource queries and resource configuration the same (borrowed from ClassAd). We should be able to define "matcher" for each of the components of a hierarchical resource and make this part of the resource query language. (e.g. we should be able to query against `id`, `basename`, `name` or any defined resource component) In fact, a resource definition in this language (e.g. the serialized version proposed by @trws, should also be a valid match for that exact resource)

Using this spec as the serialization language seems like a very good idea, and it would be interesting to explore that after we've settled on the basic components here (e.,g. merge this PR) ;-)

(Sorry if I kind of got off on a tangent here)

dongahn:
BTW at the expense of making noise, i sort of see commonality between extensible walltime spec and task shape spec. 

Do we want to make it so that task count/shape spec extensibl also? Right now we support task per slot and total bur later we might want to extend it to cover some odd shapes?  A task counts list or some mapping function etc which conventional distribution policy and count cannot easily create the shape?

To be clear, i am not suggesting to specify these now to make it extensible for later use.

grondo:
> Do we want to make it so that task count/shape spec extensibl also?

I like that idea. Are you thinking of making `count` a dictionary, initially supporting `total` AND/OR `per_task`, with other keys reserved for future or custom extensions?

I actually find that more readable and powerful. Win Win!

dongahn:
> I like that idea. Are you thinking of making count a dictionary, initially supporting total AND/OR per_slot, with other keys reserved for future or custom extensions?

Exactly. (Sorry for the typoes above; fat fingers on my smart phone.)

grondo:
I'm trying to keep track of changes we're proposing that affect this PR, so far I've got:

 - Update `walltime` to a dictionary with currently supported keys `duration` and `range`, other keys reserved for future or site specific extensions
 - Add resource `id`, `basename`, `name`, `attributes`, and `tags` keys to the resource query spec. 
 - Clarify "slot" as a full type, at least one slot SHALL be defined, slot type requires a `label` and `count`
   (@lipari, I think this may address your concern about slot counts, this gives each defined task slot a specific name and count)
 - In tasks, promote `count` to a dictionary with currently supported keys `per_slot` and `total`. Other keys reserved for future extensions

Anything missed? 

I still feel like we need a way to allow a list of alternatives (logical or), though we shouldn't have to support it yet. Perhaps that could be supported in the future with more `type`s under `resources`, (e.g. a `type: or` group?) As long as we think we can support it in the current form, I'm comfortable moving forward.

lipari:
> @lipari, I think this may address your concern about slot counts, this gives each defined task slot a specific name and count

Sounds good!

tpatki:
Shouldn't resources also have a range? Especially for flow resources like power/bandwidth? We don't have to do this for the current version but I think we'll need it in the future.

For example, give me 100 to 120W of power per socket with a performance range of say 10 min to 8 min (in scenarios where for the application giving more power corresponds with running faster).

grondo:
@tpatki: Yes resources have an explicit range now `count: { min: , max, operator, operand}`, but I'm wondering if we should just make the resource count for the query into a dictionary like @dongahn suggested for tasks `count`.

dongahn:
> We should be able to define "matcher" for each of the components of a hierarchical resource and make this part of the resource query language. (e.g. we should be able to query against id, basename, name or any defined resource component) In fact, a resource definition in this language (e.g. the serialized version proposed by @trws, should also be a valid match for that exact resource)

This is a tangential topic, but in terms of well-known design patterns to look at, compilers are all about visitor and matching patterns. For example [clang's AST matchers are here](http://clang.llvm.org/docs/LibASTMatchersReference.html). 

BTW, it will be internal to our implementations (i.e., not part of our spec), but we will probably want to provide visitor APIs as well. One of the visitor interfaces I have used in the past, which I though was relatively well design was ROSE, like Figure 7.6 of (this)[http://rosecompiler.org/ROSE_Tutorial/ROSE-Tutorial.pdf]. Clang/LLVM also supports visitor interfaces. 


We have converged to a general graph as a way to represent resources and their relationships so things will be a bit more complex but I thought these compiler design patterns are reasonable examples to look at.

grondo:
One more question for the general audience: Given we have `type: slot`, do we still need the general grouping `type: group`? (If we don't want "unnamed" slots it would seem the answer is yes)

BTW, it occurs to me that these reserved types `group` and `slot` imply resource types of the same name SHALL NOT be allowed. Is that ok?

dongahn:
> As long as we think we can support it in the current form, I'm comfortable moving forward.

@grondo: I suspect we can support this in the current form. My thinking on the bottom line is: If resource/program specs that are OR'ed are completely orthogonal, a canonical form will contain two distinct specs. A complex case will be two specs each specialize its slot different on the common resource spec. But that's an optimization we can worry about later.

grondo:
Just to clarify, current definition of jobspec has *one* `resources:` list, and each list item is implied to be ANDed together. In the current form there is no way to specify multiple resources sections for a single program. If I understand your point @dongahn, you are thinking OR requests might be handled by submitting multiple jobspecs with some higher level indication to the scheduler/matcher that only one of the submitted jobspecs should actually be run?

dongahn:
@grondo: Ahh. Actually I mistakenly thought a job spec can have distinct resource specs that can be ORed. 

>  If I understand your point @dongahn, you are thinking OR requests might be handled by submitting multiple jobspecs with some higher level indication to the scheduler/matcher that only one of the submitted jobspecs should actually be run?

But if multiple jobspecs are somehow supported this way, the net effect of being able to support orthogonal ORed requests should stay the same. I just wanted to have an assurance that we even have coverage for the bottom line case...

garlick:
That reminds me, do we need something in here for requires/provides arbitrary dependencies?

grondo:
> That reminds me, do we need something in here for requires/provides arbitrary dependencies?

I thought we had thrown that out when we decided canonical jobspec doesn't support more than a single program. All of the inter-job dependencies, array jobs, etc will have to supported at the non-canonical jobspec level.

grondo:
Ok, I pushed a few minor edits resolving all but one of the todo items above. Make sure I"m on the right track

trws:
I'm kinda popping these off the stack as I get caught up.

We will need expressions for dependencies, but haven't got any infrastructure for that yet nor have we defined how they'll work, so I'd hold off on that at this phase.

Having a way to do an OR is a good idea, but I'm not sure how that would be expressed.  Perhaps we could have a way to use dependencies to say "if this job runs, don't run me" or similar.  I think PBS/Moab actually has this in their dependencies, and I wondered why until now...

@dongahn, I completely agree that we'll need a visitor for working with the actual resource graph in any kind of sane way.  The matches will actually be a special-cased subgraph isomorphism check, but that will probably be implemented internally with something equivalent to a depth-first search visitor so it can be kept reasonably general.

dongahn:
> Having a way to do an OR is a good idea, but I'm not sure how that would be expressed. Perhaps we could have a way to use dependencies to say "if this job runs, don't run me" or similar. I think PBS/Moab actually has this in their dependencies, and I wondered why until now...

I like this idea so that we can cover both cases under a common infrastructure. 

My main question would be whether we want to flesh this out at this time or defer it. In terms of managing things a bit more tractable, my preference is to push forward the job dependency dialogue as a separate item than the current jobspec dialogue. I have to believe that is also @grondo's position...

grondo:
> Having a way to do an OR is a good idea, but I'm not sure how that would be expressed. Perhaps we could have a way to use dependencies to say "if this job runs, don't run me" or similar. I think PBS/Moab actually has this in their dependencies, and I wondered why until now...

Good point, however this scheme only lets you define the total resource request as a set of alternatives (really alternative jobspecs). While I think that feature should certainly be supported, it precludes the use of OR within the hierarchical query, e.g. I need two nodes, each of which has 2 sockets OR 4 sockets. Definition of a group that is OR instead of AND might be a way to allow this at any level in the query, which could greatly simplify some requests.

Whether we actually *want* to do that is a different question. I'm willing to kick the can down the road a bit on that one, as I feel more confident what we're doing now doesn't preclude either of these options.

grondo:
Well, it just occurred to me that a range can be decomposed to, or represents, and OR of a set of alternatives. So, very bad example in my comment above since it could be expressed as a request for node with socket with a count `{ min: 2, max: 4, operator: "+", operand: 2 }`

Maybe that actually covers most of the OR cases I was thinking about, besides attributes, tags, etc which could have their own query language that includes "OR".

trws:
I could see adding an edge type or qualifier of some kind that's only valid in matchers for something like "one-of" where it would then take a list of resources but resolve to only one of those things, probably the first that matches or something like that? I'd leave that out for now, but it doesn't seem like it would be a terribly difficult extension if we decide we need it later.

garlick:
Since we are talking about reusability and saving jobspec in the provenance record, should we add a top level "version" key?

trws:
That's an excellent point @garlick. 👍

garlick:
As discussed with @grondo, here is a stab at [JCR](https://tools.ietf.org/html/draft-newton-json-content-rules-06.html) for the above, very rough, just to see how it goes:
```
{
   "resources" : [ 1*resource_vertex ],
   "tasks" : tasks,
   "walltime" : walltime,
   "attrs" :  { /.*/ : any },
   "version" : 1,
} 

resource_vertex {
    "type" : string,                         
    "sharing" : ( :"shared" | :"exclusive" ),
    "count" : resource_count,
    "id" : string,
    "uuid" : string,
    "tags" : resource_tags,
    "attributes" : resource_attributes,
    "with" : resource_vertex,
    "edge" : TBD,
    ?"unit" : string,
}

resource_count {
    "min" 0..,
    "max" 0..,
    "operator" : ( :"+" | :"*" ),
    "operand" : 0..,
}

resource_tags {
    ?"list" : [ 1*string ],
    /.*/ : any,
}

resource_attributes {
    ?"list" : TBD,
    /.*/ : any,
}

tasks {
    "command" : [ 1*string ],
    "slot" : string,
    "count" : { "per_slot" : 1.., "total" : 1.. }
    "attrs" : { /.*/ : any },
    "distribution" : string,
}

walltime {
    "duration" : string,
    "range" { "min" : 0.., "max" : 0.. },
}
```

We have several different "attributes" object defintions - it would be nice to settle on one that works for tasks, resources_vertex, and attrs.  The description of resource attributes was a bit muddled IMHO though I may have missed some discussion that led us there.

Also, if this is version 1 of the canonical jobspec, could we drop the allowance for future keys in the resource_attributes and resource_tags definitions, and assume we'd add those in a future version?  (possibly that assumption could simplify prose as well).

spec_14.adoc
```diff
@@ -95,7 +95,160 @@ FIXME: Fill in
...
+A resource vertex MAY additionally contain one or more of the
+following keys
+
+ *unit*::
+ The `unit` key, if supplied, SHALL have a string value indicating
+ the chosen units applied to the `count` value or values.
+
+==== Reserved Resource Types
+
+*group*::A resource type of `type: group` SHALL indicate an anonymous
```
lipari: missing a space between :: and A.  It cancels the elevation of "group" to a term.

grondo: Thanks!

grondo:
> As discussed with @grondo, here is a stab at JCR for the above, very rough, just to see how it goes:

Nice, thanks, this is helpful!

> `   "resources" : [ 1*resource_vertex ],`

Can we use `+` instead of `1*`? I find that clearer, but may have missed some  difference between the two when reading the JCR RFC.

> `    "operator" : ( :"+" | :"*" ),`

We should also add ` | :"^"` ? -- I think this was  a direct example from @trws 

>`    "operand" : 0..,`

Should this be `1..`? (I can't think of where an operand of `0` makes sense, but I'm sure you had a reason for this.

> `    "count" : { "per_slot" : 1.., "total" : 1.. }`

Does this require `per_slot` AND `total` be defined? Would this be better as `"per_slot":: 1.. | "total": 1..`? 

> We have several different "attributes" object defintions - it would be nice to settle on one that works for tasks, resources_vertex, and attrs. The description of resource attributes was a bit muddled IMHO though I may have missed some discussion that led us there.

Very good point. I don't like the term `attrs` as applied here, but didn't change it yet because I don't really have a better name. Summing up, here is what I *think* we want to be able to do

 * Currently, the only exposed program/job "attribute" or option is the `walltime` -- I guess because that is required for scheduling, so it was promoted to a position of importance. All other possible non-resource related options would fall under what is currently `attrs`, including things like `account` IO options, plugin related options, (though these could go under `tasks` I suppose), notification options, etc, etc. I guess since this is a free-for-all it is currently ill defined. So I think the current top-level `attrs` is just a place to store various values the scheduler might not care about... I keep thinking `options` might be a better name, but that doesn't quite fit either.

* The `attrs` key for tasks serves a similar purpose as above, generic catch-all, place for plugin options, etc. Maybe both could be described as a dictionary for program data outside the scope of this RFC?

* At one point I had though RFC4 made a distinction between resource *type attributes*, and resource *instance properties*. A type of resource has attributes (number of sockets, model #, etc.) and a resource instance inherets those attributes, plus has instance properties (serial #, location, etc.). However, I don't see attributes discussed in RFC4 so I guess that isn't the case anymore. For now the language settles on static properties per *resource instance* so I guess we should follow that here. The jobspec language therefore needs a way to match on these properties (which may have values) and this is what the `attributes` is for. Perhaps we should rename that to properties, or update the language in RFC4?

* Tags are dynamic, valueless strings attached to instances of a resource. We also need a way to match a set of tags, and either a list of required tags `[ "foo", "bar", "baz" ]` which is assumed to be an AND list, or a gmail/xapian style query string `"foo OR bar (NOT baz)"` might be required at some point.

> Also, if this is version 1 of the canonical jobspec, could we drop the allowance for future keys in the resource_attributes and resource_tags definitions, and assume we'd add those in a future version? (possibly that assumption could simplify prose as well).

That is probably fine. I was trying to avoid having to write "Version 2" of the jobspec definition if someone wanted to implement a scheduler or resource matching algorithm that took a custom query  language (e.g. xapian example above). If we think we only need support for AND for now, then we could drop that language. (Note I was thinking of the use case of Slurm's `--constraint` option)

grondo:
> ```
walltime {
    "duration" : string,
    "range" { "min" : 0.., "max" : 0.. },
}
```

Should this be
```
walltime {
    "duration" : string |
    "range" { "min" : 0.., "max" : 0.. },
}
```

Honestly I don't know JSON Content Rules syntax that well, so it might not do what I think it does....

**Comment by grondo** [comment]
> ```
resource_attributes {
    ?"list" : TBD,
    /.*/ : any,
}
```

From [jcr-version 0.6](http://www.ietf.org/id/draft-newton-json-content-rules-06.txt) S6.2:

> By default, members of objects which do not match a rule are ignored.

Therefore, I think we can drop any rules of the form
```
 /.*/ : any
```

From the spec. Hopefully that somewhat alleviates your complaint about reserving "future keys"?

However, in cases where extra keys are *not* allowed, e.g. perhaps the program itself (?), we may want to add the rule
```
+ @{reject} /.*/ : any
```
as suggested in the draft rfc to reject any unspecified keys.

garlick:
@grondo feel free to make all your changes and if appropriate, add to PR.   They all sound fine, and I am not attached at all to the version I posted.  It does seem a good exercise to go through - a sort of OCD desire to refactor the "protocol" seems to kick in after a while, a good thing I think :-)

grondo:
Yes, I am finding it and especially your starting point very helpful. Just trying to keep the proposed "changes" out in the open!

grondo:
Note I did stumble across the following *other* draft rfc related to JCR: [Co-Constraints for JSON Content Rules](http://www.ietf.org/id/draft-cordell-jcr-co-constraints-00.txt), which might be useful in the future to even further drill down on content constraints.

grondo:
Ok, I quickly added @garlick's JCR draft to the RFC, then added some revisions of my own on top. I attempted to split out the different "vertices" `slot` and `group`, but I kept the commit separate because I may have failed miserably!

This was quickly done so there may be some typos too

lipari:
Flow-wise, I think the JSON content rules appear too early.  The _Jobspec Language Definition_ language section introduces the YAML format, and before any of the resource and task terms are mentioned, and before YAML is even discussed, the JSON content rules are displayed.  I suggest the following order:  resource/task terms -> YAML representation -> associated JSON representation.

grondo:
@lipari, thanks that is good feedback, I kind of threw the JCR ruleset in there without thinking.

Are you thinking it would be enough to move the JCR to after all the definitions, in its own subsection?

In your suggested order, by "YAML representation" do you mean the definition list of "dictionary" keys and their required form?

Thanks!

lipari:
The jobspec definition section starts with the resource/task building blocks, then shows how those blocks fit together to create a spec (i.e., the canonical jobspec YAML document).  Then, the JSON equivalent following the content rules.  Then the examples we worked on in the use cases [PR](https://github.com/flux-framework/rfc/pull/52).

grondo:
I may still be confused. YAML is a superset of JSON, so the JCR is meant to be a ruleset applied to a canonical jobspec YAML document (after perhaps converted to JSON form).

However, I'll move the ruleset to after the keys definitions which I think satisfies your suggestion (or we'll see anyway!)

lipari:
OK. I understand.  And moving the JSON to be after the terms are introduced flows better for me.  Thank you.

grondo:
Would a small example help after all the keys definition and before the Content Rules section?

lipari:
> Would a small example help after all the keys definition and before the Content Rules section?

Yes. I think it would - a simple snapshot to show how it all fits together.  Then, mention the fact that the job spec as portrayed in YAML needs to conform to the JSON content rules (that follow).

grondo:
Good idea, thanks @lipari! I'll push some changes a bit later today.

garlick:
> Would a small example help after all the keys definition and before the Content Rules section?

Good idea IMHO.

Is this PR getting close to mergeable?  We can always fine tune it in a new PR as the examples are reworked.

grondo:
> Is this PR getting close to mergeable? We can always fine tune it in a new PR as the examples are reworked.

Yes, I think so. Let me just do some minor fixups and add the canonical example. While writing up the example, the following things came up (so far)

 * Would it be ok to have some keys in resource vertex be optional that are presently required, e.g. `id` `uuid` and perhaps even `attributes` and `tags`? I could go either way whether requiring all keys be defined in canonical spec is good or  bad.

 * Does it make sense to have `sharing` on `type: group` and `type: slot`? At first I thought so, but now I'm not sure since this key is currently required at every resource vertex.

lipari:
I lean toward making id, uuid, attributes, and tags optional.  And I agree that having sharing at every resource vertex precludes its value at the group level.

trws:
On Jul 22, 2016, at 11:03 AM, Mark Grondona <notifications@github.com<mailto:notifications@github.com>> wrote:
Would it be ok to have some keys in resource vertex be optional that are presently required, e.g. id uuid and perhaps even attributes and tags? I could go either way whether requiring all keys be defined in canonical spec is good or bad.

Yes.  I really think we want to make this three pieces that build on each other.  There's a resource spec, as in RFC4, a jobspec, and a resource matcher, which is much of what we're specifying in the jobspec.  Since the resource to be matched must match everything specified, they either have to be optional, or required to be present but nil, and I don't see any particular benefit to the latter personally (since they will just become nil on deserialization anyway).


  *   Does it make sense to have sharing on type: group and type: slot? At first I thought so, but now I'm not sure since this key is currently required at every resource vertex.

I'm not sure we need type group at all anymore, honestly I think I'd just remove it.  As to slot, having sharing there seems useful even though it itself isn't a resource... It's a useful key for whether a container should apply across its constituent parts or only in tighter bounds below that... Or at least that's what seems to make sense right now, thoughts?

grondo:
> I'm not sure we need type group at all anymore, honestly I think I'd just remove it.  As to slot, having sharing there seems useful even though it itself isn't a resource... It's a useful key for whether a container should apply across its constituent parts or only in tighter bounds below that... Or at least that's what seems to make sense right now, thoughts?

I'm not sure I follow what you mean by "a container should apply across its constituent parts or only in tighter bounds below that", I had thought that `type: slot` would be used to specify what goes in to a container. Can you give a quick example what you mean?

I'm actually still not sure I understand why `sharing` is specified in the jobspec, as it seems like the number of "shares" of a resource should be determined by its children. I've probably got a naive view, but it seems like if you want to allocate a thing exclusively, you should specify that you need it and all of its children (maybe that is what `sharing: exclusive` means here?) As applies to a leaf resource, a virtual "share pool" could be attached as a child to simulate a sharing level without special casing certain resource types.

Sorry, I'm getting a little off topic, but it sounds like you might have a different idea for what "sharing" would mean for a task slot, and if so my opinion would be to choose a different keyword for that.

grondo:
>  they either have to be optional, or required to be present but nil, and I don't see any particular benefit to the latter personally (since they will just become nil on deserialization anyway).

Not to belabor the point too much, but there are two specific reasons I thought we might want optional fields to actually be specified in the canonical jobspec

 * It simplifies the *reader* code a bit, i.e. you remove a lot of conditionals in the reader code that go something like `if not jobspec.uuid then jobspec.uud = default` or `if jobspec.uuid then do_match_uuid()`. That is, you can always depend on values being set and *not* nil.

 * If unset keywords are replaced by defaults, then a "canonical jobspec" could actually mean different things in different versions of Flux. Now that we are using a versioned spec, this might be a non-issue.

Even given those two points, I still agree that specifying AT LEAST `id` and `uuid` are optional and "match any" if not set. For `attributes` and `tags` I'm less certain, but they can each be treated as empty set/dict if missing, which should also "match any" if not set.

lipari:
Let me give a couple examples where I think sharing will be useful.  We've already said, a request could be for just one node.  With just that spec, we might grant the job a default child of that node (e.g., a core and some memory).  This default behavior could be overridden by asking for a node "exclusively".

We could assign a size value to a node and require the job spec to request an "amount" of that size.  Jobs that request all of the size elements of a node would get it exclusively.  That seems awkward to me.

Second example is the job that wants 4 nodes in the same rack vs. just a rack and everything in it.  To get the first, the rack spec would be shared.  To get the second, the job spec would be for a rack exclusively.

grondo:
> We could assign a size value to a node and require the job spec to request an "amount" of that size. Jobs that request all of the size elements of a node would get it exclusively. That seems awkward to me.

In terms of the resource graph query, I was considering that if a user specified they require a "type: node" then it is assumed they want a whole, unshared node. Any parent of "node" is clearly shared in this case (e.g. a rack in your other example). If a user requests a "type: socket" as their leaf resource, then it is assumed they want a whole, unshared socket, but unless there is only one socket on the node, then the node, rack, cluster, etc. are shared. 

How does the binary shared/exclusive system work in terms of the hierarchical model described above? Like you say, you could "translate" a jobspec resource request of a single node into a core and some memory, or in the case of `sharing: exclusive`, you would somehow translate the request into "any node and all of its children". I guess where I'm confused is how to reflect that in the canonical job spec, which should express these defaults.

I would currently lean toward removing "sharing" from the canonical jobspec and instead require that *actual* resource request be reflected in the canonical job spec, reflected in the resources at the leaf of any resource request (if that makes any sense). For instance, in the example above a request for one node must either include a request for a specific set `with:` child resources, or the request *only* SHALL match a whole node and all of its children.

In the non-canonical job spec, the `sharing` level may be specified and MAY affect the application of defaults to non-leaf resources (such as a request for a shared node SHALL be replaced by a site defined default number of cores and Memory.) In this way, a site can clearly define any level of sharing that they wish to support on any level of resource, without adding any special "sharing" or "shares" attributes to specific resources.

I'm sure I missed something, and if I'm way off base here I'm willing to drop this line of reasoning for now in order to move on, but I feel like *I'm* still confused on the purpose of a "sharing" attribute in the heirarchical resource model.

lipari:
> I would currently lean toward removing "sharing" from the canonical jobspec and instead require that actual resource request be reflected in the canonical job spec, reflected in the resources at the leaf of any resource request (if that makes any sense).

@grondo, you make good arguments and I am close to agreeing with you.  So let's consider the rule that anytime a hierarchical resource is specified, we only grant the leaf resource exclusively, and all parent resources would be potentially shared.  How would we differentiate between a request for an exclusive allocation of node that has at least two sockets from a request for two sockets on the same node?  In the first case, we would set the sharing flag to exclusive at the node level.  In the second case, it would follow the above rule and the node could be shared if it had more than two sockets.

grondo:
Excellent point @lipari, thanks!

trws:
> Even given those two points, I still agree that specifying AT LEAST id and uuid are optional and "match any" if not set. For attributes and tags I'm less certain, but they can each be treated as empty set/dict if missing, which should also "match any" if not set.

This is more or less what I meant.  These are all options for which the default is "Nil" or "None" or some similar in a jobspec resource matcher, though probably not in a resource spec that would populate the resource database, so having it set to nil or simply unset seems equivalent.



> @grondo, you make good arguments and I am close to agreeing with you. So let's consider the rule that anytime a hierarchical resource is specified, we only grant the leaf resource exclusively, and all parent resources would be potentially shared. How would we differentiate between a request for an exclusive allocation of node that has at least two sockets from a request for two sockets on the same node? In the first case, we would set the sharing flag to exclusive at the node level. In the second case, it would follow the above rule and the node could be shared if it had more than two sockets.

This is where we started, to some extent, but it's not quite right to say that it's the leaf resource.  Where I remember us last being is that everything under the level of a slot is exclusively allocated to that slot.  If we take that definition, then we don't need the "sharing" key anymore, because it can always be expressed by locating the slot correctly, but doing it by what the leaf resource is doesn't cut it.  Otherwise there's no way to specify that you want to exclusively allocate a node with four cores, because adding the cores would cause the cores to be exclusively allocated instead of the node.  Does that make sense?

grondo:
> Otherwise there's no way to specify that you want to exclusively allocate a node with four cores, because adding the cores would cause the cores to be exclusively allocated instead of the node. Does that make sense?

Yes, this was the point I was missing (sorry), and @lipari helped me see that above.

> Where I remember us last being is that everything under the level of a slot is exclusively allocated to that slot. If we take that definition, then we don't need the "sharing" key anymore, because it can always be expressed by locating the slot correctly, 

This idea is attractive, if it does work out, then I'm all for it. I'm finding the `sharing` key to be pretty awkward. (Especially if it is required at every level -- it is strange to say `type: cluster, sharing: shared`)

trws:
I agree, it's definitely clumsy.  The one part I'm not sure about is how we should handle resources exclusively allocated to the job, but not to an individual task slot.  Maybe allow extra slot vertices that aren't themselves actually used with tasks directly to set exclusivity farther up the tree?

lipari:
> Does that make sense?

I believe it be more understandable to me anyway if you would provide the YAML slot definitions for each of the two scenarios:   a request for an exclusive allocation of node that has at least two sockets and a request for two sockets on the same node.

trws:
> a request for an exclusive allocation of node that has at least two sockets

```yaml
type: slot
with:
  type: node
    type: socket
    count: 2
```

> a request for two sockets on the same node


```yaml
type: node
  type: slot
  with:
    type: socket
    count: 2
```

grondo:
> I agree, it's definitely clumsy. The one part I'm not sure about is how we should handle resources exclusively allocated to the job, but not to an individual task slot. Maybe allow extra slot vertices that aren't themselves actually used with tasks directly to set exclusivity farther up the tree?

Good point. Your idea here works pretty well, however I fear use of `type: slot` here would be confusing to users and could cause problems down the road. I do not have a better idea though, but maybe this is a good case to bring back `type: group` as an unlabeled slot? Alternately, it is good reason to keep an optional `sharing` keyword, though I'd say it would be more clear if it was an optional `exclusive` boolean placed on a vertex (and obviously inherited by `with:` type children).

trws:
I like the optional exclusive boolean idea, it shouldn't be needed very often, and if it is, it will get set.

grondo:
@trws's case of resources exclusively allocated to a job but not part of any task slot makes a great use case. However, I'm having trouble coming up with a non-trivial example, like a license which is obviously exclusively allocated since it will be a leaf in the request graph.

Any ideas of a use case requesting a hierarchical resource exclusively, but not part of any task slot? (I guess a node that doesn't run anything is an example, but anything better?)

trws:
@grondo, if we go by the resources under a slot are exclusive, then we wouldn't have the exclusive leaf resources anymore necessarily.  A license is actually the best case I can immediately think of.  Alternately, a user might want to exclusively allocate every node they run on, but confine their task slot to the socket level.

grondo:
Good point, shared/exclusive and the slot are really two different things, and I agree it does seem we need both. The slight changes we've kind of agreed on so far are indeed making it less awkward for me. 

Let me make sure I've got the rules correct:

 - Optional `exclusive: true` setting on resource marks that resource and all `with:` children as required to be exclusively allocated.
 - `type: slot` automatically assume `exclusive: true`
 - Leaf vertices in the request graph automatically assume `exclusive: true`

That works for me, but I realize the last point might still be in question. For me, it doesn't make sense to request a resource "shared" without saying what portion of that resource you actually need to allocate.

trws:
I agree with the first two, the last I'm less sure.  That said I'm not sure I have a good counter-example.  The only thing that's bugging at me is it seems possible, despite my present inability to come up with an example, that a user would want to request that the structure of the graph contain something at a leaf location that they are not actually allocating.  Actually, maybe that's the issue.  If it's limited to leaves of the `with:`-link-based hardware tree, it seems to work, but if it's leaves of the request *graph* then it would become problematic?

grondo:
>  Actually, maybe that's the issue. If it's limited to leaves of the with:-link-based hardware tree, it seems to work, but if it's leaves of the request graph then it would become problematic?

That is what I was thinking, too. My rules above only apply to the `with:` tree. We haven't fully specified how a generic graph query would look, so I'm having trouble wrapping my head around one at all.

My problem with *not* requiring leaf vertex in `with:` tree to be exclusive is that it does seem like the system has to "fill in" the missing children when they are unspecified, and the parent is "shared". E.g. a request for a node "shared" makes no sense, as you have to allocate *something* from that node, whether it is one core and some RAM, or 1 "compute share" -- and that should be reflected in the canonical jobspec IMO.

However, I realize it is taking me awhile to get some of this stuff, so I'm willing to admit I may be wrong.

trws:
That's actually not quite how I see it.  If the thing is shared, it has 
to exist, but nothing from it is actually allocated at all.  It serves 
as a requirement on the system structure, but contributes nothing to the 
list of things that the scheduler actually has to consider for 
allocation.  Perhaps this is actually a third state we haven't actually 
talked about where resources that are shared but partially consumed, 
like pools of memory for example or nodes allocated shared, are distinct 
from resources where the allocation is just none.  That's what I'm 
thinking of when I think of shared resources that are leaves outside a 
slot, they're relevant to the query as structural requirements, but 
ignored for allocation purposes.

grondo:
Ah, I think your conceptual view of resources differs from mine. In my mental model (of the `with` hierarchy at least), every resource that can be allocated is a composite. So a node is a composite of its memory, sockets, cores, gpus, etc, if any of these children are allocated or consumed, the node itself is partially consumed, as is the cluster which is a composite of all its nodes.

Your mental model is more what I was thinking of for something like the "topology" graph, which as you say could have entities in it which are non-allocatable, and which could affect the structure or relation of the resulting allocated resources -- but I would hesitate to call these things "resources" if they cannot be allocated by the resource manager.

So I do agree with you that a leaf vertex in the topology or other strucutural matching case does not imply exclusive. However, if the resource is non-allocatable then the exclusivity of leaves could just be ignored?

Also, I still wonder if a request for a single shared node by itself makes any sense, as the request has no "shape".

trws:
> Also, I still wonder if a request for a single shared node by itself 
> makes any sense, as the request has no "shape".

Maybe a good way to deal with this is to say that a slot must have at 
least one "with:" child, and that's what's exclusively allocated.  That 
way, there's no way to request a single shared anything without at least 
one exclusive resource?

tpatki:
@grondo 
I'm not sure I'm following. Why doesn't a request for a single shared node by itself make sense? 

As a user, there can be a scenario when you want to allocate say a core/set of cores to yourself because you don't need the full node, right? 

I think the question is what does the default graph look like if someone doesn't specify a "with" clause (one core and some memory?)? Or making a "with" clause mandatory in that scenario as @trws suggested.

grondo:
> Maybe a good way to deal with this is to say that a slot must have at
least one "with:" child, and that's what's exclusively allocated.  That
way, there's no way to request a single shared anything without at least
one exclusive resource?

That works for me, and actually seems similar to what we were discussing before.

grondo:
> I'm not sure I'm following. Why doesn't a request for a single shared node by itself make sense?
> 
> As a user, there can be a scenario when you want to allocate say a core/set of cores to yourself because you don't need the full node, right?
> 
> I think the question is what does the default graph look like if someone doesn't specify a "with" clause (one core and some memory?)? Or making a "with" clause mandatory in that scenario as @trws suggested.

Yes, I think we're saying the same thing. It may have been incorrect to say it "doesn't make sense",  but rather a request for one shared node *alone* is incomplete, and therefore should not be allowed in canonical jobspec.

tpatki:
Another question: what if the user gives a request that can't be translated successfully into a "shape"? Do we have a mechanism for addressing this yet -- do we just throw an error or do we default to a basic setup (for example, if you request a shared node and forget the with clause)? 

I don't know how SLURM or current resource managers address this.

trws:
@tpatki It depends on the type of issue.  If what they specify doesn't follow the spec, it will get rejected with an error by the parser.  If it's something we can determine can't be supplied as part of the resources that can reasonably be made available, they'll get an error or a long-waiting job depending on policy.

grondo:
OK, in this PR I've update the description of the `tasks` `slot` key to allow either `label` (explicit task slot) or `level` (implicit task slot) as suggested by @trws. I quickly updated the example, and actually added another simple example using the implicit task slot.

grondo:
If we're getting close to ready to merge this, I can squash down all the incremental work. I think maybe there is a lot of editing to do, and the spec language definition could be presented better, but perhaps we could do that as a future PR

trws:
I'm good with that.  It seems like we're at the point where it doesn't really make sense to do much more tweaking without trying it out and seeing what happens.

grondo:
Ok, squashed!

trws:
Very nice, thanks for all the hard work on this @grondo!

grondo:
Thanks!!

PR #59, opened 2016-10-04
rfc4: Fix italics formatting error

chu11:
There appears to be a formatting error here.

> Allocate (N, S):: Allocate _N_ resources from the pool under the name _S. The available resources in 
> a pool is its size minus the total number of allocations. The allocation _S SHALL be stored as a 
> searchable attribute along with the resource for later use with _Find_ and _Match_ methods. If an
> allocation under _S already exists, then the allocation SHALL be grown by amount _N_.

I don't think the variable ```_S``` or method ```_Find``` is intended.


garlick:
Yep.  Thanks!

PR #60, opened 2016-10-04
rfc4: Fix variable name in description to match method prototype

chu11:
Description states ```string``` while prototype lists ```S```.


garlick:
Thanks!

PR #64, opened 2016-10-05
rfc14: Fix paragraph identation

chu11:
I don't think this paragraph is supposed to be indented further, I think it's supposed to be at the same level as the previous paragraph.


grondo:
Yes I believe you are correct. Thanks!

PR #65, opened 2016-10-11
Add pdf make target for all docs

morrone:
Add pdf make target that use a2x to convert asciidoc files
into pdf files.

a2x is using docbook, and docbook does not not like empty
sections or sections named "References".  We ad "TBD" to empty
sections to make them non-empty, and explicitly declare the
sections type as "[sect2]" for "== References" sections.

Note that an alternate way of dealing with the "References"
section name problem is to convert to using actual footnotes
and/or bibliography support in asciidoc.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,28 +1,37 @@
-HTML = \
-	README.html \
-	spec_1.html \
-	spec_2.html \
-	spec_3.html \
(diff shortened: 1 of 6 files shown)


garlick:
This is fine for now.  Actual references would be nice down the road though.

PR #67, opened 2016-11-08
rfc11: add hdir object type

garlick:
See commit description.  This is a proposed change to the proposed KVS metadata format which I think may address the large directory performance problem we've observed and begun to work around.

Note that to try this out directly, the KVS will need to be converted the proposed format.  I will try to think of a way to test this idea out sooner using the old metadata format, but I'd be interested in any feedback on the general approach.

### DIFF
--- a/spec_11.adoc
+++ b/spec_11.adoc
@@ -100,11 +100,12 @@ Hash buckets MAY be sparsely populated.
   "val":{
     "size":8,
     "func":"city32",
-    "bucket":{
-      "0":{"ver":1,"type":"treeref","val":["sha1-aaa"]},
-      "6":{"ver":1,"type":"treeref","val":["sha1-eee"]},
-      "7":{"ver":1,"type":"treeref","val":["sha1-fff"]},
-    }
+    "bucket":[
+      {"ver":1,"type":"treeref","val":["sha1-aaa"]},
+      ,,,,,
+      {"ver":1,"type":"treeref","val":["sha1-eee"]},
+      {"ver":1,"type":"treeref","val":["sha1-fff"]},
+    ]
   }
 }
 ----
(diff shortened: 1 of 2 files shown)


grondo:
This seems like a very clever idea!

Do you want to call out what SHOULD or MAY occur if the `size` of an `hdir` shrinks back below `maxdirents`? I don't know enough to guess if it would be worthwhile to convert back to a `dir` object, but for now it is an open question.

Would the actual conversion from `dir` to `hdir` happen at the time of `put` or during the `commit`? Sorry that one should be obvious. It will be interesting to probe the performance characteristics of the approach in the end, but I think it will obviate the need to worry about directory sizes, perhaps completely!

garlick:
I think we can leave those details out for the moment pending an experiment, but will definitely want to cover that stuff eventually.  In the KVS, an hdir would be created during commit processing (when the size of the directory can be determined).  However the RFC also applies to offline processing of metadata, through direct content store access.

It looks like JSON arrays can be sparsely populated so I'll probably push an update to make the hash buckets an array rather than an object with keys of "0", "1", etc.

grondo:
> It looks like JSON arrays can be sparsely populated so I'll probably push an update to make the hash buckets an array rather than an object with keys of "0", "1", etc.

Hm, are sparse JSON arrays defined in the standard, or is it implementation dependent? The standard just says " An array structure is a pair of square bracket tokens surrounding zero or more values. The values are separated by commas." I guess if an empty, non-quoted string is a value then something like `[,,,"foo"]` is a valid JSON array, but I wonder if other implementations will trip over that?

I'm sure I'm missing something, what are the other benefits of using sparse array over an object? I kind of wonder about the performance impact of one vs the other, if there is even one? A sparsely populated object would seem to beat a sparse array on space efficiency of the textual representation, if that even matters here.

Sorry if the answers here should be obvious!

garlick:
Sorry, that was sloppy.  Indeed [JSON lint](http://jsonlint.com/) doesn't like that.

I guess we could set unpopulated slots to `null` but that increases the size of the object unnecessarily.

I thought arrays would be a little more efficient than a dictionary since in the dictionary case a computed integer index must be converted to a string so it can be looked up.  But actually that can be done in a lightweight manner.  I'll revert that patch, thanks for the oversight.

grondo:
OK, merging this now

PR #71, opened 2017-01-26
Replace the "bootstrap instance" term with "system instance".

morrone:
We want to leave the term "bootstrap" as a verb that descibes
the process of starting up process, providing their initial
configuration information, and performing network discovery.

Instead we call the long-lived multi-user instance that is one
of the primary parent instances in cluster or center a "system
instance".

Since exact details of the "system instance" architecture are still
a hotly debated topic (e.g. whether or not the system instance spans
both management and compute nodes as a single instance), we leave
the "Bootstrap Mechanism" of spec 8 to only descript bootstrapping,
not system instances.  We can describe system instances later
when we know what exactly they are.

### DIFF
--- a/spec_12.adoc
+++ b/spec_12.adoc
@@ -46,14 +46,14 @@ Flux runs in single and multi-user mode.  A single-user Flux instance
 only launches work as the instance owner.  A multi-user Flux instance
 launches work as other users.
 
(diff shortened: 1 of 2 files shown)


garlick:
Excellent, thanks.

PR #73, opened 2017-02-03
RFC4 reorg

lipari:
Reorganize the structure of the Flux Resource Model to more closely mirror the requirements as they have evolved over the last couple years.  This helps prepare the way for migrating flux-sched/resrc into a standalone flux-resource project.

Focus the scope of this RFC by removing the specs to define resources requests.  Requesting resources will involve [RFC 14 - Canonical Job Specification](https://github.com/flux-framework/rfc/blob/master/spec_14.adoc) and will probably merit a dedicated RFC.


grondo:
This looks good lipari, thanks!

lipari:
Thanks!

PR #74, opened 2017-02-05
update security model with per-message userid and capmask

garlick:
This PR adds _userid_ and _capmask_ fields RFC 3 messages, and discusses how they are to be used in RFC 12.

Incomplete and not for merging yet, as I wanted to get a bit of feedback on the "overview" rewrite in RFC 12 before adding details to that RFC.  Perhaps it goes too far though - let me know if I should leave out the bits about deferring setuid/containment to privileged executable.


grondo:
> Services that handle requests must perform a capability check on each request message, and accept or reject the request depending on the result.

The *must* here might be too heavy-handed. Aren't there cases where a uid check will suffice? e.g. a non-owner uid sends a signal to their own job?

I'm also wondering if we might need a more flexible system of capabilities. Are only flux-core services going to be able to register new capabilities, or should any service be able to dynamically register and use capabilities (e.g. a scheduler might offer external users a capability to adjust job priorities)?

In the latter case we might not be able to come up with an exhaustive list of capabilities, and would they eventually exceed a total of 32? We might want to think about how to make capabilities extensible. I hate to say it, but if most messages have only one or two attrs would it make sense to use a text list of attrs instead of a bitmask?

garlick:
You're right, not a "must" there.   Will fix.

On the limit of 32 attributes:    I'm kind of on the fence on this one!   It's a good point that the limit seems out of sync with the plan of allowing arbitrary services to be built on the flux framework.  On the other hand, it seems like there's a usability argument to make capabilities somewhat coarse grained.  When I envision a tool an instance owner would use to grant capabilities to people, it would seem onerous to make them manipulate a ton of fine grained capabilities; wouldn't we instead have them assign "roles" to users and map _those_ to capabilities for them?   Or should capabilities reflect roles in the first place?  (FLUX_CAP_SYSADMIN is kind of a role).  Hmm, not sure.

garlick:
Actually, I can readily see that even if we are successful in keeping the number of capabilities low, there still might be an external project that wants to add one without modifying flux-core.  Hmm.

Could we add a separate "named capability" feature to the capability db so that a service could check if an arbitrary capability is set for a given userid?

grondo:
> Could we add a separate "named capability" feature to the capability db so that a service could check if an arbitrary capability is set for a given userid?

That would seem to functionally work and I do not have a better idea. In fact, maybe instead of enumerated and named capabilities the "named capabilities" could be the "roles" you spoke of earlier, with hierarchical names grouped by service? (e.g. "sched.admin") (Just throwing ideas out now)

The only drawback I see is that the message uid will have be checked against possibly global roles each time a message is received. Also, the dichotomy between in-core and out-of-core services bothers me somewhat.

Looking at this from another angle, what capabilities (besides CAP_SYS_OWNER or similar) do you see for services in core... I can't actually think of any presently, but perhaps just a case of the Sunday evenings..

garlick:
Hmm I dunno, maybe:

* FLUX_CAP_SYSADMIN - owner (or designee)
* FLUX_CAP_RUNJOB - allowed to run
* FLUX_CAP_KILLJOB - ability to kill (others) running job
* FLUX_CAP_EXPEDITE - ability to manipulate queue priority
* FLUX_CAP_LISTJOB - ability to list (others) detailed job requests

Yeah I would really like to propose one mechanism and not a core one plus the moral equivalent of supplementary groups :-)

grondo:
Those seem like a good start, though I thought it at first a bit strange to have capabilities like `FLUX_CAP_EXPEDITE` which don't have any function in flux-core. It could probably be considered an obvious capability for a resource manager so perhaps that doesn't matter.

garlick:
I wonder if instead of "capabilities" we could switch to "roles" and "policy" and make things a bit simpler?

Roles could be quite general, hopefully enough so that the core could define all the roles needed by the system.  E.g.  (just off the top of my head): "owner", "admin", "hotline", "user", "nobody".  A user db could map users to roles, and these could be stamped on messages as described for capabilities.  There could be one role per user, or we could allow a set of them as with capabilities.

Policy could map roles to policy attributes.  There could be an instance policy db, namespaced as @grondo suggested.  For example a "sched.queueadmin" policy attribute could be assigned to "admin" and "hotline".  Sched would need to test the role in each inbound queue admin request to see if the role has the "sched.queueadmin" attribute in the policy db (obviously policy caching required).

Or for performance, perhaps sched's policy could actually live inside sched rather than a centralized db?  Could we define some Flux API functions to allow a default policy to be established and modified at module load time so that site-local policy modifications would be expressed as command line arguments in rc1?  E.g. `flux module load sched --setpolicy hotline+queueadmin --setpolicy...`

garlick:
Just pushed an update that addresses the initial minor issues with the last version - spell check and "must" on service capability check.  I also described the association of capabilities with specific operations within a service as "policy" and alluded to it being site configurable.

Details are lacking but I think the description is more or less compatible with what I proposed in my last message except s/capability/role/ has not been done.  Feedback would be good - I hope I'm getting closer to something that could be expanded in the detailed sections?

dongahn:
I like the role/policy idiom. Because a plain user as the instance onwer can set different capabilites for other related users, being able to avoid the term like FLUX_CAP_SYSADMIN seems to have as a good merit! Whether sched will implement this through a DB or its code itself is implementation details at this point.But when all is said and done, there will be a lot of policies (job management, resource selection, backfill, other queue optimization parameters ...) so we will need a clean way to manage the complexity in the end ( a side point).

garlick:
Thanks @dongahn, a good perspective to keep in mind.  In this RFC I'd like to try to keep focused on the main abstractions for security (e.g. "policy" limited to "security policy") but we'll definitely want everything packaged up sensibly for users in the end.

dongahn:
Thanks. Yes, to be clear that's what I assumed when I liked your role/policy idiom.

grondo:
So I understand:

 * flux-core comes with static list of "roles" which can be assigned to connections, these are not extensible at runtime or by external flux framework projects.
 * If a user is not in instance user db, then a conformant connector shall deny connection. Otherwise, a capability mask is built from the list of "roles" assigned to the user, and this mask plus the userid are set for each message originating from the user connection.
 * Services must take action to handle messages without CAP_SYS_OWNER, in which case they are responsible for having an adjustable policy as to which capabilities allow which operations.

I'm still skeptical we could come up with a master list of roles that would work for all scenarios (I assume the roles must be static and local to flux-core since they have to be mapped to bitmasks). Additionally, I worry about complexity because this approach seems to have two lookups -- connector maps user to role and service maps role to policy. Would it be easier to just map user to policy? That is, I guess I'd still like to hear more arguments for why the "capmask" is essential here. If you have a caching policy service, then couldn't services just as easily map UID directly to policy, and then the caching policy service itself could group users by roles, which now become site-extensible?

Sorry if that is just confusing, and I'm sure you have valid arguments, I'd just like to hear them out.

garlick:
Right I think you've got it.

If we do cache the looked up user info at connect time, and stamp on each message, then we only have the "policy" to lookup in services.  If policy is co-located with services as proposed above, then that's just a local operation, not an RPC.

On the other hand if we don't stamp the messages, then we have to look up at connect time (to accept/deny connection) and look it up in every service used by that connection.  I'm presuming that's probably an RPC to the local broker, and after that the service would cache the results, and expire it temporally.

It seems clearer to me to do this once at connect time, and manage the caching by attaching it to the connection state within the connector.

I had been thinking (after posting above) that maybe "roles" could be extended as long as they are statically assigned to bits for the instance duration.   User-defined roles would be referenced from the user db and in policy modifications, but default policy constructed by a service would need to stick to well known roles.  The bitfield size in the message would limit how many could be added (e.g. 32 - well known ones), unless we resorted to maybe adding a frame to each message that contained a textual list, though that would be less efficient than setting bits in a pre-allocated bitfield.

If it seems like I'm trying to over think this at the wrong time, I could back off and just attach the userid in this PR and leave all the capability/policy stuff out for now.   We could play around with the user db and the default "reject non instance owners" behavior and see how that goes, then revisit?

grondo:
I may be the one over-thinking it as well. 

It may be that best approach is a static list of roles assigned to users at connect time. I get nervous about allowing roles to be extended could set up services for serious and subtle security flaws, it might be best for now to go with static roles. Unless roles can be looked up by name (which as you said is a non-starter for efficiency), and used by name within service, I worry greatly about roles being "mixed up" in out-of-tree flux services...

grondo:
I don't mean to leave this PR to languish, sorry! It does seem like we could merge it now as is, and make any adjustments to the details of how roles or capabilities translate to service policies in later updates.

@morrone did have a valid comment about how the userid domain of an instance maps to UNIX uids on which the instance is running. He made a good case that these be kept separate to future-proof against Flux instances that do run across systems with different UID domains. I'm not sure how and if that affects this rfc (which certainly does not require or imply that the userid in the user db match a unix UID or anything similar), but thought we should make the point clear somehow?

garlick:
No worries I wasn't really done.  Please don't merge.

Let me think about the multiple uid problem while I eat my lunch :-)  Do we think it's a requirement?  I have visions of krb5 cross realm authentication and NFSv4 id mapping, which are a little frightening.

grondo:
> Do we think it's a requirement? I have visions of krb5 cross realm authentication and NFSv4 id mapping, which are a little frightening.

I'm thinking we should put a little time up front deciding if multiple uid domain could be easily handled, but that a full solution is certainly not required for any near term milestones. First step I guess is to think through if even a single user instance over multiple UID domains could be handled easily ;-) (It certainly means that munge credentials can't be used across these instances)

It seems like if you treated the user and auth problems more like cloud services do, then users connect to instances with a "name" and a "credential" which authorizes and authenticates them as a specific user in the Flux universe. The specifics of the UNIX credentials on the system then become irrelevant, and (for multi-user cases) the privileged service of Flux can handle translating between the user credentials and local UID.

grondo:
I'm also hoping @morrone will chime in here with specific difficulties he mentioned with multi-uid-domain Lustre, which might shed some insight on what not to do.

garlick:
It seems like this gets complicated quickly.  In order to make progress with this design I'm going to propose that for the near term we assume one administrative domain for Flux and state what that means in the RFC.  If there's a specific use case I'm forgetting that makes this assumption a bad one to make right now, please remind me.

The only thing that bugs me a little is it might be possible to do something with idmapd to allow Flux to both piggy back on existing user mappings, and do so consistently with a shared NFSv4 file system.  This might be especially useful in a workgroup environment (e.g. workstations without consistent uids but with a shared NFSv4 file system).   However, it feels like too much of a diversion before we've nailed down a first cut.

I'm assuming full grid mode like [globus](https://www.globus.org/) is not something anybody is pushing?

garlick:
Just squashed the changes and reworked the overview in RFC 12 to reflect the discussion.  Also renamed capabilities to roles in RFC 3 and 12.  Note I am following the convention that the overview section is written informally, thus RFC 2119 language will start later.

If the overview looks OK I'd like to start adding some of the detailed stuff in this PR.

grondo:
> If there's a specific use case I'm forgetting that makes this assumption a bad one to make right now, please remind me.

I'll just reiterate that @morrone had mentioned significant engineering effort currently going into Lustre to undo this assumption. Perhaps he'll eloborate here.

> The only thing that bugs me a little is it might be possible to do something with idmapd to allow Flux to both piggy back on existing user mappings, and do so consistently with a shared NFSv4 file system. 

I wouldn't go so far as to propose that. I think in all cases username/uid resolution can be pushed off to the local system. When starting an instance, the enclosing resource manager (or system in case of a system instance) will resolve user to proper local credentials. When that enclosing RM is Flux, the username to uid resolution can occur in the security shell, which already is required to validate the user based on a global secret or keypair (which probably aren't registered to a UID, but to a global username).

grondo:
This looks good as is, shall I merge it?

garlick:
OK, I guess we can merge and do another PR for the detailed sections.  Thanks for the review.

PR #75, opened 2017-02-15
rfc7: Add rule about unique typedef names

morrone:
Add new rule to require that typedef names be unique
accross a project (e.g. avoid using ctx_t in multiple places).

This change was suggested in issue #972.


grondo:
Looks like `ctx` and `fooservice` need to be added to the spellcheck dictionary. Otherwise I'd merge it

Issue #55, opened 2016-08-09
Add allocation-delay-until request in canonical jobspec?


dongahn:
Migrating this question from PR #29. I think we need something like this to support "delay allocation until the specified time: e.g., `sbatch`'s `--begin` option.

> Does it make sense to change walltime key to time key and the time key expresses both the walltime requirement and the start time requirement?

morrone:
I think that a set of wall-clock time constraints would be reasonable.  Perhaps:

* start - allocation must begin at this fixed time
* start-after - allocation can begin at any time after this fixed time

The former, "start", might be configurable to be capability-limited (for instance, only LC staff can use it for scheduling DATs...or maybe it the capability is more complex: anyone can use "start" for 50 or fewer nodes, but more than 50 nodes requires LC staff capability).

And maybe more complex time-based constraints in the future "I would like 3 hours between 9am and 5pm on a Weekday".  That might allow for some medium size interactive jobs that don't fit in the interactive pool, but also would not require administrative staff to get involved to schedule.

dongahn:
Yes, I think being able to specify such constraints in the job spec would be a great start! 

On, the "start at this fixed time": 

Since this is a guarantee the scheduler will have to provide, it seems the scheduler needs to be advanced reservation-aware? Something to think about, as we map out our scheduling primitives in a new RFC...

dongahn:
@morrone: does the current canonical jobspec have ways to specify start time requirement variants as described above? At this point, I believe I can schedule them if they are specified with our jobspc. (Even starting at a fixed time -- which folks typically call advanced reservation.)

morrone:
@dongahn Yes.  A scheduler may support any key/value pairs it likes in the attributes->system area.  "duration" is just an example.

SteVwonder:
>    start - allocation must begin at this fixed time
>    start-after - allocation can begin at any time after this fixed time


Based on the discussions in https://github.com/flux-framework/flux-core/discussions/3243, I believe we will be implementing the `start-after` using dependencies.

Exactly how we plan to handle `start` is still up in the air.  As @dongahn mentioned, it will require the scheduler to make an advanced reservation to guarantee the start time.

PR #76, opened 2017-02-23
RFC 4 revisions

lipari:
Added methods to RFC 4 to support resource reservations, staging and scoring.  Resources can now be reserved and reservations, cancelled.  The optional Staging method allows allocations and reservations to be made atomically,  The optional Scoring method supports the selection of the optimal set of resources to allocate from all the matched candidates.

Revised the notion of sharing.  Sharing now follows from the rule that units of a resource's size can be allocated to at most one job.  A new "shared" field was added to track the allocations of child resources and to facilitate the search for resources that can be allocated exclusively.

Finally, a number of minor changes were made to group methods by functional category and bring consistency to the formatting.

### DIFF
--- a/spec_4.adoc
+++ b/spec_4.adoc
@@ -156,40 +156,16 @@ scheduling jobs in the future.
 When operating on a resource as an object, the following methods
 SHALL be supported.
 
-Size:: A method to query the current size of a resource pool SHALL
- be provided.
+=== Load / Save
 
-Tag (K, [V]):: A method for tagging resource pools with
- arbitrary key/value pairs SHALL be provided. The value _V_ SHALL
- be optional.
+Serialize:: A method for serializing/deserializing a resource pool and
+ its children SHALL be provided to allow for transmission for resource
+ pool hierarchy and data over the wire, saving state to a file, etc.
 
-State:: Methods for setting and returning the state of the resource
- SHALL be provided.
+=== Copying and Destroying
 
-Aggregation:: A method for returning resource contents of composite
- object _in aggregate_ SHALL be provided. The aggregate method SHALL
- return the sum of available resources by type name. Resources with an
- available count of 0 SHALL be pruned from the results by default,
- since the composite model implies that all children of an unavailable
- resource are themselves not available.
-
-Traversal:: A method for traversal SHALL be provided to visit each node
- in the hierarchy rooted at the current object. The traversal method SHALL
- allow for optionally provided methods for determining the traversal
- pattern for each child resources. This interface SHALL allow, at least,
- the pruning of non-matching subtrees and the order of visitation of
- children during traversal.
-
-Match:: A method or set of methods for resource pool matching
- SHALL be provided by the implementation. Resource pools SHALL
- be matched on tags, properties, size, type, name, basename,
- ids, etc.
-
-Find:: A search method SHALL be provided by the implementation to
- traverse the tree and return all matching resource pools, along with
- their children, as well as ancestors up to the root of the hierarchy.
- The _Find_ method MAY be implemented as a combination of _Traversal_
- and _Match_.
+These methods below MAY be limited to the composite resource pool and
+not the associated graphs.
 
 Copy:: A method for copying a resource composite to a new instance SHALL
  be provided. This method MAY be used to create a new instance of
@@ -221,9 +197,43 @@ Unlink:: A method for removing or ``unlinking'' a resource from a hierarchy
  data table. If there are no more entries in this Resource's Hierarchy
  table, then the Resource data object MAY be garbage collected.
 
-Serialize:: A method for serializing/deserializing a resource pool and its
- children SHALL be provided to allow for transmission for resource pool
- hierarchy and data over the wire, saving state to a file, etc.
+=== Resource Pool Data Methods
+
+Size:: A method to query the size of a resource pool SHALL be
+ provided.
+
+Tag (K, [V]):: A method for tagging resource pools with
+ arbitrary key/value pairs SHALL be provided. The value _V_ SHALL
+ be optional.
+
+State:: Methods for setting and returning the state of the resource
+ SHALL be provided.
+
+Aggregation:: A method for returning resource contents of composite
+ object _in aggregate_ SHALL be provided. The _Aggregate_ method SHALL
+ return the sum of available resources by type name. Resources with an
+ available count of 0 SHALL be pruned from the results by default,
+ since the composite model implies that all children of an unavailable
+ resource are themselves not available.
+
+=== Find and Stage Methods
+
+Match:: A method or set of methods for resource pool matching SHALL be
+ provided by the implementation. Resource pools SHALL be matched on
+ tags, properties, size, type, name, basename, ids, etc.
+
+Traversal:: A method for traversal SHALL be provided to visit each
+ node in the hierarchy rooted at the current object. The traversal
+ method SHALL allow for optionally provided methods for determining
+ the traversal pattern for each child resources. This interface SHALL
+ allow, at least, the pruning of non-matching subtrees and the order
+ of visitation of children during traversal.
+
+Find:: A search method SHALL be provided by the implementation to
+ traverse the tree and return all matching resource pools, along with
+ their children, as well as ancestors up to the root of the hierarchy.
+ The _Find_ method MAY be implemented as a combination of _Traversal_
+ and _Match_.
 
 === Job Allocations and Reservations
 
--- a/spec_4.adoc
+++ b/spec_4.adoc
@@ -240,7 +240,8 @@ Aggregation:: A method for returning resource contents of composite
 
 Match:: A method or set of methods for resource pool matching SHALL be
  provided by the implementation. Resource pools SHALL be matched on
- tags, properties, size, type, name, basename, ids, etc.
+ tags, properties, size, type, name, basename, ids, etc.  The _Match_
+ method MAY consider the resource pool's allocations and reservations.
 
 Score:: A method to score the degree of a match MAY be provided.  A
  resource score would be used to identify the optimal set of resources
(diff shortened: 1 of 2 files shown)


lipari:
Rebased and ready for review.

dongahn:
Thanks @lipari. I will review this today.

grondo:
@lipari, so far I've only had a cursory read of some of these changes.

One thing that occurred to me, though, is that some of the items being added here (and perhaps even in previously merged PR), are details of an *implementation* of the Resource Model.

I wonder if would be more effective to pare this RFC down to just a clear description of the Flux Resource Model, and move implementation specific designs into a more detailed, but separate RFC? I know there is kind of a blurry line where details start and the conceptual model ends, but I didn't think it would hurt to suggest this. (I could also be very wrong)

lipari:
@grondo, as the editor of record for RFC4, I defer to your judgement.  I liked the way the original version of this RFC included the methods which acted on the resource model.  But if you think that the latest updates to those methods bring too much complexity, then I will spin it off to a separate RFC.

grondo:
Well, I didn't mean to make it an ultimatum, but I was hoping to get your opinion on the idea, as well as others in the group -- perhaps someone else could chime in?

lipari:
My opinion is that I think it is convenient to have the methods included in the RFC.  While you are correct that there is a distinct boundary between the Flux Resource Model and the operations that act on that model, it is helpful to me anyway to see the model and operations in one place.

It is possible that I introduced more detail in the specification of each method than is needed - or worse - commits us to a design that is too confining.  My goal was to provide the core set of operations as a foundation to how one interacts with the model.  Once we're clear on the model and its associated operations, the use of the model should be unambiguous.

dongahn:
@lipari: Before reviewing this PR, I re-read the original RFC, and here are some random thoughts to begin with. I don't know if you and/or @grondo want to discuss such considerations as part of this PR, though. Let me know if and how you want to push forward this.

Anticipating the work we envision to do in the area of Job Spec and Resource API refactoring, it feels like some languages in this RFC can somewhat misguide our implementation choices. 

Here are two items that caught my eyes:

(1) **Resource pool data** currently contains **Resource composite data** and **Graph data** in its fields. IMHO, this can lead to an implementation choice where the resource pool data type directly embeds graph nodes in them. But this doesn’t have to. In fact, what would be more flexible and extensible would be to decouple the resource pool data type from resource relation type. And represent as a graph type. 

Perhaps, we can remove both of these data from **Resource pool data** and instead have a separate “Resource Pool Data Relations as Graphs” section and say to the effect that:

“We SHALL describe various relations among resource pool data using graphs. The relations shall include, but not limited to, composition (e.g., ...) and flow (e.g., power, bandwidth etc) … A resource pool data can belong to one or more graphs and will be able to identify what graphs it belongs to…” The current `resrc` implementation is one way to comply with this. But such a language would not discourage one from implementing this as general graph API and hang resource data as a `void *` field within a graph vertex, either, which may be arguably easier for a resource data to belong to more than one hierarchies...

(2) **Resource pool data** currently has over a dozen fields. Just like the above concern, I understand it will be an implementer’s choice how to organize this data (e.g., as a flat struct/class or a composition of some separable data types). Regardless, I thought it would not be a bad idea for this RFC to be more proactive and explicitly describe some groupings. This can come in handy soon, as we expect there will be a common resource representation being shared by both Job Spec, schedulers, and some core management services. One organization I thought about:

==== Resource pool data:

===== Base data
* Type
* UUID (Unique ID for this resource)
* Basename
* Name
* ID (optional numeric ID to be appended to basename to get name)
* Properties (static properties associated with this instance)
* Size (Total number of resources in this pool)
* Units (optional units associated with the size value)


===== Extended data (or however appropriate we name this)
* Tags (dynamic list of tags)
* State (e.g., up, down, degraded, failing, unknown, null)
* Resource allocation data (see below)

And removing graph/composition data as discussed above. 

I think itemizing different resource pool types has RFC benefits, allowing related RFC to cross reference a bit better. E.g., Job Spect can refer to “Base data” as a common resource representation it shares with Resource Model RFC. Job scheduling can refer to the overall resource pool data… etc.

dongahn:
> perhaps someone else could chime in?

@grondo: I would turn around your question and first ask what you think should belong to implementation details and what belong to the abstract model RFC? A model means different things to different people, and the line can be blurry... and what constitutes a model often really depends on what we try to achieve with it.

Looking at the high level section, though, my thinking is that it would be good to include 
- background and goals
- data representation for resource pool (please consider my comments above though)
- Graphs as ways to establish relationship among resource data pool instances (such as resource containment)
- abstract methods that operate on either the resource pool data and relationship among them as well as probably also their semantics

I am ambivalent whether the scheduler's use of this model (allocation/reservation) should be included in here.

I am against having implementation details of either the basic abstract model or scheduling operations -- I think this part can be lifted from here and be in a separate RFC describing how schedulers can use Job Spec and Resource API (as well as its own) to implement various scheduling-relevant policies (e.g., resource selection, job management, backfill polices ...). Even then, the software design and implementation details may belong to a separate doc from abstract one..

grondo:
@dongahn, great points and I thank you for including your thoughtful comments.

>  I would turn around your question and first ask what you think should belong to implementation details and what belong to the abstract model RFC? 

Yes, good question! I think you are correct that scheduler implementation details belong in a separate RFC,  but you and @lipari have convinced me at least that the abstract methods are fine here. Thanks again.

dongahn:
@grondo: thanks! If you agree, I can take the fist step at moving some of the scheduler aspects of this RFC into a separate scheduler RFC. 

IMHO, this would be extremely helpful at this point as we are refactoring resrc into a first class API and scheduler as the first user of that API.

grondo:
Yes, sounds like a great idea and thank you for volunteering for the work!

lipari:
I'm not clear on the plan w/r/t this PR.  Withdraw it and begin anew with a new RFC that siphons off the sched references?  Or, address outstanding suggestions, merge it and then siphon off the sched discussion?

grondo:
It might be most tractable to address @dongahn's suggestions, merge your PR, then @dongahn could do his work to move any relevant implementation specific information to a new RFC in a single PR. I think this is simplest, but it is just my opinion.

dongahn:
@lipari and I just discussed this. Essentially, we will do what @grondo suggested above. 

On a separate topic, @lipari asked a very good question regarding how to raise the abstraction of any resource relationships (e.g., composite, flow, network connection) to a graph. I think this warrant some brain storming at some point.

morrone:
There is a lot that I don't really understand in RFC4.  For instance, "Allocate (N, S)  Allocate N resources from the pool under the name S".  It isn't terribly clear, but the implication from context seem to be that "N" is a number.  That doesn't seem to fit the flux model to me.  Shouldn't we be sending a resource description rather than a number?

I also don't really understand why the model should have four different identifiers for the same resource, UUID, basename, name, and ID.  Are we mixing implementation and model here?  I readily admit that all four might be needed, and I just don't have the full understanding to grok it yet.  But if they are needed, then we really need to explain their purpose better.

morrone:
Another ambiguity for me is "Allocations" vs. "Reservations".  Why do we have both?  How are they different?  How do they interact?

lipari:
> Shouldn't we be sending a resource description rather than a number?

There was a subtle shift from the original concept of allocating some quantity of resources to allocating *these specific* resources which I chose not to revise.  I'm inclined to just drop the "(N, S)" arguments and define it loosely as "earmark a set of resources as in use by a Flux instance/program".

Regarding the four different ways to identify a resource, we could indeed boil it down to just a single "name".  However, to the extent that that name would typically involve a unique identifier, a base name, an ordinal, and some composite of base name+ordinal, I think it is valid to include the allowable or recommended pattern for this construction in the RFC.

Allocations and reservations borrow heavily from the scheduling domain, and should probably be defined better.  Based on earlier comments, this will all probably be purged from this RFC and relegated to a Scheduling-specific resource RDL.

dongahn:
> Another ambiguity for me is "Allocations" vs. "Reservations". Why do we have both? How are they different? How do they interact?

@morrone: as discussed above, I will try to move these scheduling related data and abstraction into a separate RFC. But just to answer your question, reservations are needed so that a backfill scheduler can decide whether the target resource pool data can be used to backfill a job without having to delay higher priority jobs. Also, for advanced reservation (AR) scheduling, this is needed to actually reserve a job. 

But I think this really belong to the scheduler RFC that uses abstractions provided by both this resource model RFC and job spec RFC.

dongahn:
FYI -- there would be different ways to implement backfill algorithms. But http://docs.adaptivecomputing.com/maui/8.2backfill.php was a good reference for me.

dongahn:
(While we are on a side topic :-) This [paper]( https://pdfs.semanticscholar.org/6dac/a04a6e08b3ac9a3ba2036e16f0763e99e010.pdf) also gave me some inspiration when I was struggling with the planner algorithms.

morrone:
> Also, for advanced reservation (AR) scheduling, this is needed to actually reserve a job.

I don't understand that.  Why don't we just add time as another constraint on allocations?  We already have duration constraints.  Can we not make specific wallclock time spans another kind of time contraint for allocations?  Then we don't need to special-case LLNL DAT-style reservations.

dongahn:
> Why don't we just add time as another constraint on allocations?

Yeah, I think this can be done this way as well. In fact, I vaguely remember we discussed "begin time constraint" as a job spec. Let me look.

lipari:
> I don't understand that. Why don't we just add time as another constraint on allocations? We already have duration constraints.

Reservations are needed for now-based scheduling.  You have a job at the top of the queue that cannot run because not enough resources are available.  You "reserve" the resources that satisfy the job's request and are available.  Then you look to the next job down in the queue and see whether there are resource for it.  You avoid the resources "reserved" for the top priority job.  Perhaps you reserve some different resources for the 2nd job (that the top priority job couldn't use).  The you visit job #3.

Once you make it to the bottom of the queue (or as far down as you want to go), you release all reservations and start again at the top priority job.

dongahn:
There is this issue: https://github.com/flux-framework/rfc/issues/55

But now that I think about this, I am unsure if the "begin time" constraint can give a guarantee, though.

morrone:
> Reservations are needed for now-based scheduling. You have a job at the top of the queue that cannot run because not enough resources are available. You "reserve" the resources that satisfy the job's request and are available. Then you look to the next job down in the queue and see whether there are resource for it.

But isn't that detail entirely internal to the scheduler implementation?  Why are we exposing that to the outside world?

dongahn:
> But isn't that detail entirely internal to the scheduler implementation? Why are we exposing that to the outside world?

@morrone: good point! I guess this was a part of the reason why we agreed to lift some of the scheduling-related sections to a separate scheduling RFC. Allocation, reservation etc becomes no longer internal to the scheduler implementation there, as scheduler plug-in writers will look at that RFC and write scheduling algorithms atop of the abstraction?

lipari:
Just created [PR 78](https://github.com/flux-framework/rfc/pull/78) to address the comments above.  Withdrawing this PR.

PR #78, opened 2017-03-02
rfc4:  Pare the RFC down to a simpler model

lipari:
This commit implements much of the discussion in (https://github.com/flux-framework/rfc/pull/76).

This involves two main changes:

* Eliminate aspects of the model that relate to scheduling operations

* Provide a distinct abstraction between a resource and any graphs to
  which it is associated.  This includes the composite hierarchy.
  Description of the composite hierarchy is retained, but becomes just
  another graph.

Additional RFC's are planned to specify:
* Flux Resource Graph Model
* Flux Scheduling of Flux Resources Model

### DIFF
--- a/spec_4.adoc
+++ b/spec_4.adoc
@@ -90,7 +90,7 @@ MAY be configured to operate at the racks and aggregates on their
 containing nodes while a lower-level instance MAY actually operate at
 the nodes and cores as the finest resource granularity.
 
-==== Resource pool data:
+==== Resource Pool Data
 
 * Type
 * UUID (Unique ID for this resource)
@@ -98,13 +98,8 @@ the nodes and cores as the finest resource granularity.
 * Name
 * ID (optional numeric ID to be appended to basename to get name)
 * Properties (static properties associated with this instance)
-* Tags (dynamic list of tags)
 * Size (Total number of resources in this pool)
-* Units (optional units associated with the size value)
-* State (e.g., up, down, degraded, failing, unknown, null)
-* Resource composite data (see below)
-* Graph data (see below)
-* Resource allocation data (see below)
+* Units (optional units associated with the `size` value)
 
 The default value for `basename` is the `type`.  The default value for
 `name` is a concatenation of `basename` and `ID`, or just `basename`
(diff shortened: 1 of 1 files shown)


dongahn:
@lipari: looks like a great start! 

What I was thinking about was going a bit deeper though. Like:

#### Basics:

- Resource Pool: as one concept — a means to describe any resource item

- Graph: another basic concept — a means to describe any relationship among resource pool instances


#### Data models:

- Resource Pool Data: separate out the base data and extended data

- Graph Data: Vertex and Annotated Edges


#### Example Composition of Resource Pool and Graph Elements

- Composite Resource Pool: Composition of hardware resource pool and graph representing a resource composite hierarchy -- most of the wording goes into there

- Power Resource Pool as a composition between power resource pool and graph representing power hierarchy

- Maybe how above two can even be connected under the same abstract “resource pool” and “graph” paradigm (i.e., our general Flux resource model)

#### Then, methods, we may need to separate out 

- Methods exposed from Resource Pool Concept

- Methods exposed Graph method


Does this generalize things too much? Thoughts?

lipari:
@dongahn, rather than channeling your ideas through me, why don't you just merge this PR and then create a PR with your ideas?

dongahn:
Sure. If it is okay with @grondo, i will merge it and then submit my mod in a separate PR.

grondo:
@dongahn, yes, you don't need my permission! It seems @lipari and you are making great progress IMO!

dongahn:
> Example Composition of Resource Pool and Graph Elements

> Composite Resource Pool: Composition of hardware resource pool and graph representing a resource composite hierarchy -- most of the wording goes into there

> Power Resource Pool as a composition between power resource pool and graph representing power hierarchy

> Maybe how above two can even be connected under the same abstract “resource pool” and “graph” paradigm (i.e., our general Flux resource model)

A stupid question. I am not super facility with doc, but is it possible to directly embed images into ascidoc? 

I thought one way to explain our system is through some simple examples -- how simple concepts can build up. 

So, I came up with three png files illustrating how the general graph concept can describe various relationships beyond composite relationships. 

in

Explain that "HAS" and "IN" edges describe composites

channel

Explain that one can model parallel file IO bandwidth with CHANNEL_BW_OF relationships

combined

Finally, explain that these two relationships can be easily combined under the same graph concept. 

I think, having such illustrations will make our life easier in explaining what scheduling can do to use and act on this Flux Resource Model. There, I can even explain some advanced resource selection strategies. Like, File IO BW-aware scheduler can have a few strategies

1) Use the root vertex of IO BW hierarchy as the starting traversal point and get to the compute node vertex to allocate BW and node at the same time: Stephen's algorithm.
2) Use the root vertex of the composite hierarchy as the point and get to the node vertex -- but then on visiting each compute node vertex, do a "up walk" toward the IO BW hierarchy to do further constraint check
3) Employ hierarchical scheduling -- have the upper level scheduler perform IO BW-aware scheduling for a group of jobs as in 1) and then a child instance does topology-aware scheduling on the allocated resource set (IO-aware primary, topology-aware secondary).

If there isn't a good way to embed those figures, I will embed a URL to a file (hopefully I don't have to do this.)

morrone:
@dongahn , yes you can: http://asciidoctor.org/docs/asciidoc-writers-guide/#images

What is "Cab" in your example images?  I initially thought it was short for "Cabinet", because the next box was "Rack"...but that doesn't really make sense.  I am guessing "Cab" is the name of a cluster and "Cab11" is the name of a node?  It might make sense to replace those with "Cluster" and "Node11".  "cab" wouldn't necessarily be interpreted as a cluster/node name anywhere but at LLNL where we have a cab cluster.

PR #79, opened 2017-03-02
Add RFC15 for independent privileged service

grondo:
This PR includes a new RFC describing the single privileged Flux service enabling multi-user execution -- The Flux Security Independent Minister of Privilege, or Flux Security IMP.

I've already been through a couple revisions of the design with @dun, and now we're opening it up for further discussion among Flux developers.

The overall design is of a separately installed, setuid executable, `flux-security-imp`, which is used by multi-user flux instances to execute processes as other users on behalf of the instance, execute job prolog/epilogs if required, and signal and terminate multi-user jobs.

The design will require public/private keypairs for all users, but the RFC doesn't (yet) go into detail on how that is managed. At this point we assume that users will have access to private key anytime a job is submitted (to sign a request credential), and that the security imp has the ability to obtain and verify public keys scalably when launching a job (possibly using public certificates signed by a trusted CA). We still need to embark on this part of the design,  but for now the key distribution can be thought of as out-of-scope for this PR.

Finally, one thing that should probably be added somehow to the PR is that this whole design does not protect user data or offer any privacy from the *instance owner* when a guest user runs a job in an instance they do not own. There is a presumed level of trust that a guest user has for an instance owner when they make use of a multi-user instance running as that owner.


garlick:
Nice!  Very nice!

Couple minor comments so far:

User roles: any reason not to use "instance owner" here for consistency with other rfcs?  Possibly "guest" could also be used in RFC 12 for consistency (I'll remember that when I propose an update for rolemask, etc). (minor!) formatting of this section could be better.

> For instance, a site may configure permissions such that only a flux user has execute permissions, thus allowing a mult-user system instance running as flux, but disallowing sub-instance jobs access to multi-user capabilities.

Very cool!  I suppose sub-instances running as Flux will work though so sub-instances per se are not restricted.

The "User Request" has an nice _U_ representation.  Should "Job Request Credential" be _R_?  Under Credential Verification, _U_ is referred to as "User Credential".  Maybe simply _U_ will suffice after it is defined to avoid the possibility of inconsistency?

garlick:
Should the root user be explicitly forbidden to run things via the IMP?  (I don't think there's any reason root wouldn't work here, but would forbidding root reduce the seriousness of an exploit if there were one?

grondo:
Thanks! Good suggestions!

This is still a very early draft so any and all suggestions are nice.

> Very cool! I suppose sub-instances running as Flux will work though so sub-instances per se are not restricted.

Yeah the key here was "disallowing sub-instance job access to multi-user capabilities"

Maybe there is a better way to word that?

> User roles: any reason not to use "instance owner" here for consistency with other rfcs? Possibly "guest" could also be used in RFC 12 for consistency (I'll remember that when I propose an update for rolemask, etc). (minor!) formatting of this section could be better.

To satisfy requirement that the IMP can run standalone, the idea of "instance owner" was generalized to "resource owner", however the follow through on that concept was probably not good, because eventually I reworked the RFC to be more "instance" oriented (otherwise it hurt my brain)

garlick:
> Yeah the key here was "disallowing sub-instance job access to multi-user capabilities"

But a sub-instance running as flux could still run the IMP.  Only sub-instances running as other users would be restricted. Right?

grondo:
> But a sub-instance running as flux could still run the IMP. Only sub-instances running as other users would be restricted. Right? 

Yes, I see what you are saying, sorry I misread your original statement  -- some wordsmithing needed there -- Perhaps something like 

 > For instance, a site may configure permissions such that only a flux user has execute permissions, thus allowing a multi-user system instance running as `flux`, but disallowing sub-instance jobs running as normal users access to multi-user capabilities.

grondo:
> Should the root user be explicitly forbidden to run things via the IMP? (I don't think there's any reason root wouldn't work here, but would forbidding root reduce the seriousness of an exploit if there were one?

Good point! Though to run the IMP as root, you'd already need root, but I don't see a good reason for enabling it, so it probably should not be allowed.

garlick:
> Though to run the IMP as root, you'd already need root.

I was thinking of whether the IMP would execute things as root if it was instructed to do so with a root-signed credential.  The damage that could be done with a forged credential might be limited somewhat if root was explicitly not allowed to run, ever.

grondo:
> I was thinking of whether the IMP would execute things as root if it was instructed to do so with a root-signed credential. The damage that could be done with a forged credential might be limited somewhat if root was explicitly not allowed to run, ever.

Yes, good point sorry. (my reading comprehension skills appear to be sub-par today), though a site would have to first create root keypair and a root public certificate signed by a trusted CA -- so in some ways every user enabled in Flux is "opt-in" only. (e.g. not every user in /etc/passwd can run)

Also, this does prevent the ability to leverage system instance of flux from running root commands (e.g. as scalable pdsh alternative), cron jobs, etc. However, I'm sure there's other, safer ways to implement this, and taking the safer approach for now seems reasonable.

morrone:
Are we sure that we want to explicitly require "access to the public keys"?  Would that allow a Kerberos plugin in a reasonable way?  Perhaps the specific authentication technology (public key, or something else) should be left up to the plugin.

grondo:
Good point thanks! We definitely don't want to preclude the use of kerberos or other SSO and Identity Management solutions, so a pass through with that in mind is needed.

garlick:
How does the IMP know who is a valid signer of resource allocations, when resources could have been passed through several layers of sub-instances, handed off between multiple users?  Put another way, if _I_ sign a resource credential, how does the IMP detect that I'm an imposter trying to grab resources that aren't mine?

grondo:
> How does the IMP know who is a valid signer of resource allocations, when resources could have been passed through several layers of sub-instances, handed off between multiple users? Put another way, if I  sign a resource credential, how does the IMP detect that I'm an imposter trying to grab resources that aren't mine?

Hm, yeah that really should be explicitly described in this RFC, shouldn't it? I think the logic can go something like this

 * A process is always started in a process tracking group (e.g. container). 
 * A process may only start subprocesses in its own process tracking group
 * There is a way to determine the "owner" of the current process tracking group
 * Process tracking groups are hierarchical

Therefore, the IMP can always determine under which process tracking group it has started, and the user that owns that group. Assuming at least the local resource set is kept as part of that group (again, e.g. containers), the IMP can know if its invocation has access to the assigned resource set.

Unfortunately, I now realize, in practice this could mean keeping some system-wide state for the IMP to track resources for which there is not a system "container" (e.g. licenses?) Perhaps this state could be kept as part of the containment plugin stack, and would "simulate" a container/cgroup?

Now that I'm thinking about this, I wonder if a chain of signed resource sets could be used here, if we have to keep state anyway. Part of the IMP configuration could be the system owner (e.g. `flux` user), and a stack of signed resource set allocations cached locally could be verified up to original grant from the system owner on each new IMP invocation. (Really this is similar to the stack of containers, but more general)

Does any of that make sense?

garlick:
Yeah, definitely.  That seems very clear and should be added to the RFC.

How did you want to handle the evolution of this RFC?  Since this PR is the result of quite a lot of up front work, did you want to get it merged now and open new PR's for mods/additions?

On plugins for security, does the design fundamentally require public/private key pairs for every user and a CA of sorts, or could kerberos or munge provide the basis for the required functionality?  Just wondering what pieces need to be implemented by us in order to achieve the design and still be usable in environments that have only munge or only kerberos.

grondo:
I'd like to get buy-in from the group that at least the concept is sound
here before merging. For the edits to the PR I will push each change as a
new commit with a commit message that reflects the source of the change
request. Once we're happy with the draft, I can squash and get it merged.

@morrone asked the same thing about fundamental requirement for key pairs.
I was hoping that the design would not preclude the use of kerberos, but
I'm not sure how this scheme would be tied to a system like that.
Fundamentally, it does require that all users have some method of signing
data, and that the IMP has a method of verifying those signatures. I think
we should work through this with @dun to see how to generalize it in
practice.

I was hopeful that somehow Munge could be leveraged for initial design, or
single-munge-domain instances. It seems at least possible from the
description above.

The other piece of the puzzle here is the scalability of key distribution.
I was hoping that the existing instance could be leveraged for distribution
of public keys (e.g. when you register a user with your instance, the
process could include adding a public certificate to the userdb). However,
we *also* need a way to distribute the private key, or the ability to
non-interactively sign data, because otherwise a user would not be able to
submit jobs as part of their new job...

garlick:
What if the IMP had a pre-shared key like munge and used SO_PEERCRED to authenticate users and sign requests/validate jobspec with its key?

For example, I submit my job request by sending jobspec to the IMP over a PF_LOCAL socket.  The IMP determines my identity, adds my identity to the jobspec, signs the whole lot with its key, and sends me back  _U_.  I then submit _U_ to a flux instance.

The flux instance passes _U_ with resource allocation to the IMP over the PF_LOCAL socket.  The IMP determines the instance owner identity, checks it against the resource owner identity, and verifies the signature of _U_ with its own key.

Providing my own critique:  I guess this proposal is the moral equivalent of munge with its replay attack safeguards disabled, which begs the question of why that would be acceptable and how it could be addressed.  Specifically I mean the short TTL and refusal to decode a credential more than once - these features would be at odds with the need to queue _U_ for a long time before execution and the need to validate  _U_'s authenticity at different points in its life cycle, for example at submission by the instance owner and at execution by the IMP.

garlick:
The above comment was maybe off topic, diving down into _one_ possible way to satisfy the architecture with a munge-like scheme.  Sorry for the stream of consciousness post.  I thought I understood something and got excited :-)

Trying to get back on point:  maybe the public key language in the RFC ought to be replaced with a more general requirements for verifiable auth and integrity on _U_ and the resource cred, provided by plugin, since it seems there probably are ways to achieve that without presuming every user has a public/private keypair?

If we go that route, I would say that the specific requirements for credentail reuse (re-verifiability anyway) and long TTL (or at least TTL correlated with max TTL of life in the job queue) should be stated also.

morrone:
Perhaps we could consider GSSAPI.  It might not fit, but that API might provide some inspiration for ours.

morrone:
Can you please elaborate on what this line means?:

> 2. Verify owner has access to assigned resource set

How does the IMP do that?  Is this just implying that it checks for a signature of a trusted system resource owner, or does the IMP need to go make queries with some external entity?

morrone:
Do we perhaps also need to add support for explicit revocations in the IMP?  What if a job is cancelled 10 hours early, but valid signed certs are still out there for the resources?  Perhaps the IMP needs to log revocations until the cert window has expired so that out-of-date certs can't be used.

grondo:
> Can you please elaborate on what this line means?:
> > 2.  Verify owner has access to assigned resource set

Yeah, @garlick also called out the lack of description here, see the currently accepted response in [this comment](https://github.com/flux-framework/rfc/pull/79#issuecomment-284001594)

When containers are active and working as designed, a *resource owner* will only be able to start the IMP on resources to which they have access. Therefore, the IMP can just check the contents of the current container to determine if the owner has access to the assigned resource set.

I do realize that some resources can't be contained, so there may be some ancillary data kept along with the container plugin system to track these resources. Alternately, the IMP could keep a database or cache of the tree of current job requests, and could verify ownership up to some system user, e.g. `flux`.

grondo:
> Do we perhaps also need to add support for explicit revocations in the IMP? What if a job is cancelled 10 hours early, but valid signed certs are still out there for the resources? Perhaps the IMP needs to log revocations until the cert window has expired so that out-of-date certs can't be used.

You are correct, this might be necessary. However, I wonder if it is a critical issue if the only user able to use the expired request is the instance owner, the same user that presumably handled or initiated the cancel request? (I admit that I could be missing some key corner cases here, because we haven't yet fully thought through all the possible ways to break this current design)

grondo:
> maybe the public key language in the RFC ought to be replaced with a more general requirements for verifiable auth and integrity on U and the resource cred, provided by plugin, since it seems there probably are ways to achieve that without presuming every user has a public/private keypair?

This is a great idea, I'll try to remove references to specific keypair language.

garlick:
Hmm, the GSSAPI thought is interesting.  Apparently [SPKM](https://tools.ietf.org/html/rfc2025) is a public key GSSAPI mechanism.  I didn't realize that there was one.

morrone:
> You are correct, this might be necessary. However, I wonder if it is a critical issue if the only user able to use the expired request is the instance owner, the same user that presumably handled or initiated the cancel request?

What stops that owner from using the credential after he has killed the allocation, and someone else is now using the resources?  I suppose rule "2", but I'm still fuzzy on how that would work.

morrone:
I guess what seems a bit odd to me is that IMP needs to go to some other entity to verify that the resources are really accessible to someone when the certificate lists exactly what resources are granted to someone.  Perhaps what could happen for sub launches that switch users is that an additional certificate can be issued.  So we could have this:

User A allocates resources to user B.  A resource allocation signed by 'A' and a task allocation signed by 'B' is sent to IMP.  'A' is also a known system-level user so we trust it to allocate resources at the top level.

Now perhaps the tasks run by user B is a new flux instance.  If it only runs tasks be user B, then it only needs to provide the same sort of data to IMP each time: the resource allocation by 'A', and the task allocation signed by 'B'.  But if B's allocation shares resource with other users, perhaps it needs to provide more information:

- Resource allocation signed by 'A' (says that it is granting access to user 'B')
- (Sub)resource allocation signed by 'B' (says that it is granting acces to user 'C')
- Task allocation signed by 'C'

Then there is a verifiable chain of signatures back up to a trusted user.  It imposes a higher level of authentication overhead proportional to the number of changes of ownership, but in reality the number of changes will probably be fairly small I would guess.

grondo:
> What stops that owner from using the credential after he has killed the allocation, and someone else is now using the resources? I suppose rule "2", but I'm still fuzzy on how that would work.

It is necessary to have a mechanism that prevents a user to start a process on resources that aren't allocated to them. This is a general Flux requirement. The owner can't use the credential on resources they don't own because of this requirement.

If you're asking if an owner can kill the allocation of guest 1, assign new allocation to guest 2, and then use the original guest 1 credential to run something on guest 2's allocation. They can. This would be considered a bug I guess, or at least unsociable behavior. But as resource owner, they can do as they please, and by submitting the job request to the instance owner, these guests have agreed to some level of trust relationship with the resource owner.

I guess the crux of this design is that by submitting a request to a Flux instance you don't own, you are trusting the owner of that instance to some level of honest behavior. This is pretty straightforward for a system instance running as some trusted user (e.g. `flux`), but is perhaps more ambiguous when the owner is another normal user.

Perhaps there at least needs to be a list of trusted users kept in a flux config dir somewhere, and connecting to a non-trusted instance would get you a "are you sure?" prompt?

garlick:
> Perhaps what could happen for sub launches that switch users is that an additional certificate can be issued. So we could have this:

Search for "chain" above - I think you and @grondo had the same idea :-)

grondo:
> I guess what seems a bit odd to me is that IMP needs to go to some other entity to verify that the resources are really accessible to someone

Hm, I didn't mean to say that some other entity was involved here. In reality, the IMP will need to create a new container on the local system before it invokes the user processes. This will fail if the IMP is started on a container that doesn't contain the resources in the request. By "some other entity" did you mean cgroups?


> Then there is a verifiable chain of signatures back up to a trusted user. It imposes a higher level of authentication overhead proportional to the number of changes of ownership, but in reality the number of changes will probably be fairly small I would guess.

This is a good idea, and is what I was hinting at with 

> Alternately, the IMP could keep a database or cache of the tree of current job requests, and could verify ownership up to some system user, e.g. flux.

The idea of requiring the instance owner to supply the chain directly is a good one though! I can't believe I didn't think of it that way. I was stuck on requiring some kind of per-system database, but it would be great if we didn't need that. Thanks!

grondo:
OK, the current revision has updates that

 * Attempt to remove references to explicit "keypairs" in favor of simply "signature" or "data integrity and authenticity"
 * More consistent use of *R* and *U* for job and user requests
 * Remove use of "credential" for *R* and *U* as I'm not sure they are technically any kind of credential
 * note instance (owner) provided plugins
 * add Resource Ownership Verification and Revocation sections

garlick:
This is looking good!  Here's one thing I noticed on first read

> The system owner is the only trusted user and requests from this user SHALL NOT require verification.

Did you mean that it isn't required to present a resource grant from a higher authority?  Wouldn't a request from the system owner still needs to be signed or verified in some way?

grondo:
> Did you mean that it isn't required to present a resource grant from a higher authority? Wouldn't a request from the system owner still needs to be signed or verified in some way?

Sorry, this should be "resource ownership" of requests from this user SHALL NOT require verification

grondo:
One open question that still remains is how the IMP safely turns the *jobspec* in *U* into a set of processes for the user. I had in the back of my mind that the IMP would (after containment, system plugins, prolog, etc), switch UID and exec the "job shell" (somehow passing required jobspec). However as @garlick astutely pointed out, the instance owner cannot supply the job shell, so the path to the job shell must either
 
 A. be part of the jobspec request
 B. be part of IMP system configuration

Or perhaps it could be a bit of both? e.g.

 * Part of the IMP's configuration is a *list* of allowed job shell paths (e.g. like `/etc/shells`)
 * If no shell is provided in user's jobspec, the IMP invokes a default job shell
 * If a shell is provided in user's request, then it must be in allowed job shell list

This ensures that the instance owner cannot substitute a job shell upon the user which may not run what the user intended.

However, the drawback is that the set of supported job shells may have to work with multiple Flux versions, to support the case where an instance owner is running test code in a multi-user mode. This could be mitigated by carefully designing the instance-to-job-shell interface, such that instance or version specific functionality could be abstracted into the instance itself (e.g. how I/O and task state are propagated) and a simplified interface could be presented to the job shell.

Instance specific functionality is already abstracted in the IMP via the "owner provided plugins", and perhaps something similar could be done for the job shell, though the code would have to be sandboxed since it would be provided by the instance owner, but run as the guest user....

garlick:
Just pondering the role of the job shell and how it relates to the other entities.

For reference the job shell was described flux-framework/flux-core#334 (possibly needs an update)

Just to get clear on how the shell interacts with the instance - is this correct
* instance spawns IMP, giving it _R_
* IMP authenticates _R_, unwraps _J_ and resource grant
* IMP execs shell (as user, in container), giving it _J_ (shell's parent is instance)
* shell authenticates _J_, unwraps job request
* shell connects back to instance as guest user
* shell spawns processes as specified by _J_
* shell handles PMI, ptrace, signals, stdio; reports phase changes, resource usage
* eventually processes exit, shell reports exit codes, shell exits
* even if shell exits uncleanly (e.g. container resources exhausted, user kills shell, etc), the instance can detect this with `wait()` due to parent relationship
* the instance can also restart the IMP and tell it to destroy the container

So the user has to trust the shell to dutifully authenticate _J_ and carry out the job request, and the instance has to interact with the shell, though the shell would be accessing the instance as a guest, and the instance services _should_ be resilient to the antics of a misbehaving guest.

Possibly the instance could provide in _R_ an "interface version" which would help the IMP select the right shell from a list in its /etc/shells style config?  From a packaging perspective, one could have flux-compat style packages that side-install older versions of the shell to support older user-compiled versions of flux?

I think it would be OK to allow the user to "force" in their own shell to support testing of new versions of Flux with multi-user.  This could be the default for Flux in-tree testing of multi-user where the interface may be changing, for example.  It seems like we do have to be careful about letting users routinely trust any old shell though, so this might need to be a configurable policy?

spec_15.adoc
```diff
@@ -0,0 +1,352 @@
...
+
+==== Flux job prolog and epilog
+
+A multi-user instance of Flux may also require the ability to run job
+*prolog* and *epilog* for jobs. These scripts are typically site-supplied
+administrative scripts which are expected to run as a privileged user,
+and therefore will require the IMP to execute. In support of prolog
+and epilog scripts, the IMP has the following requirements:
+
+ * The IMP SHALL implement `prolog` and `epilog` subcommands which allow
```
morrone: It is not clear to me how the IMP would implement an epilog command, given that earlier it says "After the call to exec(2) the security IMP is replaced by the guest user processes."  If the IMP is gone, how does it run an epilog later?

I wonder if the IMP is really the right layer to implement epilog and prolog?  Perhaps that should be a higher-level concept, and the eplilog/prolog just look like any other signed execution request to the IMP.

garlick:
Sorry, I think the discussion is leading the document a bit. and I may have dragged us a bit off topic.  My guess is @grondo is trying to write up the larger security issues and get feedback before delving down into the details, but it's hard not to think about the details when discussing these issues.

grondo:
Ok, sorry guys, obviously the document is still not clear.

> I feel like we are reading different documents. :) I read it as saying the the IMP at one point forks a version of itself that drops privileges and then interacts with the still-privileged IMP part, but I don't see anything that describes a second program. Can you point me to what I am missing?

@morrone, the privilege separation should say that the IMP drops privileges to the *instance owner*. This is so that the impact of parsing, libflux, options process, and processing environment variables happens as the *calling user* and not in a process with superuser privileges. This is the same reason sshd uses privilege separation. After the parsing is complete the unprivileged process can exit. When the IMP is satisfied that the request is valid and verified, it can exec the "user work" which (so far) is assumed to be a job shell (rexecd equivalent)

As far as I'm concerned, whether the IMP also provides a job shell is still an open question, though there are many benefits to keeping that a separate utility, mainly for the reasons @garlick described.

> It is not clear to me how the IMP would implement an epilog command, given that earlier it says "After the call to exec(2) the security IMP is replaced by the guest user processes." If the IMP is gone, how does it run an epilog later?

I thought it was spelled out in the document that the IMP provides subcommands "kill" "prolog" "epilog". The IMP runs an epilog later because  the instance separately calls "flux imp epilog". The instance may also want to run a job prolog without actually running work for a user, and therefore a subcommand "flux imp prolog" is required, and so on.

> I wonder if the IMP is really the right layer to implement epilog and prolog? Perhaps that should be a higher-level concept, and the eplilog/prolog just look like any other signed execution request to the IMP.

I'm not sure I agree as that would require allowing running general commands as root through the IMP.
However, I'm sure I could be convinced if you have some reason for this argument...

grondo:
> I wasn't under the impression that the IMP/shell would handle PMI. Handling PMI implies that the IMP/shell is a parallel application requiring network bootstrapping

The job shell *is* a parallel application. At least if you consider that it has to orchestrate when it considered the program it is launching "started" and "completed", handle potential broadcast of stdin/signals, etc. I would argue it doesn't need to handle its own bootstrapping because it is using an existing Flux instance (e.g. it is the *instance* that provides PMI for the job, however the job-shell must supply some sort of context for programs to actually use the Flux PMI implementation)

grondo:
> My guess is @grondo is trying to write up the larger security issues and get feedback before delving down into the details, but it's hard not to think about the details when discussing these issues.

Yes, perhaps it was premature to attempt to write up these concepts as an RFC

garlick:
> Yes, perhaps it was premature to attempt to write up these concepts as an RFC

No I think we need to do this to move forward, and you've captured a good overall design.

morrone:
> The job shell is a parallel application. At least if you consider that it has to orchestrate when it considered the program it is launching "started" and "completed", handle potential broadcast of stdin/signals, etc.

I guess I don't understand why the piece of software that only needs to execute processes on a local node needs to be network aware.  Can't we leave that to the flux brokers, which are already networked?  Shouldn't the instance be the place that decides when a parallel job has finished, because it has the global knowledge of when all of the per-node applications have finished?

> I thought it was spelled out in the document that the IMP provides subcommands "kill" "prolog" "epilog". The IMP runs an epilog later because the instance separately calls "flux imp epilog".

I see that, but I don't understand it.  Why does the IMP need to know what an "eplilog" is?  Why can't the instance sign an properly formed IMP execution request that says "run this".  Why would we introduce extra complexity of making the IMP more aware of what it is running any why?  What are we gaining by that extra complexity?

garlick:
> I guess I don't understand why the piece of software that only needs to execute processes on a local node needs to be network aware.  Can't we leave that to the flux brokers, which are already networked?

The job shell is really just a piece of Flux that implements a distributed shepherding service.  It would leverage instance services, collectives, etc. not re-implement those.  No need to use the network directly for example.

garlick:
If a drawing like this helps at all I can try to make one for real (oops: _J_ should be _U_):

img_20170316_074850

springme:
Let me know if you want an artist to render this for you.

grondo:
@garlick, thanks that drawing is pretty good! I have some lying around but none that do a good job simplifying the overall picture like that. If it makes sense to choose better symbols for any of the tokens and blobs in the document, I'm all for it!

However, my worry is that we don't yet have consensus on what entity in the instance actually invokes P1...Pn, so this picture is just one possibly so far (though it accurately depicts the version I was thinking of)

> Let me know if you want an artist to render this for you.

It would be nice if we could generate in pgf/Tikz for easier editing when small changes and tweaks are being made along the way.

grondo:
> Why does the IMP need to know what an "eplilog" is? Why can't the instance sign an properly formed IMP execution request that says "run this". 

The instance could definitely do this, but the epilog would then run as the instance owner. Perhaps it isn't clear that the prolog/epilog, when run via the IMP, run as root? Or were you implying that there could be a sudo-like capability where instance owners are allowed to run some commands as root? To me, this seems like it adds more complexity, not less.

I apologize if I'm just misunderstanding your point.

garlick:
> If it makes sense to choose better symbols for any of the tokens and blobs in the document, I'm all for it!

It feels nitpicky but _J_ for some reason seems more natural to me than _U_, since I think of it as signed jobspec.

grondo:
> It feels nitpicky but J for some reason seems more natural to me than U, since I think of it as signed jobspec. 

No, I agree and thanks.

grondo:
@garlick are you working on diagram above, or should I take a stab at it while I rework some other parts of the rfc?

garlick:
Feel free if you are so inclined.

morrone:
> If a drawing like this helps at all I can try to make one for real

It will definitely be nice to have that in the finished document.  In the mean time it'll give me something to point at when  chat about it in person.

grondo:
I've pushed some slight improvements here, though more work still needs to be done (I need an editing pass, there are known typos)
 
 * Move `spec_4_data` to `data/spec_4` so we can also have `data/spec_15`
 * Inclusion of @garlick's diagram, improved description of overall design (including use of *job shell*)
 * Rename `U` to `J` as suggested by @garlick 
 * Remove mention of `epilog` and `prolog` subcommands -- how any prolog/epilog is run is outside scope of the document
 * Remove mention of `R'`, but keep specification of privilege separation so parsing is safe.

spec_15.adoc
```diff
@@ -0,0 +1,358 @@
...
+of the _instance owner_, including all processes running on computational
+resources of the instance. This design works well for single-user instances
+of Flux, but multi-user capable instances require some mechanism to perform
+privileged operations, most notably when executing work on behalf of a
+non instance owner (a guest).
+
+In the Flux system, this privilege - along with all related operations - is
+contained within a single service, the _Independent Minister of Privilege_
+(IMP), which is responsible for allowing instance owners to run work on
+behalf of a guest when the guest user has authorized the instance to do so.
```
lipari: This runs counter to what I expect.  I expect:
the instance owner has authorized the guest to do so.
or
the guest user has requested use of the instance from the instance owner.

spec_15.adoc
```diff
@@ -0,0 +1,358 @@
...
+ * More fine grained administrative control of privilege. For example,
+   simple filesystem access controls may be used to limit which
+   users are allowed to run multi-user without preventing these users
+   from launching Flux instances altogether.
+ * Arbitrary users can run multi-user instances of Flux, thus allowing
+   users to share their jobs 
+
+=== User Roles
+
+For the purposes of this RFC there are three main user roles:
```
lipari: I count four roles.

spec_15.adoc
```diff
@@ -0,0 +1,358 @@
...
+ * environment
+ * processing of marshaled job request *_R_* on stdin
+
+Privilege separation is used to minimize the code running with root
+privileges, and to avoid parsing and processing of possibly untrusted
+data in an address space with elevated privileges. For instance, an
+attacker could attempt to execute the IMP with poorly formatted
+data in an attempt to exploit a weakness in a JSON library. When
+run with privilege separation however, JSON parsing would occur
+with credentials of the IMP inoker, thus eliminating any impact
```
lipari: replace inoker with owner?

spec_15.adoc
```diff
@@ -0,0 +1,358 @@
...
+
+== IMP Internal Operation
+
+=== Privilege Separation
+
+When the `flux-security-imp` is invoked _and_ has setuid permissions, it
+SHALL first perform privilege separation. An underprivileged child is
+invoked _as the instance owner_, and this temporary child handles
+processing any input, including but not limited to
+
```
lipari: I'm unclear.  Does having IMP able to run as a "flux" user and run shells on behalf of guest users eliminate the need to run _any_ Flux process as root?  - assuming that root does the appropriate system setup as described, of course.  It so, that's a big deal that could be added to the benefit bullets:
* Code running under root privilege is eliminated.

spec_15.adoc
morrone: This paragraph is correct...maybe it isn't entirely clear that "instance owner" is intended to be an unprivileged user?

Since the instance owner is unprivileged, we can't allow it to go around running anything it wants as either people.  Those other people (the guests) need to give their authorization for things to be run as them.

Now that is not the _only_ authorization that is needed at the IMP.  The IMP also needs authorization from the resource owner.  That part of the explanation is left until later.  Maybe it should be quickly mentioned here as well to make things clearer?

spec_15.adoc
morrone: There won't be any need to have long-standing flux daemons running at root, but the IMP may take flux-provided plugins that run in privileged context.  So we might not want to overstate the elimination of flux code that runs as root.

spec_15.adoc
```diff
@@ -0,0 +1,358 @@
...
+ * arguments
+ * environment
+ * processing of marshaled job request *_R_* on stdin
```
morrone: I know that we discussed this in a meeting, but I am back to not understanding why this isn't a gaping security hole.  If the processing of the marshalled job request is done as the instance owner, what stops the instance owner from attaching to that process and changing what it does?  At this point, can it not tell the IMP "the guest asked us to run rm -rf in their home directory"?  I get running as much code as possible in unprivileged mode, but it doesn't seem same to trust any processing if it is done as the instance owner.  Maybe it could run as yet _another_ unpriviledged user that has access to nothing on the system.  Perhaps "nobody", or if that isn't kosher, then another user created explicitly for that purpose such as "dummyimp".

grondo: @morrone, this is a good point. The unprivileged part of the privsep code would only have to run as the instance owner if it had to connect to the instance for some reason. I think you made a good argument that this should be necessary (nor encouraged?), so perhaps a `nobody` user would work.

However, it should be noted that the privileged part of the IMP would do the verification of *R* and *J* so if the unprivileged part attempts to modify them, that should be detected. The details of how this works still needs to be nailed down.

spec_15.adoc
grondo: Good catch thanks! Should be "invoker", e.g. the user running the IMP, which will be the instance owner under normal circumstances, but any user can *attempt* to invoke the imp process and feed it data in an attempt to exploit a vulnerability.

spec_15.adoc
grondo: Heh, thanks, forgot when I added *system owner* to fix that.

spec_15.adoc
grondo: @morrone, @lipari, good ideas. I can add "*unprivileged* instance owners" instead of just "instance owners". 

> Now that is not the only authorization that is needed at the IMP. The IMP also needs authorization from the resource owner. That part of the explanation is left until later. Maybe it should be quickly mentioned here as well to make things clearer?

I thought authorization from the resource owner was implied here since the instance owner is invoking the IMP to run work for the guest user. However, I'm completely open to rewording.

What about something like:

> In the Flux system, this privilege is contained within a single service _The Independent Minister of Privilege_,  which is responsible for allowing *unprivileged* resource owners to execute processes of a job as another *unprivileged* guest user. This privilege is only extended when the guest has authorized the instance owner to execute the job in the form of a job request, and the instance owner has authorized the job to be run.

lipari: The model that I have, that is causing me confusion, is this:  The guest requests to run a shell from the instance owner.  The instance owner grants the request and spawns a shell which is owned/run by the guest.  That's the model I'm expecting coming into the text, and I haven't quite adjusted to what is really being described.  So the guest doesn't _authorize_ the instance owner to execute the job, it _asks_ the owner to run the job.

garlick: That's the net result, only in this model the instance owner doesn't have the privilege to run things as other users _directly_.  That is outsourced to the IMP.

The IMP, as a standalone entity requires authorization from the resource owner as well as the guest in order to run the shell as the guest on the resources.

This architecture achieves the Flux goals of allowing the instance owner to fully customize and control the instance, without requiring the instance owner to be granted any special system privileges (just access to the resources - possibly granted through a sequence of nested instances back to the system instance).

spec_15.adoc
garlick: > However, it should be noted that the privileged part of the IMP would do the verification of R and J so if the unprivileged part attempts to modify them, that should be detected.

Wouldn't it need to verify that the parsed results were _derived_ from *R* or *J*?

grondo: It depends on the mechanism doesn't it? The details are missing here, but if, say, the input to the IMP was in some serialization format, the serialization would need to be processed before *R* is verified. We don't want the de-serialization library running as root, so we offload that to an unprivileged process. Once the data is deserialized into a format where authenticity and integrity can be verified, this is done by the privileged process. This ensures that even in the unlikely event the owner is allowed to `ptrace` the unprivileged child process, if there is an attempt to modify data after parsing, those modifications are detected.

spec_15.adoc
grondo: I'll just add that this RFC is specifically about the requirements of the IMP as a core piece of a working multi-user Flux system. It isn't so important to the IMP that the user at some point "asks" the instance to run something, but that the instance owner, who is requesting to run something *as another user*, have the specific authority to do so, as granted by the guest user's signed request (which indicates exactly what to run, and by whom those processes can be started (the *intended recipient*)).

spec_15.adoc
grondo: BTW, under normal circumstances, any process which has called setuid() is not ptrace-able, so even if we do run the unprivileged child as the instance owner, that user should not be able to attach to the process, cause it to coredump, etc. This is because any time a process (in linux at least) changes uid or gid, the `PR_DUMPABLE` flag is set to 0 (unless overridden in /proc), which not only disables coredumps but also ptrace access to the process from any user other than root.

See the entry from `prctl(2)` for the `PR_SET_DUMPABLE` op:

> PR_SET_DUMPABLE (since Linux 2.3.20)
              Set  the  state  of  the flag determining whether core dumps are
              produced for this  process  upon  delivery  of  a  signal  whose
              default behavior is to produce a core dump.  (Normally this flag
              is set for a process by default, but it is cleared when  a  set-
              user-ID  or set-group-ID program is executed and also by various
              system calls that manipulate process UIDs and GIDs).  In kernels
              up  to  and  including 2.6.12, arg2 must be either 0 (process is
              not dumpable) or  1  (process  is  dumpable).   Between  kernels
              2.6.13  and 2.6.17, the value 2 was also permitted, which caused
              any binary which normally would not be dumped to be dumped read-
              able  by  root only; for security reasons, this feature has been
              removed.       (See      also      the      description       of
              /proc/sys/fs/suid_dumpable in proc(5).)

grondo:
The most recent push guts the privilege separation section and replaces it with a note that privsep *MAY* be used if appropriate, and references the original OpenSSH paper on privilege separated ssh.

garlick:
This seems like excellent progress.  I propose we merge this and take up further discussion in new PR's/issues.

grondo:
OK, allow me to make one editing pass then squash all intermediate commits

grondo:
Ok, everything squashed down now

garlick:
Great!  Merging.

grondo:
Thanks!

lipari:
@grondo, sorry for not catching this sooner... You're missing an entry for the new RFC in the README.

grondo:
Thanks @lipari, completely forgot about that!

Issue #72, opened 2017-02-03
rfc3, rfc12: add uid to messages


garlick:
RFC 12 does not currently describe mechanisms used by a multi-user Flux instance, although it includes the statement that users other than the instance owner shall not be allowed to make a direct broker connection.  That is one approach to securing a multi-user instance - requiring new multi-user facing services, connectors, etc. to be implemented that act as secure bridges to internal services.  It is probably not desirable because reserving all "internal" communication to the instance owner unnecessarily partitions the code into "internal" and "external" parts, and may result in duplication of infrastructure like connectors and (likely) poor modularity.

Another approach is to allow other users to connect to the broker, and to add an integer user id field to the base message format (PROTO block, RFC 3).  When a connection is accepted by the connector, some authentication would take place; for example, the local connector uses SO_PEERCRED to authenticate connections.  The uid determined during authentication would be "stamped" on every message flowing from that connection into the broker.

In addition to the uid, we could add a "capability" bitfield to the PROTO block similar to [POSIX capabilities](http://man7.org/linux/man-pages/man7/capabilities.7.html).  At connect time, a user's capabilities could be looked up in an instance table of some kind and set in every message (like the uid above). The equivalent of CAP_SYS_ADMIN could be assigned to the instance owner.   Other capabilities could be added for listing the queue, submitting jobs, killing jobs, etc..

When the broker dispatches requests to services, it could by default require CAP_SYS_ADMIN otherwise return an EPERM response.  Event and response messages lacking CAP_SYS_ADMIN could be dropped.  If a service wants to allow other users "in", they could register additional capabilities with the broker.

I propose to submit PR's to RFC 3 and RFC 10 to implement uids and capabilities in Flux as a first step towards a multi-user security architecture.  Any thoughts?

grondo:
In general this seems like a clean first addition of multi-user functionality to the protocol. I do have some questions, probably more due to my lack of knowledge than any missing pieces in this proposal:

 * How do the instance CURVE keys figure into this plan? I kind of assume that messages would still be encoded with the CURVE key of the instance owner, and the API or other socket providing user access would have access to the key because it is a service running as the instance owner offering access to other users.

 * If capabilities are per user, there is no strict need to have a capability field per *message* right? I can still see the benefits of having a per message capability, since this simplifies code on the *receiving* side, but did you consider and reject the idea of leaving capability lookup as a broker service? (It seems like you made the right choice to me, but it would be interesting to hear the argument against )

[...]

PR #80, opened 2017-03-08
rfc12: add description of userid, rolemask based security

garlick:
This PR amends RFC 12 with the userid, rolemask, and user database concepts discussed in #72 and prototyped in flux-framework/flux-core#980.

I stated right at the beginning the assumption that POSIX UID's be synced across systems spanned by a Flux instance even though we discussed the possibility in #72 that non-synchronized systems might need to be handled at some point.  That's not strictly built in to the design, but it is assumed in a few places now, e.g. the local connector authenticates with SO_PEERCRED (directly mapping uid to Flux userid), and some connectors compare `geteuid()` with the Flux userid to determine whether it's the instance owner.   The fact that we have defined a "Flux userid" separately from the POSIX UID most places except at these ingress points should make that change tractable later on.

The only other thing that's slightly tricky is that we have a "setuid" feature of sorts where the instance owner can dummy up messages with any old userid, rolemask values for testing.  Since the instance owner is effectively omnipotent in the instance, even controlling the code, this should be safe.  It also turns out to be necessary to allow connectors to be implemented as comms modules.

### DIFF
--- a/spec_12.adoc
+++ b/spec_12.adoc
@@ -93,19 +93,111 @@ could extend job cancellation to anyone with the "user" role.
 
 == Implementation
 
+=== Assumptions ===
+
+Users of a Flux instance MUST be assigned the same POSIX UID on all systems
+spanned by the instance.
+
+=== User Database ===
+
+The optional Flux user database contains entries for each user of a
+a Flux instance.  A valid entry SHALL contain a 32-bit _userid_ key,
+corresponding to the user's POSIX UID;  and a 32-bit _rolemask_, which
+assigns one or more privileges to the user.  A userid with no assigned roles
+SHALL be considered invalid.
+
+A single-user Flux instance MAY load the user database.  If the user database
+is not loaded, then only the instance owner SHALL be allowed to access the
+instance.
+
+A multi-user Flux instance SHALL load the user database.  Only users in
+the database SHALL be allowed to access the instance.
+
+FLUX_USERID_UNKNOWN (2^32^ - 1) SHALL be reserved to indicate "invalid user".
+
+FLUX_ROLE_NONE (0) SHALL indicate "invalid rolemask".
+
+FLUX_ROLE_OWNER (1) SHALL confer the maximum privilege upon the user,
+and is REQUIRED to be assigned to the instance owner.
+
+FLUX_ROLE_USER (2) SHALL confer access, but no administrative privilege
+upon the user.
+
+Other role bit definitions are TBD.
+
+=== Connector Security ===
+
+Flux connectors SHALL authenticate each connection, mapping it to a valid
+Flux userid and rolemask, or rejecting it.
+
+If the authenticated userid can be proven to be the instance owner without
+accessing the user database, the connection SHALL be allowed, and assigned
+a rolemask of FLUX_ROLE_OWNER.
+
+If the user database is loaded, and the authenticated userid has a valid
+entry, the connection SHALL be allowed, and assigned the looked up rolemask.
+
+Other connections SHALL be denied.
+
+As indicated in RFC 3, Flux messages have a userid and rolemask field.
+In messages received en route to the broker, the connector SHALL rewrite
+these fields from the expected values of FLUX_USERID_UNKNOWN and FLUX_ROLE_NONE
+to the authenticated userid and rolemask.
+
+If the user is not authenticated with FLUX_ROLE_OWNER, and a message is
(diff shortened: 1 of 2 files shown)


grondo:
This looks great. We might want to generalize the description of the user database, since there could be multiple implementations with varying complexity -- from the simple version seen here, to an implementation that might need to also contain user credentials such as a valid public certificate, bank and account information, or backed by LDAP or something... Though I guess nothing in the current RFC states that *only* the data described shall be in the userdb, so if you feel current language is sufficient, I can accept that.

Does the UID requirement as stated in the current RFC put any burden on system maintainers to ensure that users have /etc/passwd or similar information on hosts where that configuration would not currently be required? For example, for a system instance running on RPS and management nodes, I assume users will need to be configured on these hosts, where they may not have had even an account before. (I hope that made sense). I guess it is not an onerous requirement, and the users account could be configured but disabled, but I thought to mention it as a possible drawback to not completely separating Flux and system users.

garlick:
I'll see if I can make it a bit more clear that the design of the userdb should _at minimum_ perform the required function of mapping userid's to rolemasks.  That was my intent.

On the UID requirement  - the main point of the stated assumption was that users need to be assigned the same numerical UID on each system so we don't confuse identities within the instance.

On knowing where a user is valid, should it be
* stored (e.g. in the form of a nodeset?) in the userdb, per user
* a function of the resource db to designate some nodes as off limits to users (e.g. RPS case), and we presume the same set on the rest?
* up to an IMP plugin to verify that a user has an active, non-system account (is PAM the main way to do that?) prior to starting processes as that user?

It does seem like you'd like to know prior to launching a program what nodes a user has access to.

grondo:
Sorry I didn't mean to conflate the resources to which a user has access for programs, and valid users for an instance. I think you're correct that the resources to which a guest user has access is a function of the resource/scheduler configuration, and is probably not germane to this RFC.

I was specifically thinking about how a guest user would access an instance on which their UID happens to not be in /etc/passwd. Specific example: if you have a system instance with brokers only on management nodes, do you have an extra burden of adding all users to local passwd file (with disabled access) so that they can proxy in to the instance, or that their UID is valid where ever a broker runs?

garlick:
Authentication is only performed by connectors, so I guess as long as the userdb contains your userid and you can authenticate to one connector, you are a valid user in the instance even if not present in /etc/passwd.  That may not mean much if you can't run anything.

Currently the only way for non-instance owners to authenticate is via local:// which uses SO_PEERCRED, which implies some access to the host (enough to start a local process).

The ssh:// connector requires you to start `flux-proxy` on the remote host, which then connects via local://, so same thing.

Any future connector that provides remote access would need to do some work to map a user's credentials to a local uid.  For example a connector that authenticates kerberos would ( think) consult /etc/krb5.conf to map principal@realm to a local user, and then the connector would need to map that to a uid, which would get mapped (directly, as currently designed) to a Flux userid.  So again, password file entry required at connector ingress.

Do we need to call any of that out in the text added in this PR?

grondo:
> Do we need to call any of that out in the text added in this PR?

I don't think so, however I did want to think through how the implementation would look for some of the system instance designs we've been proposing. I was also pondering how a center-wide instance, which might only be actually running on a cluster of protected systems, would actually implement multi-user access. I was assuming in that case sites won't want to give all users ssh access to those nodes, so an alternate remote connector would be required. I was thinking it would be nice if administrators could add users to these system instances through some other means than making them valid users on the nodes of the instances. However, possibly standard Identity Mangement services could be used here like LDAP, so I'm probably making something out of nothing. Sorry.

grondo:
> Authentication is only performed by connectors, so I guess as long as the userdb contains your userid and you can authenticate to one connector, you are a valid user in the instance even if not present in /etc/passwd. That may not mean much if you can't run anything.

I guess what I was thinking is that, since an instance in distributed, you might be a valid user on some resources the instance spans, and not others. This doesn't mean you can't run anything just because you don't exist in /etc/passwd on the system to which you've *connected*.

garlick:
I think you are right, authorization to use Flux via a given system could be separate from authorization to login/rsh to the system.  A future connector would just need a way to map a user in a site security realm to a Flux userid in the Flux userdb, potentially using an external service like LDAP (which I believe contains UNIX UID's at our site?) or even via a new key in the Flux userdb should we want to add one.

The userdb could even be implemented wtih an LDAP back end.

grondo:
OK, great. Apologies, I was kind of muddling through that thought process and you've now made it clear

PR #81, opened 2017-03-09
rfc 4: generalize our model

dongahn:
- Generalize the Flux Resource Model using two basic concepts: resource pool and graph.

- Make the composite resource pool as a specialization of this model

- Add abstract interfaces


dongahn:
Oops. Let me resolve this conflict and do another push.

dongahn:
OK. This should be ready for your review. @lipari, @tpatki and @SteVwonder have provided great feedback and I incorporated most of them into this revision, though I understand @lipari is having his second round!

morrone:
@dongahn, do you have source files or .svg for those images by any chance?  It would be good to check those in too so the images can be edited in the future.

dongahn:
@morrone: good point. Yes, I can export them to svg format. I will check them in.

grondo:
@dongahn, nicely done, thanks!

I wonder if the figure for composite resource graph could benefit from multiple links from the rack object to many nodes, just to give a visual hint that the composite hierarchy? Also, it is explicit that a composite resource instance has 0 or more children (in has-a relationship), but should it also be mentioned that a composite may have at most 1 parent?

Also, does a mention of the resource class system belong here, i.e. all resource pool data objects are instances of a resource type which ultimately inherits from a "Resource" base class. Static data common to a resource type may be associated with the resource class instead of propagated to the resource instance. Maybe this is too far into implementation, but the composite resource model + class-based type system together lends itself to easily constructing large, complex composite resource, and is perhaps an important part of the overall conceptual design.

dongahn:
>I wonder if the figure for composite resource graph could benefit from multiple links from the rack object to many nodes, just to give a visual hint that the composite hierarchy? Also, it is explicit that a composite resource instance has 0 or more children (in has-a relationship), but should it also be mentioned that a composite may have at most 1 parent?

@grondo: good idea! I will work on this. 

> Also, does a mention of the resource class system belong here, i.e. all resource pool data objects are instances of a resource type which ultimately inherits from a "Resource" base class. Static data common to a resource type may be associated with the resource class instead of propagated to the resource instance. 

I think it does. This should be a key resource pool data model. We do have this in RDL, but not in resrc, yet though... So the question is, do we want to flesh out details on this point now? Or do we want to put a placeholder now and then later fills out details when we learn more about this from the "the upcoming Resource API? My preference is latter, but I can go either way. 

> Maybe this is too far into implementation, but the composite resource model + class-based type system together lends itself to easily constructing large, complex composite resources, and is perhaps an important part of the overall conceptual design.

Great point. I just want to add, though, this now goes beyond "composite": "graph model" + "class-system based resource pool model" allow us to describe "any" resources at any levels of details.

lipari:
>>  Also, does a mention of the resource class system belong here, i.e. all resource pool data objects are instances of a resource type which ultimately inherits from a "Resource" base class. Static data common to a resource type may be associated with the resource class instead of propagated to the resource instance.

> I think it does. This should be a key resource pool data model. We do have this in RDL, but not in resrc, yet though... So the question is, do we want to flesh out details on this point now? Or do we want to put a placeholder now and then later fills out details when we learn more about this from the the upcoming Resource API design? My preference is latter, but I can go either way.

While many developers will see the advantages of implementing resource pools in an OO design of base class + a derived class, I question whether we need to impose such a design in the RFC.  What would be lost if a developer chose to implement the resource pool's data members with a flat structure?

grondo:
> While many developers will see the advantages of implementing resource pools in an OO design of base class + a derived class, I question whether we need to impose such a design in the RFC. What would be lost if a developer chose to implement the resource pool's data members with a flat structure?

This point is why I was kind of wondering aloud if the concept belongs in this document. I could conceive of the type system as part of the conceptual design (e.g. the RFC already defines the members and suggested methods of a base resource pool type), however I also see your point that it is not *necessary* to specify a type system here (though in my opinion they are strongly correlated)

dongahn:
> While many developers will see the advantages of implementing resource pools in an OO design of base class + a derived class, I question whether we need to impose such a design in the RFC. What would be lost if a developer chose to implement the resource pool's data members with a flat structure?

I guess one argument would be on easier support for "abstract interfaces." Unless we have something like this, writing abstract interfaces on resources of any type can be difficult. If we want to use this RFC to make certain recommendations to the implementors, it seems to make sense to includes class-system as part of resource pool data and abstract interfaces. 

However, I didn't feel like I understand fully what are the needed class-systems, so I wanted to defer fleshing out its details at least until we have a reasonable resource API design.

Just so that you know where I'm coming from, C++ was my language of choice before, so I'm apparently biased ;-)

grondo:
> I just want to add, though, this now goes beyond "composite": "graph model" + "class-system based resource pool model" allow us to describe "any" resources at any levels of details.

Yes, good point. Though it still feels to me like the default graph for compute resources will likely be a "composite hierarchy" though maybe I'm just stuck in my own way of thinking. It may be much easier for task launcher if there was some kind of requirement for a default populated topology of compute resources at least.

dongahn:
> Yes, good point. Though it still feels to me like the default graph for compute resources will likely be a "composite hierarchy" though maybe I'm just stuck in my own way of thinking. It may be much easier for task launcher if there was some kind of requirement for a default populated topology of compute resources at least.

Understood. 

For full disclosure, what I wanted to capture was, a graph is now a general concept such that a graph can contain multiple hierarchies (default, power, io...) in it. It is not like we have separate graph instances.On this graph, you can name the hierarchy to use and traverse from, and we can have some naming scheme to nominate a hierarchy as default. 

I am ambivalent whether we want to add this point to this model RFC, though. 

I am thinking this may fit better in the RFCs of the other services that use and refers to this Resource Model RFC. Certain power-aware scheduling may want to use "power hierarchy" as the dominant hierarchy to traverse from and choose nodes/cores supported by certain PDU vertices... I can see this can be described in the RFC of that scheduling strategy.

dongahn:
>@dongahn, do you have source files or .svg for those images by any chance? It would be good to check those in too so the images can be edited in the future.

Well, I don't think this machine-generated .svg files are any better for future editing. I used Neo4j graph database to visualize this. I will just check in my cypher commands as a text file, instead.

garlick:
If scriptable, maybe add a makefile target?

dongahn:
@garlick: no, unless neo4j db is running on the system. Let me see if I can redo the graphs with graviz -- dependency there would be only the `dot` command.

dongahn:
Actually, graphviz didn't give me the layout I wanted so I gave up on it. It turned out the .svg files appeared okay after I formatted it with an xml formatter so I check those in. 

@garlick: I used `convert`  that comes with ImageMagic like

`$ convert graph.svg graph.png`

So we can put this into our `Makefile.` But I didn't do this because ImageMagic probably is not in our dependent package list. 

```
convert --version
Version: ImageMagick 7.0.5-0 Q16 x86_64 2017-03-09 http://www.imagemagick.org
Copyright: © 1999-2017 ImageMagick Studio LLC
License: http://www.imagemagick.org/script/license.php
Features: Cipher DPC HDRI Modules 
Delegates (built-in): bzlib cairo fontconfig freetype jng jpeg ltdl lzma png rsvg tiff xml zlib
```

dongahn:
As part of the last push, I addressed all of the feedback so far, which is not contentious.I will create an issue to consider adding a point about the resource class-system, once we understand this point better as we make progress on the design of the new Resource API. 

Please let me know if there is anything else I need to consider as part of this PR.

spec_4.adoc
```diff
@@ -47,65 +47,54 @@ link:spec_14{outfilesuffix}[14/Canonical Job Specification]
...
+The second concept borrows from graph theory to describe relationships
+among such individual resource pools. A graph consists of
+a set of vertices and edges, where each vertex represents
+an individual resource pool and an edge a certain relationship
+between them.
+
+Each edge SHALL be either directional or bidirectional
+and have a type as well as a subsystem name such that the union
+of the set of all edges with a same name and the set of all vertices
+connected by these edges MAY represent a unique subsystem
```
dongahn: Will revise this to SHALL.

morrone:
There are some out-of-sequence headings throwing an error in this patch:

  asciidoc: WARNING: spec_4.adoc: line 78: section title out of sequence: expected level 2, got level 3

Also, the "{sp} +" thing that you are using breaks the pdf generation.  Could you please remove that?

And finally, is there any reason to not just use the .svg files as the images directly?  That avoids having fuzzy images in the pdf files and removes a step in the build.

dongahn:
> There are some out-of-sequence headings throwing an error in this patch:
> asciidoc: WARNING: spec_4.adoc: line 78: section title out of sequence: expected level 2, got level 3

Thanks. Will address.

> Also, the "{sp} +" thing that you are using breaks the pdf generation. Could you please remove that?

This was to give enough vertical space between each image and the text. Maybe there is a better way to force the spaces. I will look.

> And finally, is there any reason to not just use the .svg files as the images directly? That avoids having fuzzy images in the pdf files and removes a step in the build.

Direct replacement didn't work yesterday, and I just assumed .svg isn't supported. Do you know if this format is supported by asciidoc?

morrone:
> This was to give enough vertical space between each image and the text. Maybe there is a better way to force the spaces. I will look.

The vertical space looks fine to me without it.  I think trying to force spacing in the .html (which will look different in every browser anyway) by inserting {sp} in the asciidoc is going to be a fraught with peril.  I think that sort of thing is supposed to be handled by style sheets rather than the content material.

> Direct replacement didn't work yesterday, and I just assumed .svg isn't supported. Do you know if this format is supported by asciidoc?

It worked fine for me on my machine.  Where did you have the problem?  I kind of suspect that asciidoc just passes the file name through with minimal checking.  It is up to the various renderers to decide if they like it.  For .html, I would hope that in 2017 most browsers can handle .svg?  And the a2x seemed to handle it fine in generated a pdf for me too.

spec_4.adoc
```diff
@@ -47,65 +47,54 @@ link:spec_14{outfilesuffix}[14/Canonical Job Specification]
...
-a tree graph to capture a hierarhical relaitionship among resources,
-more complex graph forms MAY be supported. An individual resource
-MAY belong to one or more distinct vertices.
-
-==== Resource Pool Data
+individual resources and various relationships among them.
+
+The concept to describe each individual resource is called
+_resource pool_. A resource pool allows the grouping of 1
+or more _indistinct_ resources into a pool with a given size.
```
grondo: Would it make more sense if a _resource pool_ was defined as a *collection* of indistinct resources, with a size indicating the number of individual resources represented by the collection (Examples: size = 1 for a single core or node, size = 1024 for a 1G DIMM where each resource is 1M of RAM)?

Yes, a pool size of 1 is the degenerate case, but important since this is how most compute resources would actually be represented (clusters, nodes, sockets, cores, threads).

Conceptually, the pool concept just allows us to "roll up" indistinct resources (like bytes of RAM or disk space, licenses, joules of energy), into a single vertex of the resource graph, when we don't care about identifying each resource individually.

dongahn: @grondo: yeah the concept is essential. It just how to define it in a way so everyone understands it clearly. 

Yeah, seems "collection" will help. Examples and "different levels of details (LOD)" statement were my attempt to clarify this a bit better, base on @tpatki's feedback. 

I will try to have more clarity for this in my next push.

dongahn:
> It worked fine for me on my machine. Where did you have the problem? I kind of suspect that asciidoc just passes the file name through with minimal checking. It is up to the various renderers to decide if they like it. For .html, I would hope that in 2017 most browsers can handle .svg? And the a2x seemed to handle it fine in generating a pdf for me too.

On my mac (OS X El Capitan) both with Safari and Firefox. On my Android both with Chrome and default browser.  

>  I kind of suspect that asciidoc just passes the file name through with minimal checking. 

Yeah, this would be my guess because at least Safari, Firefox, and Android can render .svg files by themselves. It just that they can't render them when .svgs are embedded in asciidoc.

dongahn:
@morrone: when you have a chance, could you please check to see if these png files render a better pdf. If this still doesn't cut it, I will take your advise and just use .svg files directly at the expense of some readability issues on those who browse directly off of GitHub. Revising the resource pool is still a do-to.

dongahn:
OK. My edit in the resource pool is pushed. Let me know if there is anything else you folks want me to look into.

morrone:
> when you have a chance, could you please check to see if these png files render a better pdf.

Yes, that looks acceptable.

morrone:
@dongahn , I do think this revision is more clear.  Although, I am not sure that I would know what the phrase "fully promoted resource pool" would mean without our discussion on Friday.

dongahn:
Great point. I guess that is what parentheses are for. I will add an inline definition on "fully promoted pool"

lipari:
I'm also not entirely comfortable with the description of the term "pool".  It fits for something like memory, but when describing something like a node, it is a bit of a stretch to call a node a pool of size == 1.

Even more confusing, the pool definition says "Each resource in a pool cannot be individually identified", yet I'm sure people will think of a node or core (valid resource pools of size == 1) and think, "a node can be node5, a core can be core3, etc.  Does that mean a node and core are _not_ pools?"

I need to mull this over and see whether there is a better term that will apply equally well to a resource "pool" as well as a resource atom.

On the other hand, we've spend a lot of time on this RFC.  I'm perfectly fine is this group is ready to call it a pool and move on.

grondo:
> I'm also not entirely comfortable with the description of the term "pool". It fits for something like memory, but when describing something like a node, it is a bit of a stretch to call a node a pool of size == 1.

> Even more confusing, the pool definition says "Each resource in a pool cannot be individually identified", yet I'm sure people will think of a node or core (valid resource pools of size == 1) and think, "a node can be node5, a core can be core3, etc. Does that mean a node and core are not pools?"

Great points, @lipari. I think the concept is good here, though the vocabulary may be confusing.

This is probably obvious, but resources in a pool are not distinguished *from eachother*. The pool itself is a distinct resource, and therefore a pool of size == 1 by definition has one distinct resource. As I think @dongahn pointed out, we only call the degenerate case a "pool" so that we can treat all vertices in the graph model as the same kind of thing.

Perhaps it would be enough to say that a degenerate pool (a pool with a single resource, size == 1) may also be referred to as a single resource, however in terms of the model all resources shall be treated as pools to avoid special case semantics.

dongahn:
Good point both. I will add this adjective to the degenerate case as simply referring it as "degenerate pool," and see how it looks.

I am in favor of not special casing n=1. Actually I do think an elegant solution should use a general concept as much as possible.

An array of size 1 should still be called a valid array right? :-)

lipari:
@dongahn , please proceed with your updates.  The array analogy is very appropriate.

dongahn:
Done. Changes pushed.

lipari:
The latest changes look good to me.  You have my vote to merge.

spec_4.adoc
```diff
@@ -112,45 +118,230 @@ properties (e.g. a system might have node names formatted like
...
+=== The Channeled Resource Pool
+
+As HPC centers are becoming increasingly data- and power-constrained,
+the Flux Resource Model MUST be flexible to be able to model
+how data and/or power flow through its distribution units (e.g.,
+a high performance switch for data and a power distribution unit for power).
+One specific form of the Flux Resource Model to represent
+the notion of a flow is called _channeled resource pool_. Here,
+two resource pool instances, each representing a distribution capacity
+of a flow, are related under a _channel-of_ or _conduit-of_ relationship.
```
trws: Since these are just resources connected with edges, is there a real difference between these and resource pools?  In the model we discussed I had been under the impression that the purpose of flow resources was to extend the concept of a pool to a flow *between* resources, and thus as a property on an edge.  This leads me to believe that it's a property of a set of vertices alongside normal resources, while the current resrc code seems to store them as additional pools on existing resources by adding extra resource copies.  I'm not sure how to reconcile these.

dongahn:
Not much difference at all, Tom. This is named differently than the above composite type just to show the same concepts can be used to describe different kind of relationships (beyond parent-children relationship). No relationship info goes to the vertices.

lipari:
As I work on creating the scheduling methods RFC, I have found that there is predominant order to the opening sections of all of our RFCs.  It typically goes:  Language, Related Standards, Goals.  Then it diverges into headings like Background, Introduction, Overview, and Implementation.

I'd like to suggest that you reorder the headings in this PR to conform follow the established order:  Language, Related Standards, Goals and have Background be the fourth heading.

dongahn:
@lipari: I like that. I will do this when I circle back to this.

dongahn:
@lipari and all: Ok, I reordered the sections. As part of that I also fixed the broken link to the Canonical job spec. Looks like this is ready to be merged?

grondo:
@lipari, go ahead and merge. Thanks.

PR #84, opened 2017-03-21
rfc14: Change static references to asciidoc footnotes

morrone:
Turn the references in RFC14 into asciidoc footnotes so that the
footnote numbers are clickable.

Also update the URLs for the two changes references to point to more
static locations for the referenced location rather than "current" which
makes the linked-to page change over time and not match the version listed
in our document.

### DIFF
--- a/spec_14.adoc
+++ b/spec_14.adoc
@@ -41,7 +41,9 @@ may be generated from other user-friendly forms or interfaces.
 == Overview
 
 This RFC describes the canonical form of "jobspec", a domain specific
-language based on YAML[1].  The canonical jobspec SHALL consist of
+language based on YAML
+footnote:[http://yaml.org/spec/1.1/current.html[YAML Ain't Markup Language (YAML) Version 1.1], O. Ben-Kiki, C. Evans, B. Ingerson, 2004.]
+.  The canonical jobspec SHALL consist of
 a single YAML document representing a reusable request to run
 exactly one program.  Hereafter, "jobspec" refers to the canonical
 form, and "non-canonical jobspec" refers to the non-canonical form.
@@ -102,7 +104,8 @@ program. The dictionary MUST contain the keys `resources`, `tasks`,
 
 Each of the listed jobspec keys SHALL meet the form and requirements
 listed in detail in the sections below. For reference, a ruleset for
-compliant canonical jobspec is provided using *JSON Content Rules*[2]
+compliant canonical jobspec is provided using *JSON Content Rules*
+footnoteref:[contentrules,https://tools.ietf.org/id/draft-newton-json-content-rules-06.txt[JSON Content Rules (jcr-version 0.6)]&#44; A. Newton&#44; P. Cordell&#44; 2016]
 in the *Content Rules* at the end of this section.
 
 === Resources
@@ -340,7 +343,7 @@ attrs:
 === Content Rules
 
 A jobspec conforming to version 1 of the language definition SHALL
-adhere to the following ruleset, described using JSON Content Rules[2]
+adhere to the following ruleset, described using JSON Content Rules footnoteref:[contentrules]
 draft version 0.6.
 
 ----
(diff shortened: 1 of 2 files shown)


grondo:
At least with GitHub rendering this puts the "footnotes" into a separate section (below an hbar) after the references section. Do you know if there is a way to keep referenced and unreferenced bibliography together in the refs section?

It is much nicer that the referenced footnotes have a link, but could that also be done with anchors? (I'm actually fine with the changes, they are an improvement)

garlick:
As a general style for our RFC's, would it be reasonable to drop the "references" sections and just go all in with the footnote style?

In this RFC that would leave the JDSL ref dangling so presumably we'd want to add some text that can cite it? It's probably bad form to have uncited references anyway, so maybe we could just add a "related work" section under Overview and give JDSL a nod there?

lipari:
> so maybe we could just add a "related work" section under Overview and give JDSL a nod there?

As the one who added the JDSL reference, I wanted to acknowledge it as something we were aware of.  So, yes, a "related work" reference sounds fine to me.

grondo:
That's fine with me. In this case I think the JSDL reference could be dropped if that's ok with @lipari. I would hesitate to add it to a related work section unless we did an honest survey of other related work.

As a general question, do you know how to reference the same footnote multiple times with footnote style references?

grondo:
Oops, @lipari I missed your comment. I'm fine with keeping JSDL in related work if you'd prefer.

garlick:
@morrone has an example of that with the `footnoteref:[contentrules]` citation.

grondo:
ok thanks I missed that at first read

grondo:
My only remaining hesitation is that the documents do not read very well in text form with the footnote style references. However, I haven't tried anchors to see if it is any better, and it really isn't that important to me.

spec_14.adoc
```diff
@@ -102,7 +104,8 @@ program. The dictionary MUST contain the keys `resources`, `tasks`,
 
 Each of the listed jobspec keys SHALL meet the form and requirements
 listed in detail in the sections below. For reference, a ruleset for
-compliant canonical jobspec is provided using *JSON Content Rules*[2]
+compliant canonical jobspec is provided using *JSON Content Rules*
+footnoteref:[contentrules,https://tools.ietf.org/id/draft-newton-json-content-rules-06.txt[JSON Content Rules (jcr-version 0.6)]&#44; A. Newton&#44; P. Cordell&#44; 2016]
 in the *Content Rules* at the end of this section.
```
grondo: I'm not sure what's wrong with this footnote exactly, but it isn't rendering correctly on GitHub, see https://github.com/morrone/rfc/blob/rfc14_changes/spec_14.adoc

lipari:
> I'm fine with keeping JSDL in related work if you'd prefer.

Yes, I prefer we keep it.  It preemptively answers the question, "why didn't you just use JSDL - it's very similar."

spec_14.adoc
morrone: GitHub's asciidoc rendering seems less than perfect. There were issues with image inclusion as well.

morrone:
Currently the reference only tells folks that we have heard of it.  It doesn't explain why we didn't just use it.  I certainly didn't understand why it was listed as a reference...I thought maybe it was inspiration for the format.

lipari:
>  I certainly didn't understand why it was listed as a reference.

JSDL is the best effort I am aware of for defining a universal job specification that could be used with any scheduler.  Some commercial schedulers have support for JSDL.  So why didn't we just use it instead of inventing our own?  Well, one answer is that we were looking for a specification that was uniquely tuned to the Flux resource and job execution models.  If you'd like, you can leave the reference as-is and I will submit a follow-on PR that provides a context for the JSDL reference.

morrone:
That sounds good to me.  The follow-on will then be able to convert it into a footnote as well.

lipari:
@trws, as the owner of this RFC, please merge if you approve.  It looks good to me.

grondo:
`contentrules` and `footnoteref`, sadly, have to be added to the spellcheck dictionary.

trws:
I'm certainly good with this, just need the update to pass checks.

lipari:
@morrone, you're up.  Please rebase this PR and I'll follow with the context for the JSDL reference.

garlick:
This never got rebased and branch is gone so I'll resubmit as a new PR.

PR #87, opened 2017-03-23
rfc14: Overhaul examples to better match the canonical jobspec

### LINKED ISSUE #83: rfc14: resource count examples do not match resource count definition
The examples in rfc14 all seem to be using "count" incorrectly in the "resources" sections.  According to the RFC's definition, "count" must be a dictionary that must contain four items: min, max, operator, and operand.  But all of the examples show count being a single integer.

Do we want to change the definition to allow what is shown in the example, or do we want to change the examples to match the definition?

### LINKED ISSUE #86: rfc14: Invalid count range representation in use case 1.2
Use Case 1.2 does the following:

```
resources:
  - type: node
    count: "[3:30]"
```

This range representation does not appear to be valid according to the jobspec definition.

morrone:
Overhaul the examples to better match the canonical jobspec, so we
have a foundation to debate and improve the jobspec.

Honor the rule, "The dictionary MUST contain the keys `resources`, `tasks`,
`walltime`, `attrs`, and `version`" by adding missing keys.

Eliminate the assumed "tasks" section in the examples.  Instead directly
include complete examples so that the reader is not asked to make mental
substitutions.

Add "count" to any resource that previously had only a "type".  By canon,
resource SHALL contain both a type and a count.

Fixes #83
Fixes #86

### DIFF
--- a/spec_14.adoc
+++ b/spec_14.adoc
@@ -282,22 +282,35 @@ task slot for a total of 4 tasks.
 version: 1
 resources:
   - type: node
-    count: 4
+    count:
+      min: 4
+      max: 4
+      operator: "+"
+      operand: 1
     with:
       - type: slot
-        count: 1
+        count:
+          min: 1
+          max: 1
+          operator: "+"
+          operand: 1
         label: default
         with:
           - type: core
-            count: 2
+            count:
+              min: 2
+              max: 2
+              operator: "+"
+              operand: 1
 tasks:
   - command: app
     slot:
       label: default
     count:
-        per_slot: 1
+      per_slot: 1
 walltime:
   duration: 1h
+attrs:
 ----
 
 A simpler example using implicit *task slot* definition to run 4 tasks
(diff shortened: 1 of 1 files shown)


morrone:
Filled in a couple more spots that I missed in the first pass.  Let me know if you see anything else that I missed.

grondo:
Nice comprehensive changes. 

I noticed in some places the key `attr` is used where the spec calls out `attrs`.

Use Case 2.3 seems to have a misplaced tick somewhere which disrupts rendering of 2.4.

Otherwise seems good thanks!

morrone:
Thanks @grondo, I just pushed an update that fixes the attrs misspelling and the case 2.3 rendering problem.  It also fixes another "count: 1" that I missed.

grondo:
Great! Should these changes be merged first, then rebase the footnote PR?

morrone:
That is fine with me.

PR #88, opened 2017-04-03
rfc16:  Add Flux Resource Scheduling Methods RFC

lipari:
Building off RFCs 4, 8 and 14, the Flux Resource Scheduling methods
describe the methods used to schedule resources within the Flux
framework.

### DIFF
--- a/spec_16.adoc
+++ b/spec_16.adoc
@@ -1,0 +1,244 @@
+ifdef::env-github[:outfilesuffix: .adoc]
+
+16/Flux Resource Scheduling Methods
+===================================
+
+The Flux Resource Scheduling methods are the methods used to schedule
+resources within the Flux framework.
+
+* Name: github.com/flux-framework/rfc/spec_16.adoc
+* Editor: Don Lipari
+* State: draft
+
+== Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to
+be interpreted as described in http://tools.ietf.org/html/rfc2119[RFC 2119].
+
+== Related Standards
+
+* link:spec_4{outfilesuffix}[4/Flux Resource Model]
+* link:spec_8{outfilesuffix}[8/Flux Task and Program Execution Services]
+* link:spec_14{outfilesuffix}[14/Canonical Job Specification]
+
+== Goals
+
+RFC 4 presents the Flux Resource Model: a versatile and extensible
+system for representing a diverse collection of computing resources
+along with their complex associations.  One of the primary objectives
+for scheduling Flux resources is to enable the search, selection, and
+allocation of these resources in a simple and efficient manor.
+
+Another important objective is to enable and support recursive or
+hierarchical scheduling models.  These models include the patterns
+defined or prescribed by the user as well as the scheduling behavior
+imposed by the system to enforce policy and achieve resource
+utilization goals.
+
+Using one common scheduling framework, users or system administrators
+should be able to equally and easily create and utilize a hierarchy of
+scheduler instances with distinct scheduling policies tailored to the
+application workload and system constraints.
+
+The goal of this RFC is to describe the methods developers can use to
+identify and allocate Flux resources to Flux programs in a way that is
+flexible enough to support all scheduling policies.  The specific
+goals of these methods are to:
+
+* Load Flux resources into the context of a scheduler instance
+* Create records of resources allocated to Flux programs
+* Remove resource allocation records from Flux resources
+* Save changes in allocation information to Flux resources
+* Display the allocation status of any Flux resource
+
+These goals SHALL apply regardless of whether the resources are
+managed by a dedicated resource service or through a library call.
+
+== Background
+
+Flux programs described in RFC 8 run on or consume Flux computing
+resources described in RFC 4.  In the context of this RFC, a scheduler
+is a Flux framework service that runs within a Flux instance.
+
+A scheduler is responsible for scheduling the computing resources used
+by Flux programs.  In addidition, a scheduler MAY schedule ancillary
+resources like power and file system bandwidth needed to deliver
+computing services.
+
+This RFC defines the Flux Resource Scheduling methods for operating on
+Flux resources in support of the scheduling activities.  In the
+description to follow, the term "job" will be used as defined in RFC 4
+as a synonym for Flux program.
+
+A scheduler receives requests for resources, identifies resources that
+meet the request, and allocates resources to the requesting entity.
+An _allocation_ is a record of a resource assignment to a requesting
+entity.
+
+The methods described in this RFC are the methods that act on those
+resources by a scheduler.  A scheduler reads the resource pool data as
+well as the graph data.  When a set of resources are selected that
+satisfy a resource request, the scheduler modifies the extended data
+of the selected resources to record the allocation.  When the
+allocation expires or is rescinded, the record is removed from the
+allocated resources.
+
+How a scheduler searches the resources and selects resources to assign
+to the requesting entity is outside the scope of this RFC.  This RFC
+presents the methods a scheduler SHALL use to operate on Flux
+resources in support of Flux programs.
+
+== Resource Allocation Considerations
+
+A scheduler's core activity is to find resources that meet a resource
+request and allocate those resources to a Flux program for its use.
+The scheduler records such allocations and thereby removes them from
+consideration as candidates for future requests.
+
+There are two orthogonal characteristics of a resource, its size and
+its composite nature as defined in link:spec_4{outfilesuffix}[4/Flux
+Resource Model].
+
+It is straightforward to represent the allocation of a resource of
+size 1.  We can record the allocation of such a resource to a Flux
+program by adding to the resource's extended data a simple tag that
+represents the ID of the Flux program.
+
+For resources with a size value greater than 1, we have the option of
+allocating `size` units to one Flux program and remaining `size` units
+to a different Flux program.  Such allocations can be recorded as an
+array of Flux program IDs and the number of units allocated.
+
+The Flux Composite Resource Pool model allows for linking resources
+together in a "has-a" relationship.  When a child resource is
+allocated to a Flux program, what should be recorded for the parent
+resource and each parent thereof?
+
+Similarly, shall we allow a resource of size 1 to be allocated to more
+than one Flux program?
+
+== Resource Annotation Considerations
+
+The above resource allocation considerations address the allocation of
+resource size unit(s) to a job.  RFC 4 proposes the use of graphs to
+define the relationships of one resource to another.  Composite
+"has-a" relationships are represented with graphs.
+
+Under this plan, the question arises of how to reflect the allocation
+of a resource to a job to all of the parent resources to which that
+resource belongs.  For example, if a single core is allocated to a
+job, how should that allocation be recorded in the node that contains
+that core and even the rack that contains that node?
+
+For this, an additional system of allocation recording is proposed:
+_annotation_.  Annotations differ from allocations in that annotations
+have no associated size units.  Annotations are simply tags.  In the
+example above, when the core is allocated to a job, the node and rack
+parents MAY be annotated with the job id to which the core is
+allocated.
+
+Annotation is not a hard requirement.  It is simply a bookkeeping
+solution to facilitate the search for available resources.  If a
+resource request requires the exclusive allocation of all the
+resources within a rack for example, the absence of job annotations on
+a rack resource will readily indicate that the rack resource is a
+candidate for exclusive use.
+
+The rules below prescribe the recommended resolution to the above
+considerations.
+
+== Resource Allocation Data and Rules
+
+* One `unit` of the `size` resource pool datum SHALL be the most
+  granular allocatable resource.  Multiple jobs MAY be allocated
+  `size` units of a resource, but no two jobs SHALL be allocated the
+  same resource unit.  In other words, the total number of resource
+  units allocated to jobs MAY not exceed the `size` count defined for
+  the resource.
+
+* The number of `size` units of a Flux resource allocated to a job
+  SHALL be recorded in the resource pool extended data.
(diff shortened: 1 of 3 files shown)


spec_16.adoc
```diff
@@ -0,0 +1,188 @@
...
+questions.
+
+== Resource Allocation Data and Rules
+
+* One `unit` of the `size` resource pool datum SHALL be the most
+  granular allocatable resource.  Multiple jobs may be allocated
+  `size` units of a resource, but no two jobs MAY be allocated the
+  same resource unit.  In other words, the total number of resource
+  units allocated to jobs MAY not exceed the `size` count defined for
+  the resource.
```
morrone: What about our batch clusters where applications from different users share nodes?

Say we have a 10 nodes cluster of nodes with 8 cores.

User A asks for "10 nodes with 4 cores each".  User B asks for the same thing.

It is obvious to us that there are enough nodes and cores available to satisfy both allocations at the same time.  But flux won't know that unless we explicitly tell it, in a configuration file, that a "node" resource is shareable.

spec_16.adoc
```diff
@@ -0,0 +1,188 @@
...
+than one Flux program?
+
+The rules below prescribe the recommended resolution to both
+questions.
+
+== Resource Allocation Data and Rules
+
+* One `unit` of the `size` resource pool datum SHALL be the most
+  granular allocatable resource.  Multiple jobs may be allocated
+  `size` units of a resource, but no two jobs MAY be allocated the
```
morrone: I think this "MAY" should be a "SHALL".  Or if that was not the intent, then it should be a lower-case "may".  The same argument applies to the very next "MAY".

spec_16.adoc
```diff
@@ -0,0 +1,188 @@
...
+  job.
+
+* A resource's requirement to be allocated exclusively to a job MAY be
+  recorded in the resource pool extended data.  In the absence of an
+  exclusive allocation requirement, a scheduler MAY allocate the
+  resource to multiple jobs.
+
+* Allocations change over the course of time.  The allocation data MAY
+  become an array of job allocations over time to support scheduling
+  jobs in the future.
```
morrone: Is it worth the complexity of dealing with a change from a scalar to an array?  Perhaps it should simply be an array always.

spec_16.adoc
```diff
@@ -0,0 +1,188 @@
...
+
+* One `unit` of the `size` resource pool datum SHALL be the most
+  granular allocatable resource.  Multiple jobs may be allocated
+  `size` units of a resource, but no two jobs MAY be allocated the
+  same resource unit.  In other words, the total number of resource
+  units allocated to jobs MAY not exceed the `size` count defined for
+  the resource.
+
+* The number of `size` units of a Flux resource allocated to a job
+  SHALL be recorded in the resource pool extended data.
```
morrone: I keep seeing this term "extended data".  Can you explain what that means?  Is there "normal data" somewhere and "extended data" somewhere else?

spec_16.adoc
```diff
@@ -0,0 +1,188 @@
...
+
+* When all `size` elements of a resource pool have been allocated to a
+  job, the resource SHALL be considered exclusively allocated to the
+  job.
+
+* A resource's requirement to be allocated exclusively to a job MAY be
```
morrone: s/MAY/SHALL

spec_16.adoc
```diff
@@ -0,0 +1,188 @@
...
+  recorded in the resource pool extended data.  In the absence of an
+  exclusive allocation requirement, a scheduler MAY allocate the
+  resource to multiple jobs.
```
morrone: I would think that the opposite position, exclusive allocation, would be the default position.  At least here at LLNL that seems to be the more common situation.  Or perhaps an exclusive/non-exclusive flag should just be a required element for every resource.

spec_16.adoc
```diff
@@ -0,0 +1,188 @@
...
+
+* Allocations change over the course of time.  The allocation data MAY
+  become an array of job allocations over time to support scheduling
+  jobs in the future.
+
+* When a composite resource is allocated exclusively to a job, all of
+  its child resources MUST implicitly be allocated to the job.  To
+  save processing effort, resource allocation records NEED NOT be
+  created for child resources when the parent resource is allocated
```
morrone: I'm not sure that I am comfortable with is being optional.  I think it makes sense to pick a direction and commit to it.

spec_16.adoc
lipari: Two points... We state below that any resource has the option to specify that it be allocated exclusively to a job.  Hence, we could set the "exclusive" flag on a node if we wanted it to be allocated to at most one job.  And the node's one size unit would be allocated to the job.
For those nodes that do not have this flag set, the 1 size unit of each of four cores would be allocated to one job, the 1 size unit of each of four other cores would be allocated to the second job.  There would be an "annotation" at the node level that some of its children have been allocated to the jobs 1 and 2.
Perhaps this bears better definition in this RFC, but I deliberately left this scenario out, considering it an implementation detail.

spec_16.adoc
lipari: Agreed.  That should be "Multiple jobs MAY be allocated size units of a resource, but no two jobs SHALL be allocated the same resource unit.

spec_16.adoc
lipari: For the existing, now-based FCFS scheduler, a resource is either allocated or free.  An array for that implementation would always be a size of 1.

spec_16.adoc
lipari: Extended data refers to the definition established in RFC4.  The "the OPTIONAL extended data" as described there refers to transient data that changes over time.  "Normal" data is the resource pool's data that identify and characterize the permanent aspects of the resource.

spec_16.adoc
lipari: Agreed.

spec_16.adoc
lipari: You're correct that LC has more node-scheduled clusters than it does core-scheduled.  But I thought of a rack resource where sharing would be the default.  I'm open to making the exclusive flag a requirement of the extended data.

spec_16.adoc
lipari: I'm open to that too.  I just didn't want to mandate an economy measure that wasn't technically necessary.

spec_16.adoc
morrone: Right, so an array would cover both current and near-future (S4) needs without needed to add extra code to detect and swap between types.

lipari:
I have pushed an update to this PR addressing all the comments so far.  As we have done in the past, I suggest we start with the baseline changes and those of you with deeper thoughts on what should be added can submit follow-on PRs - to be independently discussed.

dongahn:
Hi @lipari: this is a great start! 

Can we add a high-level context to the Goals section, describing more explicitly why we create this RFC and who should be the main audience?

I think something like the following may do?

> One of the core core objective of Flux’s scheduling model is to enable effective hierarchical scheduling schemes both at the application and system levels: i.e., using one common scheduling framework, users or system administrators should be able to equally and easily create and utilize a hierarchy of scheduler instances with distinct scheduling policies tailored to the application workload and system constraints. Therefore, it is important for Flux to facilitate creation of a wide range of scheduling policies that can be used by distinct scheduler instances. 

> The goal of this RFC is to describe the ways in which developers can develop and plug in their resource selection scheduling policy logic into our scheduling framework, together forming a scheduler instance. These mechanisms are termed the Flux Resource Scheduling methods. The operations provided by these methods include, but are not limited to:

I will comment more later. But if this is the main purpose of this doc, I think we may want to add a bit more details on how one can do this...

lipari:
Amended commit just pushed that incorporates @dongahn's latest suggestions.

spec_16.adoc
```diff
@@ -0,0 +1,238 @@
...
+
+Flux programs described in RFC 8 run on or consume Flux computing
+resources described in RFC 4.  A scheduler is responsible for
+scheduling the computing resources used by Flux programs.  This RFC
+defines the Flux Resource Scheduling methods for operating on Flux
+resources in support of the scheduling activities.  In the description
+to follow, the term "job" will be used as defined in RFC 4 as a
+synonym for Flux program.
+
+A scheduler manages computing resources.  A scheduler receives
```
dongahn: 1. Maybe we can generalize the first sentence and state "A scheduler manages computing resources and other consumable resources (e.g., power, file system bandwidth, etc.)"?

2. Since the main readers of this RFC would be scheduler plugin writers, can we, very early on, define/refine what a scheduler instance consist of? Like, a scheduler instance consists of a scheduling framework to provide common scheduling services and policy plug-ins to effect certain policies using these services? The methods described in this RFC are the common abstractions that a policy plug-in SHOULD use to effect its resource selection policy: finding, selecting...

spec_16.adoc
```diff
@@ -0,0 +1,238 @@
...
+requests for resources, identifies resources that meet the request,
+and allocates resources to the requesting entity.  An _allocation_ is
+a record of a resource assignment to a requesting entity.
+
+The methods described in this RFC are the methods that act on those
```
dongahn: I think that one thing that might be good to have in this section is, how a resource selection plug-in should use these methods to make an allocation. Is it a possibility to add some prose to the effect that... Maybe, this means we need to expand this to include the tree graph traversal aspect of this.

spec_16.adoc
```diff
@@ -0,0 +1,238 @@
...
+resources by a scheduler.  A scheduler reads the resource pool data as
+well as the graph data.  When a set of resources are selected that
+satisfy a resource request, the scheduler modifies the extended data
+of the selected resources to record the allocation.  When the
+allocation expires or is rescinded, the record is removed from the
+allocated resources.
+
+How a scheduler searches the resources and selects resources to assign
```
dongahn: I am not sure if I agree with this statement. It feels like it is important to  describe how a scheduler plugin can use these methods to find and research the resources that match with the criteria...

Related to this, somewhere in this document, we may want to describe the main resource model targeted by these methods is a "tree graph" or a hierarchy...

spec_16.adoc
```diff
@@ -0,0 +1,238 @@
...
+"has-a" relationships are represented with graphs.
+
+Under this plan, the question arises of how to reflect the allocation
+of a resource to a job to all of the parent resources to which that
+resource belongs.  For example, if a single core is allocated to a
+job, how should that allocation be recorded in the node that contains
+that core and even the rack that contains that node?
+
+For this, an additional system of allocation recording is proposed:
+_annotation_.  Annotations differ from allocations in that annotations
```
dongahn: I have trouble understanding why we need to introduce a new concept here: annotation. Isn't an annotation a variant form of an allocation -- a non-exclusive allocation?

spec_16.adoc
```diff
@@ -0,0 +1,238 @@
...
+should be able to equally and easily create and utilize a hierarchy of
+scheduler instances with distinct scheduling policies tailored to the
+application workload and system constraints.
+
+The goal of this RFC is to describe the methods developers can use to
+identify and allocate Flux resources to Flux programs in a way that is
+flexible enough to support all scheduling policies.  The specific
+goals of these methods are to:
+
+* Load Flux resources into a scheduler's context
```
lipari: Change made.

spec_16.adoc
lipari: I have revised the text to address this suggestion.  I do not want to introduce the sched framework + plugin design concepts because that was the design we created.  Future scheduling module developers can implement a Flux scheduling module from scratch that does not contain a plugin architecture.  So, I'm hoping to keep this RFC as generic as possible.

spec_16.adoc
lipari: I am hoping to avoid making optional design choices a requirement in this RFC.  While RFC 4 specifies the graph model, it is an optional construct.  Simple schedulers can schedule very basic resources like nodes with no connection to graphs.

I suggest we relegate the application of graph theory to select resources to a separate RFC.  This RFC should just focus on the scheduler's need to allocate and release resources.

spec_16.adoc
lipari: As indicated above, I am hoping to draw a line between the need for a scheduler to allocate resources (this RFC) and the logic a scheduler employs to find resources that match a request and subsequently identify the best selection from those candidates.  That will involve complexities that would make this RFC overly onerous.

spec_16.adoc
lipari: This gets back to an old discussion of how and where to mark resources as allocated.  The question dealt with which resources in a composite hierarchy should be "marked" when an allocation was made.  The question was about resources up and down the hierarchy.  For example, if a node was allocated, should that allocation be reflected in the node's children or its parents?  We introduced the notion of "aggregation" at the parent (and grandparent) levels to serve as bookkeeping tricks that would bring efficiency to the search for resources which could be allocated exclusively.

I am looking to canonize the fruits of that discussion in this RFC.  "Annotation" is the embodiment of that aggregation idea as proposed in this RFC.

lipari:
@dongahn, thank you for pointing out topics that were missing, under-stated or generally
muddled.  Just pushed new commit that incorporates our discussion above.

spec_16.adoc
dongahn: Hmmm... 

> Simple schedulers can schedule very basic resources like nodes with no connection to graphs.

How do those schedulers iterate through the nodes, though? I guess my point is to traverse through the resource vertices maybe is so fundamental to the Flux Resource Scheduling?

lipari: Oops, yes, you are right.  The composite relationship is modeled with graphs.  However, I'd still like to relegate a discussion of graph traversal to a separate RFC dealing solely with finding - and maybe selecting - resources that match a request.

If we merge this PR as is, I'll take another pass at addressing "how a scheduler searches the resources and selects resources to assign to the requesting entity".  If it fits naturally into this PR, I'll submit a follow-on PR to this PR.  However, it gets complicated, then I'll probably submit it as its own PR.

dongahn: OK. This works for me. Let's defer that as future work.

Perhaps that's better given that we will accelerate our design work next week when @morrone returns next week. We can use our RFC writing for selection interface/mechanisms as a way to document the design core.

lipari:
Withdrawn.

PR #89, opened 2017-04-05
rfc:  spell check fixes

lipari:
Unsure how these problems crept in, but spellcheck complains unless
they are fixed.

### DIFF
--- a/spec_14.adoc
+++ b/spec_14.adoc
@@ -1060,6 +1060,6 @@ attrs:
 [sect2]
 == References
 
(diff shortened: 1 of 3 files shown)


lipari:
It would be nice if someone had the time to review/merge this.  After that, I suggest @morrone rebase https://github.com/flux-framework/rfc/pull/84 and I will review/merge it - and then I will submit a PR to address the comments about the JDSL reference.

grondo:
Seems reasonable. There wasn't any way to get the TM symbol added to the spellcheck I presume? (It is indeed in the original article title).

Thanks for catching this. I too wonder how we were able to merge with these errors...

lipari:
> There wasn't any way to get the TM symbol added to the spellcheck I presume?

I spent a few minutes trying some ways to accept the TM and then reached the point of diminishing returns.

PR #92, opened 2017-06-06
rfc7: Add guidelines for function names

grondo:
Add guidelines for function names as discussed in flux-framework/flux-core#1083.

The language may still need some work here.

### DIFF
--- a/spec_7.adoc
+++ b/spec_7.adoc
@@ -50,6 +50,26 @@ For example `msg_count` is OK, but `MsgCount` or `MSG_COUNT` do not conform.
 Preprocessor macro names SHOULD NOT include lower case letters. 
 For example `FLUX_FOO_MAGIC` is OK but `flux_foo_magic` and `FluxFooMagic` do not conform.
 
+=== Function Names
+
+Function names should be descriptive and concise, and SHOULD NOT contain
+upper case letters. Functions that perform an action should have a name
+that makes it clear what the function does, e.g. `check_errors()` instead
+of `checkit()`.
+
+For functions or methods that operate on a possibly opaque class type,
+the function name SHOULD start with the name of the class, and if
+exported via the Flux API, the class name SHALL start with `flux_`,
+e.g. `flux_object_do_something()`.  Functions that get or set *properties*
+of an object SHOULD be named with `obj_get_item()` and `obj_set_item()`,
+not `obj_item_get()` or `object_item_set()`.  In some cases, the `get`
+MAY be dropped, e.g. `obj_size()` instead of `obj_get_size()`. Some classes
+have extra *interfaces* that they export via the class methods, and these
+functions MAY be named with the interface name *before* the action, for
+example `flux_aux_get()` and `flux_aux_set()`, which expose an interface
+to the `aux` data of the `flux_t` object, and do not get or set an
+individual `aux` property.
+
 === Typedefs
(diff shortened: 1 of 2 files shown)


garlick:
This looks fine to me.  Anybody else want to comment before we merge?

grondo:
Yeah, we should get at least one more ACK.

lipari:
Looks fine to me.  I'm glad for the new rules because I though the original `object_item_set()` was ambiguous and could be seen as a function that returns a boolean on whether `object_item` was set.

grondo:
> Looks fine to me. I'm glad for the new rules because I though the original object_item_set() was ambiguous and could be seen as a function that returns a boolean on whether object_item was set.

Good observation @lipari, I hadn't thought of that one

spec_7.adoc
```diff
@@ -50,6 +50,26 @@ For example `msg_count` is OK, but `MsgCount` or `MSG_COUNT` do not conform.
...
+upper case letters. Functions that perform an action should have a name
+that makes it clear what the function does, e.g. `check_errors()` instead
+of `checkit()`.
+
+For functions or methods that operate on a possibly opaque class type,
+the function name SHOULD start with the name of the class, and if
+exported via the Flux API, the class name SHALL start with `flux_`,
+e.g. `flux_object_do_something()`.  Functions that get or set *properties*
+of an object SHOULD be named with `obj_get_item()` and `obj_set_item()`,
+not `obj_item_get()` or `object_item_set()`.  In some cases, the `get`
```
morrone: I'm not really a fan of this rule.  If anything, I greatly prefer the forms "obj_item_get()" and "obj_item_set()".  I know the alternate method reads a little more like english, and my preferred form is kind of like Reverse Polish Notation.  But I think it makes a lot of sense to have the related functions all have the same prefix, instead of having shared names that wrap a changing middle.

grondo: I think we kind of agreed that when a class or object exports an extra "interface" of methods or functions, then the interface name should come before the `get` or `set` (or whatever), because, as you suggest, the related functions would then all have the same prefix. (example: flux_aux)

However, for setters and getters of simple object properties, the `get` and `set` are methods on the object, so I think from your own observation it makes more sense to have `flux_object_get_*` and `flux_object_set_*` so that all related functions have the same prefix. This allows me to easily grep the API header file to see which properties are settable and get-able on an interface.

I'm actually wondering if this is all fairly obvious and we don't need to add this language to an RFC, and instead just audit naming on a case by case basis using common sense?

garlick: I'd say it's a bit subtle to be common sense, at least for me.  My brain likes it both ways depending on how I'm thinking about it, and so you're as likely as not to get one way or the other from me.   I vote we put in the rule and then if we need to fix or provide critique in a PR in the future, we don't have to recall the argument, we simply cite the RFC.

grondo: Well, it doesn't sound like we have any consensus yet so I'd be hesitant to merge this.

garlick:
I'm thinking its less important to have consensus and more important to have consistency.  Can you live with the proposed distinction between interfaces and accessors @morrone?

garlick:
Prob should get this rebased and merged.

grondo:
closed since there was never any consensus, sorry for letting it sit.

PR #93, opened 2017-06-15
Jobspec updates

morrone:


morrone:
This is a work in progress.  In particular, I know that the JSON Content Rules definition of the spec needs updating in places to match my changes.  But this should serve as a decent preview.  Any and all comments welcome.

grondo:
For 543dfc7, could you add an explanation to the commit body of why uuid, attributes, and tags are being removed from the resource vertex? All the other commits here have very good explanations in the body.

Probably we already discussed why these keys were being removed, but if so I've forgotten.

spec_14.adoc
```diff
@@ -160,8 +160,20 @@ following keys
  vertex to another resource. Therefore, the value of the `with` key
  SHALL be a dictionary conforming to the resource vertex specification.
 
- *edge*::
- **XXX**: need specification for other "edge match descriptors"
+ *label*::
+ The `label` key SHALL be a string that may be used to reference this
+ resource vertex from other locations in the jobspec. `label` SHALL be
+ local to the namespace of the current jobspec, and each `label` in the
+ current jobspec much be unique.  `label` SHALL be mandatory in resource
```
grondo: typo: `s/much/must/`

grondo:
Thanks for doing the edges examples, that was helpful. I'm still digesting, but I don't have any other comments at this time.

spec_14.adoc
morrone: Thanks

dongahn:
Thanks @morrone. I propose we discuss this at tomorrow's design meeting. 

+1 on expanding on the edges concept. It would be helpful if you can walk us through how to express various (simple to complex) resource requirements using edges. 

It is good that you restrict the graph not to have a parallel edge (https://en.wikipedia.org/wiki/Multiple_edges) because it will be much easier to handle.

But one question is, how easy can we express a case where the edge connecting to two resource vertices are used by two distinct hierarchies. Say, an edge to NIC bandwidth resource vertex is used both by the network bandwidth hierarchy and parallel file system 1 bandwidth hierarchy. 

Since I know you want to grow the resource section of the job spec language to be a resource specification language, I thought this would be worth a discussion.

morrone:
> It is good that you restrict the graph not to have a parallel edge (https://en.wikipedia.org/wiki/Multiple_edges) because it will be much easier to handle.

I was not planning to allow multiple edges, but I'm not certain that I actually stated that in the RFC.  It is a good idea for me to add that.  And also to mention that loops are not permitted if that isn't already in there somewhere.

> But one question is, how easy can we express a case where the edge connecting to two resource vertices are used by two distinct hierarchies.

Pretty easily.  The patches add an example section that shows three different ways to express the same relationship of two parents vertices to a single child vertex.  It shows it using in edges, out edges, and combined with the existing 'with' hierarchical method.

dongahn:
> I was not planning to allow multiple edges, but I'm not certain that I actually stated that in the RFC. It is a good idea for me to add that. And also to mention that loops are not permitted if that isn't already in there somewhere.

I am not sure about the loops in particular from the perspective of growing this to serve as a resource spec. You probably want to allow the site to specify the relationship in both directions where it make sense and let the scheduler choose either or both relationship. 

> Pretty easily. The patches add an example section that shows three different ways to express the same relationship of two parents vertices to a single child vertex. It shows it using in edges, out edges, and combined with the existing 'with' hierarchical method.

Great.

morrone:
In this latest push, I attempted to also fix up the JCR section to match my changes.  There seemed to be existing issues in the JCR section, so I also attempted to fix some of those.

morrone:
@trws, it would be good for you to look over these changes too.

dongahn:
Nice work @morrone! I am okay with your changes.

In my mind, the next jobspec-scheduler co-design would be two things:

- The semantics of the scheduler when the resource section is partially specified w/ respect to the scheduler's resource representation. For example, if the scheduler's node representation in the containment hierarchy is `rack->chassis->node` while the spec specifies `rack->node`, what do we expect the scheduler should do. I think the scheduler could skip the missing vertex level and try to match downward until it can or can't. But this needs to be explained somewhere.

- When you have an graph edge that spans multiple hierarchies in the job spec, how should the scheduler know, toward which hierarchy it should walk for matching. In your example, switch can be contained in the scheduler's IB connection hierarchy and some other hierarchy such as parallel file system hierarchy (-- i'm making this up.). This can be pretty unclear if we in particular allow partial spec like above. If the job spec can also encode the hierarchy information somehow, I think this can disambiguate many things. 

In any case, both of these points can ultimately go into "Resource Matching" section of this RFC. --Future discussion and shouldn't keep us from landing this PR IMO.

morrone:
@dongahn made it clear to me that there is a flaw in my current patches.  I'm not accounting for the multiplicative effect of the "with" directive.  For instance, if there is a vertex of type "node" and count 2 "with" another vertex of type "core" and count 4, that means "two nodes, each with four cores".  There are 8 total cores needed, hence my use of the term "multiplicative".  Although it isn't _just_ multiplicative, because the "cores" in this example are owned, and evenly divided amongst the "nodes"; it is not just a random pool of 8 "cores".

If one vertex has two different parents, how should the multiplicative properties be applied?  In this patch it is left ambiguous.  It is certainly nonsensical to try to apply it to both parent vertices.

In our meeting, we decided that edges need to have a property that determines whether or not it is a multiplicative relationship.  Dong used the terms "associative" (non-multiplicative) and...something else that I'm forgetting at the moment.

Also, we decided that it is probably best to move the 'edges' out of the resource vertex, and have them in their own section.  This aligns with how most tools and libraries that are graph related seem to do it anyway.

So I am going to make changes to move the edges into their own section, and then have some type of designation in the edge to not whether it is associative or multiplicative.

Then I will update the "with" definition to explain that it is not only and "out" edge, but also a multiplicative arrangement.

dongahn:
Thanks @morrone. 

Yes, I used the term "association" without knowing if that is a term commonly used in graph theory. 

But I think the association relationships can be refined to be multiple levels.

For example, one case we discussed was: we want to express `rack[2]->nodes[4]` each of these 4 nodes are connected to a same edge switch. 

But if the user wants to express `rack[2]->nodes[4]`  where all of the 8 nodes are connected to the same edge switch. How should she or he express his? 

In my mind, the first one is association with the current resource vertex without the multiplicative scaling (within its hierarchy) factored in at all. And the other is association with the current resource vertex with *some* multiplicative scaling factored in.

morrone:
We have decided that arbitrary graph resource requests are too complicated for now.  Maybe in the future.  So really we don't need "edge" or "edges" at all for now.

So now this pull request removes "edge", and does a number of other smaller tweaks and cleanup.

Ready to land if it passes review.

dongahn:
Great. If you are okay, I would like at least me and @SteVwonder to give one thorough review pass. Given that he has a poster due Tue, how about COB Wed?

lipari:
I want to go on record as against the migration of duration to be just another "attribute".  I don't buy the argument that the requested duration is just another "attribute".  I could give examples, but it appears I hold the minority opinion.  So, just venting...

morrone:
@dongahn, that is fine with me.

tpatki:
Can you provide an example use case for "minimumDuration"? I suppose you mean a job can request a shorter amount of execution time for a potentially faster turnaround time (wait time is reduced, gets scheduled faster)? 

I think we also need a "maximumDuration" attribute for jobs -- this would be the maximum amount of slowdown that a job can tolerate. Counter-intuitively, specifying this attribute can also reduce turnaround time.  I can think of two use-cases for this. 

(1) The first one is of course power, where the scheduler will look at available power to meet the needs of a job under consideration. If the requested amount of power is available (P_avail >= P_requested), the scheduler will assume that the requested job execution time can be achieved, and will launch the job. If not, it can use a model to predict how much a job will slowdown if it were allocated currently available power (suboptimal allocation), compare it with maximumDuration, and if found suitable, launch the job immediately. 
This scenario occurs more often than one can imagine with flow resources-- for example, if there are 505 W being requested by the job and 500W are available, the job will have to wait. In practice, 5W makes no difference on performance, so the job could have easily been scheduled immediately with the available 500W. Similar issues can arise with I/O and network bandwidth scheduling.

(2) The second use case is for performance reproducibility reasons. Jobs can experience up to ~20% variation with exactly the same input deck/executable/configuration, solely based on where they were placed on the cluster (network contention, topology issues, power inefficient nodes) etc. In case a user assumes a best case execution time for their job while making a request, and then gets unlucky and experiences system variation, their job can get terminated because they ran out of time. Having a maximumDuration attribute will give the ability to a "variation-aware scheduling policy" to ensure that jobs don't get killed because of run-to-run variation.

tpatki:
Another question I have is how the "duration" is determined when a walltime is not specified. For example, in your Use Case 1.1, you assumed that the duration attribute will be 1 hour. 

I'm not sure how we do this at LLNL today with SLURM, hence the question.

tpatki:
From a user point of view, I'm also a bit confused as to how the jobspec will be derived. For example, I am assuming that flags such as "-o" and "-e" (output/error files) will be captured elsewhere and not passed to the scheduler. Where will the corresponding information be stored?

morrone:
{quote}Can you provide an example use case for "minimumDuration"? I suppose you mean a job can request a shorter amount of execution time for a potentially faster turnaround time (wait time is reduced, gets scheduled faster)?

I think we also need a "maximumDuration" attribute for jobs{quote}

Yes, that is what I meant.  If a minimum is supplied, then "duration" is effectively already the maximum.   The combination of minimumDuration and "duration" allows a range to be specified.  If only "duration" is supplied then there is only a single requested value, not a range.  If "minimumDuration" is controversial as a name, I'm fine with removing it until such time as we actually have use for such a thing.

{quote}Another question I have is how the "duration" is determined when a walltime is not specified. For example, in your Use Case 1.1, you assumed that the duration attribute will be 1 hour.{quote}

I assume that the tools will apply some default, as does.  But that is outside of the scope for the jobspec RFC.  I did not introduce the time in that example, I merely converted into the new proposed format.

{quote}For example, I am assuming that flags such as "-o" and "-e" (output/error files) will be captured elsewhere and not passed to the scheduler.{quote}

The jobspec is not only used by the scheduler.  Fir instance, it will also be used by the job launch system which is why it contains a "tasks" section.  I have every expectation that additional directives will be used by the system and added to attributes->system over time.  We don't need to know all of them at this point in time.  We can add them as we find need for them.  Not all flux modules of the same type (flux schedulers, for instance) might necessarily even support all of the same values in attributes->system.  Having a generic key/value area in the specification allows for easy expansion over time and for different components.

spec_14.adoc
```diff
@@ -237,37 +223,60 @@ which to map the defined tasks.
...
+ and recognize different sets of attributes.
+
+ Most system attributes are optional.  Flux modules SHALL provide
+ reasonable defaults for any system attributes that they recognize when
+ at all possible.
+
+ Some common system attributes are:
+
+ *duration*::
+ The value of the `duration` attribute is a string representing time.
```
dongahn: I have a slight preference to time span as opposed to time.

spec_14.adoc
```diff
@@ -237,37 +223,60 @@ which to map the defined tasks.
...
+ The scheduler will make a best effort to allocate the requested resources
+ for the time specified in `duration`, but it is not an absolute guarantee.
```
dongahn: Hmmm. Assuming that this is "walltime" requirement, why no guarantee? Or does this cover something else like "start time" ?

spec_14.adoc
```diff
@@ -237,37 +223,60 @@ which to map the defined tasks.
...
+ *minimumDuration*::
+ The value of the `minimumDuration` attribute is a string resenting a time
+ that is smaller than, or equal to, `duration`.  The scheduler will make
+ a best effort to allocate the time requested in `duration`, but a job that
+ can tolerate using a smaller amount of time may express the minimum duration
+ that will allow the application to make progress.  The scheduler may then
+ be able to run the job using a smaller duration value much sooner than if
+ it had to honor the full `duration` request.
+
+ *truncatable*::
```
dongahn: Is truncatable job an well-established term? Why not, preemptable, instead?

dongahn:
>> Another question I have is how the "duration" is determined when a walltime is not specified. For example, in your Use Case 1.1, you assumed that the duration attribute will be 1 hour.

> I assume that the tools will apply some default, as does SLURM. But that is outside of the scope for the jobspec RFC. I did not introduce the time in that example, I merely converted into the new proposed format.

This would also go back to @lipari's concern above. 

It feels like how long you want to use the resources (subject to other constraints) for your job is so fundamental to the job, the duration fields need to be made mandatory. As is, the hard requirement of a job is only resource and task; the duration is optional.

dongahn:
@morrone: Thank you for putting this together. I have no other comments. Along the way, I've come across some other areas of improvements for this RFC, but they have nothing to do with your changes. I will try to address them as part of a future PR as we discussed.

spec_14.adoc
morrone: The extend of guarantee will depend on each individual scheduler implementation, and on a number of external factors (e.g.sysdmin decides to kill jobs early).  It is probably just best not to set users' expectations too high. :)  But I'm open to other wording.

spec_14.adoc
morrone: In computing, I think "preemption" generally is used to mean a temporary interruption with later resumption.  This would be more like our standby jobs where they are unceremoniously terminated with no later resumption.  I'm not in love with the term, but I was hunting for something clearer than "preemptable".  As always, I'm open to other options, or even just dropping this for now if we don't need this option right now.

lipari: In the precedent from Slurm, "suspension" is the term that refers to the temporary interruption.  "Preemption" is a more general term that includes suspension, checkpointing, and termination.

morrone:
> It feels like how long you want to use the resources (subject to other constraints) for your job is so fundamental to the job, the duration fields need to be made mandatory. As is, the hard requirement of a job is only resource and task; the duration is optional.

Duration is optional because that matches current standard practice with SLURM and other tools.  I don't know of any off the top of my head that require a user to specify a time span.  I see no issue at all with the time span being optional.  When the scheduler sees a jobspec without a specified duration, it can choose any value it likes.  The value it chooses will likely be a system configurable default. I think our systems may default to 12 hours (and maybe have different defaults for different partitions and/or clusters).

morrone:
@dongahn, you don't necessarily need to wait for a PR to explain further issues.  You can open issues right away to track work that is still needed.

dongahn:
> Duration is optional because that matches current standard practice with SLURM and other tools. I don't know of any off the top of my head that require a user to specify a time span. I see no issue at all with the time span being optional. When the scheduler sees a jobspec without a specified duration, it can choose any value it likes. The value it chooses will likely be a system configurable default. I think our systems may default to 12 hours (and maybe have different defaults for different partitions and/or clusters).

I see. So the duration is an essetial part of the job. It just that there will be a default value determined somewhere else..

I will be happy if we have a statement to the effect that the duration will be determined by the system (e.g., the scheduler, system conf...) when not specified. (implementation specific.)  Personally I like the spec that explains things as self contained as possible...

dongahn:
> @dongahn, you don't necessarily need to wait for a PR to explain further issues. You can open issues right away to track work that is still needed.

Got it.

tpatki:
> 
> > 
> Yes, that is what I meant. If a minimum is supplied, then "duration" is effectively already the maximum. The combination of minimumDuration and "duration" allows a range to be specified. If only "duration" is supplied then there is only a single requested value, not a range. If "minimumDuration" is controversial as a name, I'm fine with removing it until such time as we actually have use for such a thing.

Ah I see. After thinking about it a bit, I suppose it is safe to assume that no job ever will have a "minimumDuration", "duration" and "maximumDuration" all specified -- I couldn't come up with a use case for that. 

Specifying a range, however, seems important and I can see use in the near term with power/flow resources at least. My suggestion is to use "duration" and "maximumDuration" to identify the range for simplicity. It is harder to reason in terms of "minimumDuration".

lipari:
I don't see the need to change the term from (the conventional) "time limit" to "duration".  "Duration" does not imply the fact that there is a penalty for exceeding it.  If the job runs beyond the time limit, it will be killed or preempted in some way.  Also, what about including a minimum time requirement?  E.g., do not start job before 8am.  In addition, it would be nice to stipulate a time deadline.  E.g., do not run job if it cannot complete by 5pm.

I see the RFC as the place to define the specification.  It seems to wishy-washy to me to state the you "may" include any terms that you want.  That's great for flexibility and in general, I support that ability.  But what about stipulating conventions?  E.g., your job could include a request for a "runtime" parameter.  But if you want it to honored by most schedulers, you'd better call it "timelimit".

tpatki:
+1 on Don's comment.

I think we are getting toward 4 things here:
1) Users specifying a "minimum start-time" (don't start my job before 8am)
2) Users specifying a "maximum end-time or deadline" (don't bother scheduling my job if you can't finish it by 5p)
3) Users specifying a "execution time" (expected to finish in this time once the job starts running)
4) Users specifying a "time limit" (kill my job if I take more time than this)

A combination of (3) & (4) can be used to specify a range of slowdown for power (what we had done with SLURM as well), as well as to address the performance reproducibility issue. Note that the system will set (3)=(4) if user input doesn't exist.

SteVwonder:
+1 to the idea of duration/walltime/timelimit being an optional attribute of the jobspec. I can imagine a scenario in the future of Flux running on a small community/university cluster that doesn't enforce timelimits.  So requiring timelimits in the spec would be detrimental to those (admittedly hypothetical) future users.

@lipari: I like your idea of mentioning conventions in this RFC.  Another option could be to write a document (doesn't necessarily need to be an RFC) that details what the jobspec for a minimal Slurm replacement would look like.  This would allow us to be more specific (e.g., require timelimits,= abd specify the flags for stdout/err redirection) as well as use stronger language (e.g., SHOULD->SHALL).  +1 to using timelimit rather than duration in the minimal Slurm replacement spec.

lipari:
> So requiring timelimits in the spec would be detrimental to those (admittedly hypothetical) future users.

I'd like to clarify.  As @morrone said, we don't need to require each job request to include a time limit.  However, when doing back-fill scheduling, a time limit (or min-max range) needs to be established for every job by default if it is not present in the job request.  LC systems typically slap on a default time limit based on the cluster/node partition the job has targeted.

spec_14.adoc
```diff
@@ -237,37 +223,60 @@ which to map the defined tasks.
...
+
+ *user*::
+ Attributes in the `user` dictionary are unrestricted, and may be used
+ as the application demands.  Flux may provide addition tools that can
+ identify jobs based on `user` attributes.
+
+ *system*::
+ Attributes in the `system` dictionary are additional parameters to
+ a Flux instance that affect program execution, scheduling, etc.  All
+ attributes in `system` reserved words, however unrecognized
```
SteVwonder: in `system` reserved words -> in `system` are reserved words

spec_14.adoc
```diff
@@ -237,37 +223,60 @@ which to map the defined tasks.
...
+ words SHALL trigger no more than a warning.  This permits jobspec
+ reuse between multiple flux instances which may be configured differently
+ and recognize different sets of attributes.
+
+ Most system attributes are optional.  Flux modules SHALL provide
```
SteVwonder: This is rendering on Github as a code block for some reason.  As is the below sentence "Some common system attributes are:"

SteVwonder:
@morrone: minor nit, you might want to add yourself as an editor at the top of this RFC

morrone:
> I see the RFC as the place to define the specification. It seems to wishy-washy to me to state that you "may" include any terms that you want.

It is not being wishy-washy.  Flux in by nature a flexible framework, not a single way of doing things.  The jobspec needs to allow for flexibility.  Our goal is not to have a single scheduler that works one way, it is to allow a modular infrastructure where people are permitted to select their scheduler of choice (if they use a scheduler at all).  What capabilities a particular scheduler will offer will be implementation dependant.  We may decide that all schedulers need to support some base of system attributes...or we might not.  I think it is too early to say.

Keep in mind this RFC is about documenting the jobspec _format_.  It not about documenting every option that flux might support.

> Also, what about including a minimum time requirement?

I can imagine a _great_ many options that a scheduler might support.  This specification allows adding new options without changing the format of the jobspec.  We may choose to document the most common ones in the jobspec over time.  But I don't think we should premturely document a whole bunch of wish list options before we actually have.

> My suggestion is to use "duration" and "maximumDuration" to identify the range for simplicity. It is harder to reason in terms of "minimumDuration".

I don't understand why one would be harder than the other.  But there seems to be enough consternation about this option that I'm just going to take it out for now.  I'm also falling into the trap of prematurely documenting options.

> minor nit, you might want to add yourself as an editor at the top of this RFC

Oh, I don't know...I think the git logs serve that purpose just fine.  I'd probably rather remove the editors line all together.  But then, I don't want to offend people that might feel otherwise.

morrone:
> I will be happy if we have a statement to the effect that the duration will be determined by the system (e.g., the scheduler, system conf...) when not specified.

How an individual scheduler will handle an option is ultimately up to that particular scheduler.  One scheduler might decide to reject a jobspec that lacks a duration field.  Another might ignore the duration field entirely.  For instance, FIFO scheduler for an individual application (UQ, ATS, etc.) in a sub-instance (or running under something like SLURM) might not care about duration at all.

I'll also draw attention to this paragraph:

>  Most system attributes are optional.  Flux modules SHALL provide
 reasonable defaults for any system attributes that they recognize when
 at all possible.

spec_14.adoc
```diff
@@ -119,15 +119,15 @@ A resource vertex SHALL contain the following keys:
...
+ language itself. The currently reserved type is `slot`, used to
+ define *task slots*.  Reserved types are described in the
  *Reserved Resource Types* section below.
 
  *count*::
  The `count` key SHALL indicate the desired number or range of
- resources matching the current vertex. The value of `count` SHALL be a
- dictionary which SHALL contain the following keys:
+ resources matching the current vertex. The `count` SHALL have one
+ one of two possible values: either a single integer value representing
```
morrone: Thanks.

spec_14.adoc
morrone: Thanks.

spec_14.adoc
morrone: Thanks.

spec_14.adoc
morrone: OK

spec_14.adoc
morrone: I simplified the wording further.

lipari:
> Keep in mind this RFC is about documenting the jobspec format. It not about documenting every option that flux might support.

I agree.  And in the baseline RFC 14, there is a nice distinction between the hard requirements (resource, task, and wall time dictionaries) and the optional, free form attributes.  By moving items to become all free-form attributes, you might as well not bother specifying anything.  By your logic, it's all good and does not need to be tied down.

dongahn:
>How an individual scheduler will handle an option is ultimately up to that particular scheduler. One scheduler might decide to reject a jobspec that lacks a duration field. Another might ignore the duration field entirely. For instance, FIFO scheduler for an individual application (UQ, ATS, etc.) in a sub-instance (or running under something like SLURM) might not care about duration at all.

Nice response. Can this be somehow incorporated into the high level goal or overview section of this RFC?

> Most system attributes are optional. Flux modules SHALL provide
reasonable defaults for any system attributes that they recognize when
at all possible.

Ah. I paid less attention to this wording as I thought they were leftovers due to the formatting issue. In fact, if you can expand on "Some common system attributes are:" and including `durations`, this will address my concerns.

spec_14.adoc
morrone: Removed.  Problem solved for now. :)

dongahn: Heh... Nice approach. (this made my day)

morrone:
OK, updates applied.

morrone:
OK, updates applied.

dongahn:
> I paid less attention to this wording as I thought they were leftovers due to the formatting issue. In fact, if you can expand on "Some common system attributes are:" and including durations, this will address my concerns.

I realized that this is what you did! It was just the formatting issue then. BTW there is a typo (+ sign) in the latest commit.

spec_14.adoc
```diff
@@ -321,20 +302,16 @@ across 4 nodes
...
       per_slot: 1
-walltime:
-  duration: 1h
-attrs:
+attributes:
+  system:
+    duration: 1 hour
 ----
 
 === Content Rules
```
SteVwonder: @morrone: what are your thoughts on the JSON content rules?  I think it is nice to have a formal grammar for the spec, but the job spec examples are in YAML, not JSON.  So at minimum, the sentence "A jobspec conforming to version 1 of the language definition SHALL adhere to the following ruleset" isn't true

morrone:
> BTW there is a typo (+ sign) in the latest commit.

It isn't a typo, it is an attempt at a list continuation to address the formatting problem that Stephen pointed out.  So either I did it wrong, or github isn't parsing asciidoc correctly again.

morrone:
The formatting might be fixed now...

morrone:
Yay! That worked.

spec_14.adoc
morrone: A case can be made that the JCR works because YAML is a superset of JSON now, and everything expressed applies to YAML too.  It is just that one can also use the equivalent non-braces form in the YAML as well.  I don't have a strong opinion for or against at the moment.  On the one hand it is nice to have a more formal grammar, on the other hand it is a little strange to use JCR for YAML.

dongahn:
@morrone: ready to merge so that you can move onto your next step?

morrone:
Sure, merging is fine with me

dongahn:
OK, thanks. I will wait until COB today in case there are other reviewers who have more asks. If all good, merge.

PR #94, opened 2017-07-28
rfc7: Ban the use of "_struct" in structure tags.

morrone:
Ban the use of "_struct" in structure tags.


garlick:
I had a bad habit of doing that for "throw away" names where the structure tag was only used to define a typedef.  I will change my ways.

PR #95, opened 2017-08-01
represent KVS values internally as opaque data, and cleanup

garlick:
_Note: This spec was proposed a while back but never implemented.  We're now getting around to that, and I'm working with @chu11 to refine the spec and work out implementation details.  Once we're further along, I'll write some descriptive text to document this better._

This PR proposes that KVS values be internally represented as unstructured, opaque data, rather than requiring them to be JSON.  The previous specification was kind of trying to have it both ways with the _value_ object that was JSON and a _valref_ object that referred to opaque data in the content store.

Note that API's won't have to change - the main API's can remain JSON oriented, but internally the KVS will treat values as opaque.  We'll add functions for accessing the raw data too so users won't have to layer on base64 encoding when they want to do that.  The `flux-kvs` command will likely require some fine tuning where it expects values to be JSON.

The rest of this PR is cleanup:
* renamed _treeref_ to _dirref_, and constrained its use to only refer to directories
* renamed _value_ to _val_  so now we have _valref_ and _val_, _dirref_ and _dir_; and changed _val_ to always represent value as a base64 string as discussed above
* renamed JSON object key from `val` to `data` to avoid confusion with the _val_ object

### DIFF
--- a/spec_11.adoc
+++ b/spec_11.adoc
@@ -26,7 +26,7 @@ be interpreted as described in http://tools.ietf.org/html/rfc2119[RFC 2119].
 == Goals
 
 * Facilitate services implementing private KVS namespaces.
-* Users may directly walk a KVS namespace, starting with a blobref/treeref.
+* Users may directly walk a KVS namespace, starting with a dirref.
 * Tree objects can be exchanged between Flux instances.
 * Tree objects can be parsed years after they were written (provenance).
 * Values exceeding the blob size limit (RFC 10) can be represented.
@@ -38,6 +38,9 @@ be interpreted as described in http://tools.ietf.org/html/rfc2119[RFC 2119].
 
 == Implementation
 
+In contrast to the original KVS prototype, in this specification, KVS
+values are unstructured, opaque data.
+
 All tree objects SHALL be represented as JSON objects containing three top
 level names:
 
@@ -47,14 +50,19 @@ level names:
 
 There are six types:
 
-* _valref_, data is an array of blobrefs comprising opaque data
-* _val_, data is a base64 string representing opaque data
-* _dirref_, data is an array of blobrefs comprising a _dir_ or _hdir_ object
-* _dir_, data is a map of names to tree objects
-* _hdir_, data is a map of integer-hashed names to _dirref_ objects
-* _symlink_, data is a hierarchical key name
+* _valref_, data is an array of blobrefs.  The concatenated blobs are
+an opaque data value (not a _val_ object).
+* _val_, data is a base64 string representing opaque data.
+* _dirref_, data is an array of blobrefs.  The concatenated blobs are
+interpreted as a _dir_ or _hdir_ object.
+* _dir_, data is a map of names to tree objects.
+* _hdir_, data is a map of integer-hashed names to _dirref_ objects.
+* _symlink_, data is a hierarchical key name.
 
 === Valref ===
+
+A _valref_ refers to opaque data in the content store (the actual data,
+not a _val_ object).
 
 ----
 { "ver":1,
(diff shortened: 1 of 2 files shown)


grondo:
Sorry --  for the uninitiated, what is `"NDIK"`?

Also very minor typo in the first commit: need a second single quote on `'data`

garlick:
Sorry, random base64 gibberish. ("42" converted to base64).  Thanks for the typo catch!

grondo:
> Sorry, random base64 gibberish. ("42" converted to base64). Thanks for the typo catch!

Oh, heh, sorry, possibly the one base64 encoding everyone should know by sight

garlick:
Just pushed fixes, plus the spellcheck failures.  I changed the base64 example to `NDIyCg==` ("422") to be a little more obvious about what it is.

chu11:
@garlick Definitely like the changes, I think clears up some of the confusion.

For clarification sake, should it be made clear that the dirrefs also point to opaque data?  Perhaps its implied, but maybe good to spell out.

garlick:
> For clarification sake, should it be made clear that the dirrefs also point to opaque data? Perhaps its implied, but maybe good to spell out.

Dirrefs point to dirs, i.e. when you concatenate the blobs, the data can be interpreted as a dir object.

I was intending that Valrefs point directly to data, not to a val object.  Bleah, poor choice of naming!  Let me rethink that one.  Sheesh why did I do that?

Is that the source of confusion or is there something else?

grondo:
I think maybe the use of "comprises" in the document should be avoided. First, it tends to be confusing, and I'm probably wrong but I think that "comprises" means something like "is made up of" and I'm not sure here you mean "the blobs are made up of opaque data" (though that is technically correct, I'm *assuming* that what is meant is that the blobs are concatenated to make up a single unit(?) of opaque data.?

Sorry if I'm incorrect.

A similar case for *hdir*, I think possibly what is meant is that concatenated blobs *compose* an *hdir* or *dir* object -- though my preference would just be to say the blobs *make up* an `hdir` or `dir` object to avoid the whole compose v comprise debate.

garlick:
Thanks @grondo, you're right.  I just pushed an update that hopefully makes it a bit clearer.  I just avoided both compose and comprise and used simpler language.

Rather than change val and valref, I just noted what the behavior is for now.  We'll see how @chu11 comes out on that one.

grondo:
Thanks! New wording is much clearer to me!

chu11:
@garlick TBH, now I don't quite remember what my mind thought when I wrote my first comment.  I was clearly mixing something up.  But I think the current wording is much clearer.

Agree that ```valref``` should probably be renamed as the need for the text "(not a val object)" probably indicates that it should be changed.  Struggling to suggest a good name.  I thought ```arrayref```, but that may be confusing as the blobs are supposed to be concatenated into a single data value.  Perhaps ```objectref```?? (I don't love it, but I guess its not terrible.)

garlick:
I was kind of thinking we were OK with the current wording and caveats, but I'll ponder.  "objectref" would seem to also imply that the target should be interpreted as a tree object, which is what we're trying to avoid, so...

chu11:
@garlick I think it's something to ponder, but agree that it's not worth holding up moving forward.

garlick:
Sounds good.  This spec is only just skeletal at the moment so we'll be back.

chu11:
did you want to tweak this PR a bit more?  Or do you think this round of edits is merge-able?

grondo:
Some commits could be squashed though the certainly don't need to be...

garlick:
I'll just tidy up/squash a bit, hang on.

garlick:
Ready.

grondo:
Since @chu11 already approved, merging.

PR #98, opened 2017-09-01
rfc11: add optionall content-encoding field

garlick:
Add an optional `content-encoding` field to _val_ and _valref_ objects.

The main objective is to fix the awkward necessity to base64 all values to represent them in a _val_ object.  If they are JSON, they can be represented directly.  This provides a means to indicate that a value is JSON so it need not be put in a base64 envelope.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -85,6 +85,12 @@ submitted to a Flux instance for execution.  This RFC describes the
 canonical form of the jobspec language, which represents a request to
 run exactly one program.
 
+link:spec_15{outfilesuffix}[15/Independent Minister of Privilege for Flux: The Security IMP]::
+This specification describes Flux Security IMP, a privileged service
+used by multi-user Flux instances to launch, monitor, and control
+processes running as users other than the instance owner.
+
+
 == Change Process
 
 The change process is
--- a/spec_11.adoc
+++ b/spec_11.adoc
@@ -38,51 +38,94 @@ be interpreted as described in http://tools.ietf.org/html/rfc2119[RFC 2119].
 
 == Implementation
 
-In contrast to the original KVS prototype, in this specification, KVS
-values are unstructured, opaque data.
-
 All tree objects SHALL be represented as JSON objects containing three top
 level names:
 
 * _ver_, an integer version number
-* _type_, a string type name
+* _type_, a string object type name
 * _data_, a type dependent JSON object
 
-There are six types:
+In addition, _val_ and _valref_ objects MAY contain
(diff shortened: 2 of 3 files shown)


chu11:
I think this is a good idea.  The one part I'm wondering could be tweaked/clarified is:

```
+If set to _raw_, the value is an opaque sequence of octets. 
+A _val_ object represents the value as a base64 string.
+A _valref_ object refers directly to the value in the content store.
```

When I read this, it says "raw" in a val object points to a base64 string.  But "raw" in a valref object can point to data encoded in any way?  Would it be better for "raw" in a valref object to always point to data stored base64?

Then the question might be asked, should we call it "base64" instead of "raw"?

grondo:
I think this is fine, though it kind of feels like the discussion of content-encoding should come *after* the full definitions of `val` and `valref` since they are discussed before fully defined, but this kind of reorganization could come at a later revision... (I was initially confused when reading the content-encoding rules out of context -- they only make full sense after having read the RFC in full once.)

garlick:
@grondo: Yeah, sorry.  This RFC is definitely not fully formed.  It started as little more than notes (and sat there for a long time that way) and I've been adding to it as we tweak the content.  After the churn is done we definitely need to expand it out and edit for better clarity.  A really big problem is that the objects are "specified" as examples instead of JCR.  Anyway, coming soon.

@chu11: I think I see your point.  base64-ing data that we put in the content store would be as bad if not worse than base64-ing JSON we put in _val_ data.  Am I conflating two different concepts?  Maybe content-type is either json or raw, and content-encoding is either base64 or raw?  Hmm, pondering.

chu11:
@garlick I wasn't thinking that deep.  I think there's a subtlety that I'm now realizing is leading to my confusion.

Right now in the KVS, valrefs are only used in one circumstance.  They are used when a val object is big.  So the data from a val object (which is already a base64 string) gets stored into the content store, and the valref keeps a blobref reference to that data.  When a lookup is done and a valref is found, the data that it points to is a base64 string.  That base64 string is placed into a val object when returned to a user.

So I think I'm assuming that the data a valref points to is always base64 encoded.  But this may not necessarily be something desired or true?  B/c of text like this "If the _content-encoding_ is _raw_ or omitted, the content store contains the value with no encoding", I think I got confused.

Perhaps you might be onto something with your comment "Maybe content-type is either json or raw, and content-encoding is either base64 or raw?"

garlick:
This needs a rethink.  Closing.

PR #99, opened 2017-09-05
misc RFC 15 cleanup

garlick:
I aborted pr #98 , but it contained two patches for RFC 15, one to add proper names to the dictionary as travis was failing for me, and one to make a link/abstract for it in the top level readme.  Here are those patches by themselves.


grondo:
Thanks!

PR #102, opened 2017-09-14
rfc14: Remove implicit slots

morrone:
We have found that if we remove implicit slots from the jobspec
it makes implementation more straight forward in places.  For instance,
the scheduler can ignore the "tasks" section of the jobspec and just
focus on the "resources" section.

There is also an arugment to be made that having only one explicit
way to specify slots better honors our goal to "Emphasize expressivity
over simplicity".

### DIFF
--- a/spec_14.adoc
+++ b/spec_14.adoc
@@ -182,7 +182,8 @@ provided SHALL be local to the namespace of the current jobspec.
 +
 A task slot SHALL have at least one edge specified using `with:`, and
 the resources associated with a slot SHALL be exclusively allocated
-to the program described in the jobspec.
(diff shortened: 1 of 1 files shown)


dongahn:
This looks great! Did you want to adjust the change to Use Case 1.3 and 1.4 as well?

grondo:
Sorry @dongahn missed your comment and merged this already. If the use cases were not fixed already that will have to be done in another PR.

lipari:
Also, w/r/t the change in bb11676791fda75c582d44727c86c63192eeba32, the YAML examples are not displaying on my browser, just the text:  link:data/spec_14/exampleX.yaml[].

PR #103, opened 2017-09-14
Break out examples

morrone:
All of the jobspec yaml examples are broken out into individual
files.  This will make it easier to run each example through an
automated validator in the future.


grondo:
Oh, that was a good idea!

PR #104, opened 2017-09-14
rfc14: fix typo in use case file include

morrone:


dongahn:
I actually meant this in use case 1.3 and 1.4....

'''yaml
slot:
      type: node
'''

morrone:
@dongahn Ah, I see, thanks.  I'll pick that up in another pull request.  I'll might find some other things to fix up too.  I am running all of the examples through the parser, and if it is working it should catch the problem you note.  Found at least one parser update I should make too.

dongahn:
Thanks @morrone.

PR #107, opened 2017-09-22
rfc16: add KVS Job Schema RFC

garlick:
This PR adds a preliminary version (an incomplete rough draft really) of a "KVS Job Schema" RFC.  While I think it's probably a bit early to settle on fine details here, it's helpful now to anticipate how "instance components" (running as instance owner) and "guest components" (running as the the job submitter) of the system might use the KVS, as we approach new KVS features in the near term:

* flux-framework/flux-core#1197 KVS namespaces
* flux-framework/flux-core#1193 KVS atomic append
* flux-framework/flux-core#1185 KVS checkpoint restart

(The namespaces achieve two goals:  one, to decouple performance of the primary KVS namespace from commit activity within a job, and two, to provide coarse grained multi-user support in the KVS.  The atomic append would allow an "event log" or an I/O stream to be appended to by multiple writers without the race condition and poor performance of read-modify-write, or kludge of the "sequenced keys" approach taken in libkz.  The checkpoint restart feature would allow an instance to recover e.g. job state when restarted)

@grondo and I reviewed the [high level design](https://github.com/flux-framework/flux-core/wiki/Multi-user-parallel-job-launch---High-level-design) that captured an earlier discussion in light of feedback from @morrone as he worked on job ingest.   We worked out how the system could allow the job shells and application to control the guest namespace, and the instance components (ingest, scheduler, exec service) would exclusively use the primary namespace.  Any access needed by the guest to the primary KVS namespace would need to be proxied through a service running as the instance owner.  Tools for listing, cancelling, etc jobs would use a "job management service" in a similar manner to access job data.  This keeps the multi-user security model for the KVS simple and the code for it easily audited.

The design is basically the same as originally proposed in the wiki page.  I think the big change is that the "jobns" is now basically scratch space for the job shell and the application rather than trying to contain more of the per-job state.

@morrone had the idea to move each job through a series of directories rather than log state changes to an event log, to make it easier for the components to pick up the right jobs without searching through them all.  That could be added here (symlinks?), and is now compatible with the event log idea since the append to the event log and directory move could be done under the same commit now, without opportunities for races (since both are in the primary namespace)

Comments welcome.

### DIFF
--- a/spec_16.adoc
+++ b/spec_16.adoc
@@ -1,0 +1,226 @@
+ifdef::env-github[:outfilesuffix: .adoc]
+
+16/KVS Job Schema
+=================
+
+This specification describes the format of data stored in the KVS
+for Flux jobs.
+
+* Name: github.com/flux-framework/rfc/spec_16.adoc
+* Editor: Jim Garlick <garlick@llnl.gov>
+* State: raw
+
+== Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to
+be interpreted as described in http://tools.ietf.org/html/rfc2119[RFC 2119].
+
+== Related Standards
+
+*  link:spec_12{outfilesuffix}[12/Flux Security Architecture]
+
+== Background
+
+=== Components that use the KVS job schema
+
+Instance components have direct, read/write access to the primary KVS
+namespace:
+
+* _Ingest agent_
+* _Job management service_ (list, cancel, etc.)
+* _Exec service_
+* _Scheduler_
+
+Guest components have direct, read/write access to a private KVS namespace:
+
+* _Job shell_
+* _User tasks_
+* `flux-run` command
+* `flux-queue` command
+* `flux-cancel` command
+
+=== Job Life Cycle
+
+Jobs are submitted to the _ingest agent_ which performs a sanity
+check on the request and either adds the job to the KVS or
+refuses to accept it.
+
+The _scheduler_ takes up submitted jobs and considers them against
+the available resources.  It either rejects the job outright,
+or annotates it and eventually assigns resources to it.
+
+The _exec service_ takes up jobs that have assigned resources,
+instantiates containers for those jobs and starts job shells
+within them.  When job shells and their containers start to exit,
+the _exec service_ "reaps" these shells and allows the _scheduler_
+to reclaim resources assigned to each container.
+
+Jobs therefore transition between the following states,
+not necessarily in a linear order:
+
+submitted::
+  accepted by ingest, resources not yet assigned
+
+runnable::
+  resources assigned
+
+starting::
+  first container started
+
+running::
+  all containers instantiated and job shells started
+
+exiting::
+  first container exited
+
+completed::
+  all containers exited
+
+rejected::
+  scheduler won't run job
+
+== Implementation
+
+=== Primary KVS Namespace
+
+The Flux instance has a default, shared namespace that is accessible
+only by the instance owner.
+
+All job data is stored under a `jobs` directory in the primary namespace.
+
+
+=== Active vs Inactive Job Directories
+
+Each active job has a directory under `jobs.active.<jobid>`,
+where `<jobid>` is a unique sequence number assigned by the
+_ingest agent_.
+
+After a job is no longer active, its directory is moved from
+`jobs.active` to `jobs.inactive`.  Inactive means its state
+is read-only.  The purpose of moving inactive jobs to another
+directory is to keep the size of the active directory manageable.
+The `jobs.inactive` directory may need to be periodically archived
+and purged to keep its size manageable in long-running instances.
+
+
+=== Guest KVS Namespace
+
+A guest-writable KVS namespace is created for the use of the
+job shell and the application.  While the job is active,
+this namespace is mounted in the primary namespace, on
+`jobs.active.<jobid>.guest`.  While mounted, it can be changed
+by the guest components without impacting performance of the primary
+namespace, while still being accessible through the primary namespace.
+
+When the job is inactive, the final snapshot of the guest namespace
+is linked into the primary namespace, and the guest namespace is
+destroyed.
+
+
+=== Event Log
+
+Active jobs undergo state transitions that are recorded under
+the key `jobs.active.<jobid>.eventlog`.  A KVS append operation
+is used to add events to this log.
+
+Each append consists of a valid JSON event object, including
+timestamp and format TBD.
+
+
+=== Debug Log
+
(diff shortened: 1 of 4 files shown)


spec_16.adoc
```diff
@@ -0,0 +1,216 @@
...
+is used to add events to this log.
+
+Each append consists of a valid JSON event object, including
+timestamp and format TBD.
+
+
+=== Debug Log
+
+`jobs.active.<jobid>.eventlog` is a similar log available for
+capturing job-specific debug information from instance services.
```
grondo: typo here `.eventlog` should be `.debuglog`

garlick:
I'll go ahead and fix that and also s/container manager/exec service/ as that was confusion on my part.

spec_16.adoc
```diff
@@ -0,0 +1,216 @@
...
+the available resources.  It either rejects the job outright,
+or annotates it and eventually assigns resources to it.
+
+The _container manager_ takes up jobs that have assigned resources,
+instantiates containers for those jobs and starts job shells
+within them.  When job shells and their containers start to exit,
+the container manager "reaps" these shells and allows the _scheduler_
+to reclaim resources assigned to each container.
+
+Jobs therefore transition between the following states:
```
morrone: Job state transition will not always be linear.  We'll need a flow chart of valid state transitions.

garlick: Was that implied?  Didn't mean to, I'll add something to clarify for now.

grondo:
> Job state transition will not always be linear. We'll need a flow chart of valid state transitions.

Maybe the wording there could be cleaned up. I don't think it was meant that the job state would linearly transition between those states, but that those are expected states of jobs that are managed by instance services. (Perhaps my wording is bad too).

Does a job state transition diagram belong in this RFC (which is nominally about KVS schema for jobs), or do we need to modify (completely rewrite?)  RFC8?

garlick:
> Does a job state transition diagram belong in this RFC (which is nominally about KVS schema for jobs), or do we need to modify (completely rewrite?) RFC8?

I had the same thought.

spec_16.adoc
```diff
@@ -0,0 +1,216 @@
...
+
+
+=== Content Produced by Ingest Agent
+
+A user submits _J_ with attached signature.
+
+The _ingest agent_ validates _J_ and if accepted, assigns the jobid
+and creates `jobs.active.<jobid>.J-signed`.
+
+The _ingest agent_ logs the initial state "submitted" to the event log.
```
morrone: The creation of the event log needs to be described here too.  The event log section isn't mentioned until much later in the document after it is referenced several times.

spec_16.adoc
```diff
@@ -0,0 +1,216 @@
...
+Instance components have read/write access to the primary KVS namespace:
+
+* _Ingest agent_
+* _Job management service_ (list, cancel, etc.)
+* _Container manager_
+* _Scheduler_
+
+Guest components have read/write access to a private namespace, and
+limited access to the primary KVS namespace through the _job management_
+service:
```
morrone: I think that this part about primary namespace access should be reworded.  The flux-{run,queue,cancel} commands are never (as far as I know) given access to the primary KVS namespace, they only talk to various other services (ingest, management, scheduler, etc.).  Those services do in turn use the KVS . But I think it is strange to say that those comands "access" the KVS when they don't use KVS APIs and don't necessarily know anything about the structure of data in the KVS.  It would also seem to conflict with later rules like "The Flux instance has a default, shared namespace that is accessible only by the instance owner."

grondo: Maybe if it said "limited access *to job data in* the primary KVS namespace"?

morrone: I think it would be easiest to just remove everything after the comma.  For the most part, guest components cannot access anything in the primary KVS.  They must instead talk to services.

There are perhaps exceptions.  "flux kvs ls" might work in the primary namespace if the user employing the command is the same as the instance owner.

spec_16.adoc
garlick: Sounds good.

spec_16.adoc
garlick: It's a reasonable point - I'll fix that up some way.  It seemed useful to point out that guest components require help from a privileged service to get to the job data, but maybe that's obvious if they don't have direct access to it.

spec_16.adoc
```diff
@@ -0,0 +1,217 @@
...
+running::
+  all containers instantiated and job shells started
+
+exiting::
+  first container exited
+
+completed::
+  all containers exited
+
+rejected::
```
lipari: It is unclear to me what the fate of a rejected job is.  Does the user have the opportunity to correct the cause of the rejection and still maintain the original job?  Or will the user have to resubmit a corrected job?  If the former, then I would propose the term be changed from "rejected" to "invalid" or "ineligible".  This clues the user into the fact that they need to fix the job or it will remain pending forever.  If the latter and "rejected" means a replacement job with corrections must be submitted, then when and how is a rejected job purged from the kvs?

garlick: If rejected at ingest, the submit command would fail (no KVS entry created or job id assigned)

If rejected by the scheduler, I was thinking it would just short circuit to "inactive" with the state set to rejected (or could be one of your other suggestions).  The user would have to resubmit.

Maybe ingest should be made smart enough (say through scheduler plugin?) so that the second case is not possible?

grondo: Good point. What if we say for now, rejected jobs do not  make it to the KVS?

Once we have some way to update a job (requiring incremental signed changes to J), then we can add a rejected or some other state that allows the user to update an existing job to make it runnable?

lipari: Your suggestion sounds good to me, @grondo.

grondo:
It seems there is general consensus that this is a good starting point. Could you squash changes and re-push?

One thing we talked about (perhaps offline) was the addition of a general purpose `jobs.active.<jobid>.data` directory, for use by instance components or extensions that would like to store ancillary data along with a job. Each component would have its own subdir within this directory named after the service to avoid future collisions. E.g. a performance tracer might store per-job data under `jobs.active.<jobid>.data.trace`. The text could something like:

```
=== Content Produced/Consumed by Other Instance Services

Other services not mentioned in this RFC MAY store arbitrary data associated with jobs
under the `jobs.active.<jobid>.data.<service>` directory, where `<service>` is a name
unique to the service producing the data. For example, a job tracing service may store
persistent trace data under the  `jobs.active.<jobid>.data.trace` directory. 
```

If you prefer to leave this out of the RFC until there is a use case, I'd be fine with that.

garlick:
Thanks for bringing that up - that seemed like a good idea when we discussed it before.

I added your paragraph (verbatim) and squashed.

grondo:
Thanks!
Merging this as our starting point. 

Additions, revisions, corrections, modifications, qualifications, addenda, alterations, amendments, rewording, grammatical adjustments, and redactions can occur in future PRs.

garlick:
Thanks, sounds good.

PR #108, opened 2017-09-27
Attempt to simplify and clarify RFC 15 operations

grondo:
This is actually just a starting point for some modifications and clarifications to RFC 15 and IMP operation to hopefully simplify required behavior, as well as reduce external and circular dependencies in the eventual flux-security project.

Motivation:
 * Make some currently required functionality optional so we can circumscribe the smallest necessary functionality for S4 and earlier milestones
 * Clarify any external dependencies that will be required for the IMP and flux-security project, with a goal to eliminate most if not all dependency on functionality from flux-core or flux-sched

This first edit removes some required steps from the IMP verification which, after some thought, do not seem necessary. I don't go as far as to remove the requirement that *_R_* be signed by the resource owner, though I think perhaps that is possible. There is an explanation for why each of these verification steps might be optional in the text. Note that they are optional, but still *may* be performed in a full implementation.

As mentioned, the main goal is to eliminate potential dependencies of the IMP on libraries or code from flux-core or flux-sched, for obvious reasons. This means that the IMP should treat any objects in *_R_* and *_J_* generated by instance libraries (libjobspec, libresource or similar) as opaque data. The jobspec is already passed directly through to the job shell, so we are ok there, but how the assigned resource set in *_R_* is defined is probably the next piece to tackle.

However, I thought there was probably enough controversial stuff in this first commit that I'd  submit the PR and get some discussion going.


garlick:
One question:

From the RFC (unchanged by this PR):

> In most cases, the IMP will execute a job shell on behalf of the user, passing the verified J as input to the shell. The shell itself is provided either by the user or by system configuration, but should not be provided or modified by the instance owner. 

If the instance can't pick the shell, and the shell is started by the IMP, it seems like path to the shell has to be provided by the user in the signed J, and thus the IMP would need to verify J's authenticity and extract the shell path so it knows what to exec.

I'm not sure if that contracts what you're proposing.  Maybe the IMP could easily get at shell info in the J envelope without needing to parse jobspec.

grondo:
> If the instance can't pick the shell, and the shell is started by the IMP, it seems like path to the shell has to be provided by the user in the signed J, and thus the IMP would need to verify J's authenticity and extract the shell path so it knows what to exec.

Good point, I don't think we have specified where the shell path will be recorded, but since the shell acts as a kind of interpreter of jobspec, my feeling is that it should be specified as part of *_J_*, but not in the jobspec (and thus is in the libflux-security domain?) This should be added to the RFC if we're in agreement.

garlick:
Yes, that sounds right.

grondo:
Do you think the *job shell path* in *_J_* should be optional or required? If optional, and there is no path defined in *_J_*, the IMP would invoke a system default. If required, the job shell path could be filled in with a system default by the utility responsible for building *_J_* (which will probably be from the flux-security project)

I can see drawbacks and benefits of both approaches.

morrone:
I was kind of thinking that the jobshell would be a known application to the IMP.  In other words it would be specified in the IMP's configuration rather than changeable on a per-job basis.

If the submission command is deciding the jobshell and adding it to J, then probably 99.9% of the time the jobshell would be determined by the instance rather than the user (because I presume the submission command uses the instance's default value...which it might decidd is anythig).  That would seem to mostly defeat the purpose of having it in J.

Though, I suppose the submit commands might have a separate configuration file on a system rather than getting the default shell from the instance.  Maybe that isn't too different from the IMP having a system configuration file specifying the shell.

morrone:
> Do you think the job shell path in J should be optional or required? If optional, and there is no path defined in J, the IMP would invoke a system default. If required, the job shell path could be filled in with a system default by the utility responsible for building J (which will probably be from the flux-security project)

I like the idea of making it optional and the IMP having a configured default.

Outside of flux developers testing new job shells, do we have a use case for non-default job shells?

grondo:
> Outside of flux developers testing new job shells, do we have a use case for non-default job shells?

That is the main use case I had in mind -- support for testing. I'm thinking for S4 timeframe it is not a requirement. The way to test new flux code will be to run flux-start from your working copy as a job in the system instance.

However, since the job shell is the "interpreter" for at least the `tasks` section of jobspec, the future test case could be multiple system installed shells or custom shells for different workloads. I think we just want to have a story for how this could work. The main requirement is that user or the system controls the path to the shell, which in turn is then restrained in how it can interact with the instance.

> If the submission command is deciding the jobshell and adding it to J,

I was thinking (hoping) we'd design it such that the flux-security project would be the producer and consumer of J. 

> I like the idea of making it optional and the IMP having a configured default.

That is my preference as well

grondo:
Just pushed a small commit adding the *job shell path* to *_J_* as discussed. If missing, the IMP invokes a system default.

garlick:
It seems like we're all on the same page.  It's the IMP's default or an override from the user (not the instance), and the override must be signed by the user and verified by the IMP.

grondo:
In order to avoid a requirement that the IMP utilize resource query library or other code from flux-sched or core, I think we should require that the *instance* be responsible for determining the local resource set for each IMP, rather than having the IMP do the set intersection between the assigned resource set in *_R_*.

The local assigned resource set would have to be passed to the IMP separately from *_R_* (since *R* is global for the job). A common format for the localset would have to be defined such that the IMP plugins for containment could traverse the local resource set and create required cgroups/containers, etc, without the use of helper libraries from the enclosing instance.  This format could be the same format used for the job "assigned resource set", but it would not be required.

A big assumption here is that the IMP can definitively track the "container" in which it is invoked, and that all IMP containment plugins ensure that sub-containers can only include resources from the current container. This reduces/eliminates the risk of a user trying to expand resources available to an instance they own, or running the IMP from the cmdline with bogus input, etc.

This is just an idea at this point but later I'll try to fold it into the RFC (after discussion here)

grondo:
>  It's the IMP's default or an override from the user (not the instance), and the override must be signed by the user and verified by the IMP.

Well said, can you check my latest commit? I tried to actually change as little as possible and it may not have ended up this clear.

garlick:
Would it be clearer if "system default" was "IMP default" and "system configuration" was "IMP configuration" when referring to default job shell?

grondo:
> Would it be clearer if "system default" was "IMP default" and "system configuration" was "IMP configuration" when referring to default job shell?

Yes, probably. Good call.

morrone:
> In order to avoid a requirement that the IMP utilize resource query library or other code from flux-sched or core, I think we should require that the instance be responsible for determining the local resource set for each IMP, rather than having the IMP do the set intersection between the assigned resource set in R.

I'm a little concerned about that approach.  flux explicitly treats _all_ resources as abstract resources.  If the IMP no longer accepts resource type handling through plugins, then that limits what we can do on a node to what is explicitly built into the IMP.

Have you given any thought into what instance component would do that?  I would suggest somewhere like the execution module.

But then, this would seem to eliminate the need for R.  Perhaps we instead only need 'r', the unsigned description of all assigned resources as assigned by the scheduler.  The individual "local resource set" components will instead each need to be signed.

Or, how about this!

The scheduler generates 'r' (the complete list of assigned resources).  The next module in flux then interprets 'r' against the jobspec from j, generating all of the "local resource sets".  All of those local resource sets could still be concatenated into one big R document.  But now the IMPs can use a simpler parser to find their local assigned resources than if they needed to get that information from 'r'.

But the more that I think about this, the less convinced I become that avoiding parsing 'r' from 'R' in the IMP really makes sense...

Could you take another stab at explaining that? :)

grondo:
> I'm a little concerned about that approach. flux explicitly treats all resources as abstract resources. If the IMP no longer accepts resource type handling through plugins, then that limits what we can do on a node to what is explicitly built into the IMP.

Sorry! I put half of a half formed thought in here! I didn't mean that the IMP wasn't extensible by plugins anymore, but that the per-resource-type plugins which implement containment should not be dependent on libraries from flux-core or flux-sched. I was proposing that the IMP take a perhaps simplified format which its plugins would then either use directly or via some library in flux-security to walk the list of resource types they care about and implement containment (or whatever it is they need to do).

Another driver for this is so we can reach the "standalone testability" goal of the flux-security project.

> Have you given any thought into what instance component would do that? I would suggest somewhere like the execution module.

Yes, I was thinking exactly the same.

> But then, this would seem to eliminate the need for R. Perhaps we instead only need 'r', the unsigned description of all assigned resources as assigned by the scheduler. The individual "local resource set" components will instead each need to be signed.

I was thinking along the same lines earlier. (BTW, I like your use of 'r' for the "assigned resource set")
I think *R* is evolving into the structured "**R**equest" that the instance sends to the IMP. It does need to contain, at least by reference *J* and some form of 'r' (local or global), though I'm beginning to think the other fields are probably unnecessary. I need more time  to think it through though.

> The scheduler generates 'r' (the complete list of assigned resources). The next module in flux then interprets 'r' against the jobspec from j, generating all of the "local resource sets". All of those local resource sets could still be concatenated into one big R document. But now the IMPs can use a simpler parser to find their local assigned resources than if they needed to get that information from 'r'.

I like this, but I'm not sure you can derive local resource sets from 'r' and jobspec, or at least I was
not thinking of it that way. Please let me know if I am wrong, but determination of the "local resource
sets" is a function of the enclosing instance, and can be done solely from 'r'. The enclosing instance
decides the level at which it wants to invoke the IMP, and each IMP runs on a single "local resource set". This set is then used to create the container into which the jobshell is placed. The common case will
be one IMP per flux-broker, so the entity generating local resource sets would just need to match the resource sets within 'r' with ranks. My thought was that an execution module would do this on each rank
just before invoking the IMP.

> But the more that I think about this, the less convinced I become that avoiding parsing 'r' from 'R' in the IMP really makes sense...
> Could you take another stab at explaining that? :)

These were some of my arguments, reworded: 

 * Derivation of the local resource set from 'r' may be complex and require libraries provided by the local instance, an untrusted source. If a resource library or framework is required for this operation, it also complicates dependencies and testing of flux-security project, for which the goal is
standalone testing and independent version evolution. 

 * The enclosing instance is best equipped to determine the shape and location of containers for jobs it is managing, and thus is the correct place to compute local resource sets from 'r'. 

 * Requiring the IMP to derive local resource set from 'r' directly will make it impossible to create multiple containers/shells per flux-broker, though this is not a requirement now.

Thanks, this is the discussion I was hoping to have.

morrone:
> I like this, but I'm not sure you can derive local resource sets from 'r' and jobspec, or at least I was
not thinking of it that way. Please let me know if I am wrong, but determination of the "local resource
sets" is a function of the enclosing instance, and can be done solely from 'r'.

Oh, right!  We did discuss that elsewhere.  Regardless of the "shape" of the tasks inside of the node, and the further containment that the job shell applies, the IMP just concerns itself with a single container of all resources under the node. So, yes, that makes sense.

> The common case will be one IMP per flux-broker, so the entity generating local resource sets would just need to match the resource sets within 'r' with ranks.

OK, I can see that.  It can split the resources at "node" type resources and then map the node names to flux ranks.

> My thought was that an execution module would do this on each rank just before invoking the IMP.

OK.  That implies that even the system-level instance needs to distrubute its private key to all nodes.  But no doubt hierachical scheduling will make that a moot point anyway.  Application tasks will need to do some form of discovery to find the full set of its assigned resources (if they need to know that information, such as a child flux instance), but applications running under flux already employ discovery through the flux instance to find each other, so perhaps also using it to look up the full set of assigned resources is no big deal either.

garlick:
> OK. That implies that even the system-level instance needs to distrubute its private key to all nodes. But no doubt hierachical scheduling will make that a moot point anyway. Application tasks will need to do some form of discovery to find the full set of its assigned resources (if they need to know that information, such as a child flux instance), but applications running under flux already employ discovery through the flux instance to find each other, so perhaps also using it to look up the full set of assigned resources is no big deal either.

I think I got your main point:  exec system on a particular rank needs private key to sign 'r' before passing it to IMP, therefore private keys need to be distributed.   But the argument for why it's moot sort of lost me. Have we worked out how an instance obtains R or r from the enclosing instance, and how peer instances would discover each other?

grondo:
> I think I got your main point: exec system on a particular rank needs private key to sign 'r' before passing it to IMP, therefore private keys need to be distributed.

Just to make sure I'm still on the same page, I was making the point that signing R (which is 'r' plus other input from the instance to the IMP) is not mandatory.  Also, I wasn't arguing that a per-rank or per-IMP 'local resource set' *replace* the full assigned resource set 'r' in R, but that it will be an addition to the IMP's input.

grondo:
> But the argument for why it's moot sort of lost me. 

Hm, when I read @morrone's statement I thought he meant we need to support a *distributed* scheduler within an instance, but now I reread the comment and it does say "hierarchical" (but pretty sure he meant that a scheduler may be operating on more than one rank)

morrone:
> I think I got your main point: exec system on a particular rank needs private key to sign 'r' before passing it to IMP, therefore private keys need to be distributed. But the argument for why it's moot sort of lost me. 

I meant that hierarchical flux instances (where some set of the top-level instances are owned by the same user as the top-level system instance) will mean that even if the signing was somehow constrained to broker 0 of an instance, the private key would still wind up out on a bunch of the compute nodes in one form or another.  Therefore not constraining signing to rank 0 probably doesn't add any additional risk.

> Just to make sure I'm still on the same page, I was making the point that signing R (which is 'r' plus other input from the instance to the IMP) is not mandatory. Also, I wasn't arguing that a per-rank or per-IMP 'local resource set' replace the full assigned resource set 'r' in R, but that it will be an addition to the IMP's input.

If the IMP actually uses the "other input from the instance to the IMP" in R, than R must be signed.  But that 'other input from the instance to the IMP' could be combined with the 'local resource set', and that combination could be signed.  In effect, that would be replacing 'r' with 'local resource set'.

> Have we worked out how an instance obtains R or r from the enclosing instance, and how peer instances would discover each other?

I think those two questions are distinct.

For the first one: no.  But I can imagine a couple of ways:

1) Simply query a service in the parent instance, "what is my full 'r'?".
2) Use PMI or the job's private KVS namespace more directly to exchange all of the 'local resource set' bits and combine them.

Others are probably possible.

How peer instances find each other, I have no idea.

garlick:
> Simply query a service in the parent instance, "what is my full 'r'?".

Ah, another item for RFC 16 probably, which says:
> The _exec service_ takes up jobs that have assigned resources, ...

Perhaps the "assigned resources" would be dropped into `jobs.active.<jobid>.r` or similar.

grondo:
> If the IMP actually uses the "other input from the instance to the IMP" in R, then R must be signed. But that 'other input from the instance to the IMP' could be combined with the 'local resource set', and that combination could be signed. In effect, that would be replacing 'r' with 'local resource set'.

Yes, I'm arguing to do largely what you say in your final statement here (other input TBD).
However, from the changes to the document here, IMP verification of the signature on its input is now optional. Since only the IMP receives its own input, and it can verify the identity of the creator based on the real UID, I didn't see what verification by signature bought us in the near term exactly.

grondo:
>Perhaps the "assigned resources" would be dropped into jobs.active.<jobid>.r or similar.

Isn't `R`, which already contains 'r', alread in `jobs.active.<jobid>.R`?

garlick:
> Isn't R, which already contains 'r', alread in jobs.active.<jobid>.R?

Second "oh duh" of the morning.  Maybe I should ponder more, comment less :-/

morrone:
> Since only the IMP receives its own input, and it can verify the identity of the creator based on the real UID, I didn't see what verification by signature bought us in the near term exactly.

That I didn't understand.  The IMP is getting input from an unprivileged flux instance.  It needs the signature to verify that the input is from a trusted user, because it should only act on the actions requested by a trusted user.

When you say "verify the identity of the creator based on the real UID", are you meaning that other normal Linux mechanisms would be used to verify the UID of the process that is connected to the IMP?  

Would that also work for sub-instances?  For instance System instance grants resources to Alice, and Alice in turn wants to grants some of her assigned resources to Bob.  Does the IMP have enough information to know that Alice has authority over those resources?  The standing plan was that there would be a chain of signed R documents demonstrating ownership back to some trusted user (the System instance user).

grondo:
> When you say "verify the identity of the creator based on the real UID", are you meaning that other normal Linux mechanisms would be used to verify the UID of the process that is connected to the IMP?

The current plan was that the IMP is a setuid helper, executed by the enclosing instance, and that the request, R, is sent to the IMP on stdin (maybe not the best plan, but that is what I assume we're doing for now). The IMP can determine the executing user via getuid(), and can therefore be reasonably assured that its *input* came from that user.

At this point the IMP has been given authority to transfer local resources from the requesting user A, to the guest B. It creates a container for B, along with some state for the container, switches into the container and changes credentials to B, invoking the jobshell.

Now if guest B created a new instance, it is running in the container B created by A, along with IMP managed state that says user B has ownership of these local resources (could even just be a chown on the cgroup directory for instance). If user B invites guest C to run on their instance, then B invokes the IMP (and is in fact the only user that *can* invoke a process inside of container B). The IMP verifies that getuid() matches the owner of the current container (possibly using extra state associated with the container), and that J is signed by C, it then can create container C inside of B, with associated state, and so on...

I'm not saying that the IMP definitely should not verify a signature on its input, but it doesn't seem required, since the IMP can verify the identity of the "caller" with getuid().

grondo:
> The standing plan was that there would be a chain of signed R documents demonstrating ownership back to some trusted user (the System instance user).

My current approach is to attempt to simplify required IMP operations to the minimum requirement for S4. I can remember us discussing a case for why we needed the full chain of signed R documents back to the system owner, but working back through the normal case I couldn't see why  intially we can't leverage some concept of container ownership. That is, if a user can only run a process in a container they own, then they can only start the IMP in a container with resources they own, and therefore the IMP can assume a user running the IMP has authority over all resources in the current container.

@garlick brought up a good question about non-flux managed resources, e.g. login nodes. There are several approaches to mitigate anything going wrong there, for example by stealing a singed J from and faking an R:

 1. The IMP won't be setuid on systems where jobs are not run
 2. The IMP's "state" associated with containers won't be present if the system instance isn't running there
 3. If the system instance is runnning on login nodes, then the IMP will be able to tell when it hasn't been started within a "system container" and can fail (and log a security alert)


I should mention that all this is completely open for discussion.

morrone:
I can't find a problem with that logic.  But like you said, most of the "other stuff" in R becomes irrelevant.

In your pull request you still have "+ 4. Verify TTL on *_R_*".  But the TTL only has meaning when signatures are involved.

grondo:
> I can't find a problem with that logic. But like you said, most of the "other stuff" in R becomes irrelevant.

> In your pull request you still have "+ 4. Verify TTL on R". But the TTL only has meaning when signatures are involved.

Thanks, I should fix that. Maybe we could, for now, turn it into an optional enhanced security or additional security section of the IMP's input? This is not my area of expertise so I don't want to go too far without lots of help.

grondo:
> But like you said, most of the "other stuff" in R becomes irrelevant

Yes, as I think it through further, the required input for the IMP is:

 * The local assigned resource set, effectively the contents of the "container" the IMP will attempt to create, as determined by the instance
 * uid or username of the user that created *J* (don't know how this was missing before)
 * *J*

The optional "with enhanced security" fields would include:

 * Timestamp
 * TTL
 * UUID
 * Signature of all other fields

If a uuid is already present in either 'r' (the global resource set) or the job itself, then the uuid here might be redundant.

I'm still going through the current RFC 15 to see how this simpler approach could be merged in. Either way it is a bit of a change (I think the idea of a global 'R' goes away, which affects other RFCs as well), so some further thought and discussion needed.

morrone:
> uid or username of the user that created J (don't know how this was missing before)

The UID is already in J.

> The optional "with enhanced security" fields would include:

Those all sound like things that would be part of the security envelope rather than things needed by the contents.  Maybe it is best not to mix the two at this time.

If these things aren't going to be implemented in S4, perhaps it would be best to remove them from the RFC for now.  If we are not likely to discover problems with our design (because we are not implementing it) and we no longer have a clear use case for it, I am not sure there is a good reason for it to remain in the RFC.

Does any one have a use case in mind?

grondo:
The uid is in J but is needed to verify J, which was a bit awkward in the proof of concept we did. 

I don't have a use case in mind for the security envelope, and would be ok with removing it for now. 

(Actually, the use case is if the IMP were implemented as a service, instead of a setuid executable. However, I'm not sure how other parts of the exec service could work in that mode so probably better left out for now)

morrone:
> The uid is in J but is needed to verify J, which was a bit awkward in the proof of concept we did.

Did the proof of concept binary sign J (basically encrypting) rather than clearsigning?  If it is clearsigned, I don't see why getting J from the document would be awkward.  What am I missing?

grondo:
> Did the proof of concept binary sign J (basically encrypting) rather than clearsigning? If it is clearsigned, I don't see why getting J from the document would be awkward. What am I missing?

Yes, sorry. The proof of concept signing method couldn't fully parse *J* until it after verification. However, I think this will need to be fixed since the instance will need to read the jobspec in *J* and doesn't necessarily want to verify *J* (it gets the correct userid from the flux message).

However, It also seemed awkward to use data from *J* in its verification, though in this case that seems pretty safe.

I looked back through this RFC, and actually UID was never specified as part of *J*. I slightly prefer keeping UID outside of *J* so that the IMP can verify *J* before using data from it, but if there is a strong argument for including it in *J*, that would be ok too.

morrone:
> I looked back through this RFC, and actually UID was never specified as part of J.

You are right.  I must have misread UUID as UID.  I'm for adding it to J.

If it isn't in J, then it needs to be tracked separately in multiple places, at least: submission time, in the KVS, execution time (passed to the IMP).

At submission time, that would pretty much force us to expand the messaging interface to allow multiple payloads, or we would have to wrap J in another wrapper to add the external user information field which definitely gets awkward.

garlick:
Was there an argument to leave the userid out of J itself so that users could share them?

grondo:
> Was there an argument to leave the userid out of J itself so that users could share them?

I don't think so, I think it was just forgotten.

J can't be shared or even reused by a single user since it has a Timestamp and TTL. 
I think we do want to allow jobspec to be reused, so we definitely should not add user id there.

grondo:
> At submission time, that would pretty much force us to expand the messaging interface to allow multiple payloads, or we would have to wrap J in another wrapper to add the external user information field which definitely gets awkward.

No the uid is already part of the multi-user capable message format. See [RFC 12](https://github.com/flux-framework/rfc/blob/master/spec_12.adoc). 

I think the flux message UID will be recorded in the KVS anyway, since services that look up job information on behalf of guest users will need to quickly know job ownership in order to decide what level of job information to relay from the KVS, and shouldn't have to parse all of J to get it...

morrone:
> No the uid is already part of the multi-user capable message format.

The connecting user, but not necessarily the user that requested the job, right?  For instance, one fluxuser owned (system level) flux instance might decide to pass a J along to another fluxuser owned instance.  fluxuser didn't create J though, someone else did.  If the J contains the user info, then it could simply reuse the same submission interface to pass it along.

And really, J _does_ contain the user info.  It is just really hard to figure out because we need to brute force try all public keys until we figure it out.  I see no reason not to make decoding easy and keeping it in J.  After all, isn't that how signing email works?  The recipient looks at the "From" field (or something similar, I assume) and uses that information from _inside the email_ to figure out which public key to use to try to verify the crytpographic signature.  We don't have to send a second email to explain where the first one came from. :)

grondo:
> After all, isn't that how signing email works? The recipient looks at the "From" field (or something similar, I assume) and uses that information from inside the email to figure out which public key to use to try to verify the crytpographic signature. We don't have to send a second email to explain where the first one came from. :)

I'm not entirely sure, but I thought that the message body is signed in an email, and that headers weren't typically included? (Doesn't the `----BEGIN SIGNED MESSAGE---` go around the signed part?). How would it even work otherwise for encrypted email?

Your other point is good though. It does imply that any user can submit a job on behalf of any other user (whether they authorized it or not), but perhaps this could be disallowed by ingest policy based on user roles. However, it does mean you are always storing uids in multiple places anyway.

I'm not feeling strongly either way at this point, so I'll add a uid field to J.

morrone:
> I'm not entirely sure, but I thought that the message body is signed in an email, and that headers weren't typically included? (Doesn't the ----BEGIN SIGNED MESSAGE--- go around the signed part?). How would it even work otherwise for encrypted email?

Yes, that is true.  But it isn't really including the username/uid in the signed part that I care about so much as I do including it in the same document in a cleartext form.

The J is almost, but not _quite_, a complete representation of a user's job request.  The only thing missing is who requested it (except in the brute-force form which isn't terribly usable).  Making it a complete, self-contained request without requiring extra information to be carried in parallel wherever J goes seems elegant to me.

> I'm not feeling strongly either way at this point, so I'll add a uid field to J.

Cool!

grondo:
> The J is almost, but not quite, a complete representation of a user's job request. The only thing missing is who requested it (except in the brute-force form which isn't terribly usable). Making it a complete, self-contained request without requiring extra information to be carried in parallel wherever J goes seems elegant to me.

Yes, good argument. Thanks!

grondo:
Any preference for notation here? I'm not sure if we've established a convention, but are signed object capitalized, and unsigned lowercase?

You have introduced *r* as the global assigned resource set for a job, any similar good idea for the local resource set? (I'd choose 'l', but is it a one or an ell, nobody every knows)

Also the instance input to the IMP is no longer signed, and it differs for each IMP, should it still remain the "request" *R*, or something else?

grondo:
I've made a *very rough* pass through rfc15 and removed reference to *_R_*. *R* is now simply "Input to the IMP", and is not signed.

I also made a perhaps controversial addition to both *J* and the IMP's input -- *J* gets a set of "options for the IMP" set by the requesting user, and "input to the IMP" gets a similar "options" field set by the resource owner.  Something like this will be necessary to allow IMP plugins with optional per-job behavior (e.g. like spank plugins in Slurm). The user-specified options need to go in *J* and not the jobspec in *J* because we want to avoid parsing jobspec in the IMP if at all possible. I figured options set for the IMP by the resource owner might also be necessary so added that to the generic "input to the IMP".

See the commit log for more specific changes from this edit. (I guess I need to update the figure as well)

I expect the document is still really rough with this kind of ad hoc edit, but I wanted to get some of the details down so there is time for discussion.

garlick:
With **R** freed up, what about **R** as the full resource set assigned to the job and **r** as the local resource set?

Alternatively maybe subscripts would be clearer, e.g. **R<sub>global</sub>** and **R<sub>local</sub>**?

http://www.methods.co.nz/asciidoc/userguide.html#_superscripts_and_subscripts

garlick:
The diagram still depicts ( R(J) ) as input to the IMP.  It should reflect that J isn't "inside" a signed resource token anymore, and the resource token needs to represent the local resource set not the global one.

grondo:
> With R freed up, what about R as the full resource set assigned to the job and r as the local resource set?
> Alternatively maybe subscripts would be clearer, e.g. Rglobal and Rlocal?

Good idea. I like either approach.

> The diagram still depicts ( R(J) ) as input to the IMP. It should reflect that J isn't "inside" a signed resource token anymore, and the resource token needs to represent the local resource set not the global one.

Yes sorry I had mentioned that I still need to modify the diagram:
> . (I guess I need to update the figure as well)

But I needed to have a new symbol for the "input to the IMP". Which perhaps we can just denote as something like  `{ r, opts, J, ... }` ?

garlick:
I personally like  **R<sub>local</sub>** for the local resource set, unless that is inconvenient for some reason.

Do we need to call out epilogue execution or a "teardown" phase for containment plugins?  (possibly this is off topic for this PR).

grondo:
> Do we need to call out epilogue execution or a "teardown" phase for containment plugins? (possibly this is off topic for this PR).

This was in an early draft of the document, but was removed before it was merged.
See this [comment](https://github.com/flux-framework/rfc/pull/79#issuecomment-287906678)

grondo:
> I personally like *R<sub>local<sub>* for the local resource set, unless that is inconvenient for some reason.

It seems fine to me, a little more verbose, but lends itself to other subscripts like **R<sub>jobX<sub>** -- is it ok to shorthand the "global assigned resource set" for a job as simply **R**? Oh, also since we are talking about sets of resources here, we could borrow other set notation constructs and do things like

 { *R<sub>1<sub>*, *R<sub>2<sub>*, ..., *R<sub>n<sub>* } = **R**

garlick:
Sounds good!

grondo:
Diagram updated.

garlick:
LGTM!

morrone:
> I also made a perhaps controversial addition to both J and the IMP's input -- J gets a set of "options for the IMP" set by the requesting user, and "input to the IMP" gets a similar "options" field set by the resource owner. Something like this will be necessary to allow IMP plugins with optional per-job behavior (e.g. like spank plugins in Slurm). The user-specified options need to go in J and not the jobspec in J because we want to avoid parsing jobspec in the IMP if at all possible.

Can you expand on what use cases there might be for the user talking to the IMP?  I thought we wanted to avoid that, and have only the parent instance the the new r-local be what controls the creation of containers.  I would think that a normal user _would_ actually put system options in the attributes->system area of the jobspec, and then if the _scheduler_ decided it needed some extra plugin functionality from the IMP to satisfy the request, it would put something in r, and the execution engine would in turn put that thing in r-local.

I didn't think that the IMP would even look at most of the fields of the J.  Instead those fields are largly specific to a particular security plugin.  Different security plugins might have different metadata requirements before applying the cryptographic signature.  They might use different envelope formats too.

If there are good use cases, then it would be nice to hear what the user interface is going to look like for submitting a jobspec, because it sounds a little complicated to no longer have the jobspec be a nice self-contained, reusable job request.  Putting some parts to replicate a job run in jobspec and others in the J envelope.  Parts of the J envelope cannot be generated ahead of time...so users would need to remember the jobspec and a partial J.

PR #111, opened 2017-10-06
rfc16: replace R-signed with "R" in kvs schema

grondo:
Update kvs schema naming of *R* to remove the `-signed` suffix, since *R* is no longer required to be signed per latest RFC15.

### DIFF
--- a/spec_16.adoc
+++ b/spec_16.adoc
@@ -175,7 +175,7 @@ store in the KVS for recovery.  Annotations are stored as
 keys under `jobs.active.<jobid>.scheduler`.
 
 Upon resource allocation to the job, the _scheduler_ creates
-`jobs.active.<jobid>.R-signed` and logs a _runnable_
+`jobs.active.<jobid>.R` and logs a _runnable_
 state transition to the event log.
 
 The _scheduler_ may later revoke the allocation (TBD).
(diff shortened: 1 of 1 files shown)


dongahn:
Looks good to me. 

Since this is a PR for RFC16, let me ask one related question.

What would `J.signed` look like?

I suppose the scheduler doesn't need to verify the authenticity of `J.signed,` but just create a jobspec object from `J.signed` and emit a `R`. Would the signed J different than plain J? Or you would have two fields, a plain J and a signature of J? Sorry I'm not well versed on this topic.

grondo:
I think the intent is that *J* and *J-signed* are the same and that you can read the fields of *J* without signature verification. I think the language of RFC15 says that the signature is included in *J* and is a signature of the rest of the fields in *J*.

dongahn:
Great. That is completely aligned with what I thought this would be! My guess is I can simply construct a jobspec object directly out of the raw `J.signed` string using @morrone's `libjobspec` when all is said and done.

dongahn:
As far as I can tell, this PR is good. I can push the merge button if you want @grondo.

grondo:
Yes, go ahead since it is a minor change

PR #114, opened 2017-10-11
spec_16.adoc: copy R to guest KVS namespace

grondo:
Add text allowing data in primary KVS namespace to be copied into
the guest KVS namespace, granting "read-only" access to the data
for the guest user.

Specifically mention that "R" SHALL be copied by the scheduler into
the guest kvs namespace.

It isn't clear on quick read *when* the guest KVS namespace is created.
If the job doesn't have a guest kvs namespace until after it is scheduled, then
we could change this to have the exec service copy *R* to `guest.`*R*.

### DIFF
--- a/spec_16.adoc
+++ b/spec_16.adoc
@@ -113,6 +113,9 @@ this namespace is mounted in the primary namespace, on
 by the guest components without impacting performance of the primary
 namespace, while still being accessible through the primary namespace.
 
+Access to contents of the primary KVS namespace MAY be granted to
+the guest by copying them to the guest KVS namespace.
+
 When the job is inactive, the final snapshot of the guest namespace
 is linked into the primary namespace, and the guest namespace is
 destroyed.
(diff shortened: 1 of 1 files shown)


garlick:
> It isn't clear on quick read when the guest KVS namespace is created.
If the job doesn't have a guest kvs namespace until after it is scheduled, then
we could change this to have the exec service copy R to guest.R.

I think the exec service should be responsible for both creating the guest namespace and destroying it (after linking content back to primary).  If that's OK, then probably it would do the copying too.

Should we maybe call out a path for R, like say it is linked  to `jobs.active.<jobid>.guest.input.R`?  Alternatively could we just take a snapshot of the whole `jobs.active.<jobid>`?  The advantage of the latter is that the execution system can just do that and not need to know more detail.

garlick:
Want me to push a proposed commit to this PR ?   I hate to stick you with all the editing.

grondo:
Sorry, heading out to parent/ teacher conference. Feel free to push changes
or just make a new PR.


There's a lot of detail in the main kvs namespace, so I'm not sure all of
that should be shared wholesale with guest.

grondo:
Closing, moved to #115

PR #115, opened 2017-10-12
rfc16: clarify who creates/destroys guest namespace and content

garlick:
This builds on pr #114 as discussed there.

### DIFF
--- a/spec_16.adoc
+++ b/spec_16.adoc
@@ -106,9 +106,9 @@ and purged to keep its size manageable in long-running instances.
 
 === Guest KVS Namespace
 
-A guest-writable KVS namespace is created for the use of the
-job shell and the application.  While the job is active,
-this namespace is mounted in the primary namespace, on
+A guest-writable KVS namespace is created by the _exec service_
+for the use of the job shell and the application.  While the job
+is active, this namespace is mounted in the primary namespace, on
 `jobs.active.<jobid>.guest`.  While mounted, it can be changed
 by the guest components without impacting performance of the primary
 namespace, while still being accessible through the primary namespace.
@@ -116,9 +116,9 @@ namespace, while still being accessible through the primary namespace.
 Access to contents of the primary KVS namespace MAY be granted to
 the guest by copying them to the guest KVS namespace.
 
-When the job is inactive, the final snapshot of the guest namespace
-is linked into the primary namespace, and the guest namespace is
-destroyed.
+When the job transitions to inactive, the final snapshot of the
+guest namespace content is linked by the _exec service_ into the primary
+namespace, and the guest namespace is destroyed.
 
 
 === Event Log
@@ -178,8 +178,8 @@ store in the KVS for recovery.  Annotations are stored as
 keys under `jobs.active.<jobid>.scheduler`.
 
 Upon resource allocation to the job, the _scheduler_ creates
-`jobs.active.<jobid>.R`, copies this to `jobs.active.<jobid>.guest.R`,
-and then logs a _runnable_ state transition to the event log.
+`jobs.active.<jobid>.R`, and then logs a _runnable_ state transition
+to the event log.
 
 The _scheduler_ may later revoke the allocation (TBD).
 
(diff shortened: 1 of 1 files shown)


grondo:
Thanks! I'll close #114.

The segregated guest namespace seems like a good idea. Does `R` go under `system.exec.R` because `exec` subsystem copied it there (the origin is actually `sched`)? Maybe I'm overthinking it, but since exec creates the guest namespace and copies any data into it, wouldn't *everything* be under `system.exec.R`, so could we just say `system.R` (and, in general, perhaps preserve the paths from primary kvs namespace)?

Also, prefixing `user.` to all the paths in the "user" namespace seems a bit redundant, but I could accept it. It does feel like a good idea to set some rules for each namespace, but maybe there should be a `user` directory that is ok for arbitrary user-generated data, then a `shell` directory or something for job shell data?

Maybe out of scope here, but does it make sense to preserve the entire guest namespace, including `guest.system` which is likely a copy of data already in the primary namespace?

Sorry for the scattered thoughts.

garlick:
> The segregated guest namespace seems like a good idea. Does R go under system.exec.R because exec subsystem copied it there (the origin is actually sched)? 

Yeah that's what I was thinking

> Maybe I'm overthinking it, but since exec creates the guest namespace and copies any data into it, wouldn't everything be under system.exec.R, so could we just say system.R (and, in general, perhaps preserve the paths from primary kvs namespace)?

The choice of "system" was maybe a bad one.  I was trying to separate flux services (which can be tightly controlled by RFC) from non-flux services.  Maybe "flux"?

If the rule is that we create a subdir for each service to place the data it produces, then I was thinking `system.exec`, `system.shell`, etc.

> Also, prefixing user. to all the paths in the "user" namespace seems a bit redundant, but I could accept it. It does feel like a good idea to set some rules for each namespace.

Yeah maybe we could skip system/user so the top level would contain e.g. `exec`, `shell`, `application`.  Tools could create `totalview`, `scr`, etc..?

> Maybe out of scope here, but does it make sense to preserve the entire guest namespace, including guest.system which is likely a copy of data already in the primary namespace?

Hmm, well copies are cheap, but maybe not.  TBD?

grondo:
> If the rule is that we create a subdir for each service to place the data it produces, then I was thinking system.exec, system.shell, etc.

Ah, ok. I misunderstood and thought `system` was for data from enclosing instance and `user.` was for data from shell and friends.

I like your idea to simplify the top level to `exec`, `shell`, etc -- it is making more sense to me now.

garlick:
> I like your idea to simplify the top level to exec, shell, etc -- it is making more sense to me now.

Although probably leaving the top level open for "squatting" by tools, as implied by my example, is a bad idea.  Maybe confine the non-flux components to "tools" and "application" but flatten the flux components?

grondo:
> Although probably leaving the top level open for "squatting" by tools, as implied by my example, is a bad idea. Maybe confine the non-flux components to "tools" and "application" but flatten the flux components?

Yeah, that make sense to me. Probably good to keep the common case clean and simple for flux components, but still open the namespace for other use cases. Something between ad-hoc directory with everything dropped in the top-level, and onerous rules like Java package naming `gov.llnl.scr.v2` seems like a good place to land.

grondo:
Ok, is this ready for a merge? Up to you if you want to squash any commits.

garlick:
It's fine with me as is I guess.

dongahn:
I skimmed through this and the change looks good to me as well. 

The only input I have is, if the long directory will continue to have the overall KVS performance impact, I think we should avoid using flat organization as much as possible. There will be a limited set of flux services, so the flat structure for these services make sense to me. But applications can potentially abuse user space, introducing more rigorous hierarchical scheme makes sense to me. My $0.02.

garlick:
If I understand:  you have two concerns - 1) lack of structure in tools and application dirs, and 2) performance impact of abuse of the guest namespace.  Valid and worth pondering, thanks.

I was thinking about abuse of the guest namespace earlier with the thought experiment of dumping wholesale child KVS contents in there.  It may be that the enclosing instance (especially a system instance) needs to protect itself against DoS type situations from jobs, but I would say we should just keep that in mind for now and see how it goes in S4 timeframe.  Maybe we won't want to "advertise" the guest namespace too much until we have a plan for this.

If we wanted to impose more structure in tools or applications, we could do that in another PR.  I thought that was maybe a little premature at this point though (taking baby steps now).

dongahn:
> If we wanted to impose more structure in tools or applications, we could do that in another PR. I thought that was maybe a little premature at this point though (taking baby steps now).

Agreed. 

BTW, as a heads-up, @garlick and @grondo, I will soon create a ticket to discuss a need to have a scheduler/resource service specific area as part of the KVS job schema. 

@morrone and I discussed an item that came up at our last Monday meeting. @morrone's concern on the proposed `resource` service is that, while this service servicing the reservation requests, it won't be able to process user commands such as job status queries in a timely manner.

So I am thinking, maybe the scheduler simply dump a snapshot of schedule information to this KVS area at the end of each schedule loop, and have another module (job management module?) service these user commands from that snapshot.

garlick:
> BTW, as a heads-up, @garlick and @grondo, I will soon create a ticket to discuss a need to have a scheduler/resource service specific area as part of the KVS job schema.

Should be fine.  We've already stated that each instance component has its own area in the job schema.  We can call out specifics in the RFC when you're ready to do that.

> maybe the scheduler simply dump a snapshot of schedule information to this KVS area at the end of each schedule loop, and have another module (job management module?) service these user commands from that snapshot.

Worries me a bit, depending on frequency and how much data.

dongahn:
> Worries me a bit, depending on frequency and how much data.

Yup. This should be a part of the discussions... Thanks.

garlick:
> Yup. This should be a part of the discussions... Thanks.

Could the snapshot be shared directly with the other module (messaging, shared file,  ...)  The distributed consistency and persistence of the KVS don't seem like they are providing much benefit in this particular case?

PR #122, opened 2018-05-03
rfc6: allow multiple RPC responses

garlick:
This loosens RFC 6's definition of an RPC to allow for more than one response to a request.  This practice has been used in our code base for some time but never formalized.

My biggest concern about allowing this without some sort of open/close stream abstraction was the potential for a matchtag to be reused in another RPC, creating hard to find, intermittent bugs.  To address this concern, this PR proposes that clients and servers agree that any error response (no payload, errnum != 0) should signifiy the end of an RPC, and suggests that services could optionally use an error response of ENODATA as an EOF of sorts.

### DIFF
--- a/spec_6.adoc
+++ b/spec_6.adoc
@@ -35,7 +35,7 @@ by comms modules.  Flux RPC has the following goals:
 == Implementation
 
 A remote procedure call SHALL consist of one request message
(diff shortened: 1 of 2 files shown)


grondo:
Looks good, thanks for taking this extra step.

PR #130, opened 2018-07-11
RFC 17: describe interfaces for execution and submission of Flux jobs

grondo:
The purpose of this RFC is to collaborate and eventually specify the base interfaces in Flux for the submission and execution of jobs, i.e. define the UI for the `flux job run` and `flux job submit` commands. I also added the potential for a section on APIs, however, it is expected the API will be simple since it needs to only take (signed) jobspec, so perhaps that doesn't belong here.

The goal of the RFC is define the interface we'll be happy with (for a long time, as @trws puts it), as such we should perhaps keep it simple at first.

Here are some points to get the discussion started:

 * `flux job run` executes a parallel job, either interactively or requests execution and exits.
    It is somewhat a command with 3 phases:
  1. From user options, existing file or files, other means, etc, generate canonical jobspec
  2. Submit the signed jobspec to the instance's job ingest system (using submit/execute API)
  3. Monitor job status and I/O interactively -- this is actually dependent on the job shell being used,
      and therefore after submission `flux job run` may do something like optionally execute a front-end
      portion of the job shell, or load some extensions provided by the job shell.

For `flux job run` what we're probably most interested in the command line interface in step 1.

 * `flux job submit` submits work for later execution by the scheduler, but should take the same options and arguments as `flux job run` (if any) to generate jobspec. There are also two kinds of "jobs" users may want to submit:
  * Execution of a single parallel job (e.g. exactly what you'd run with `flux job run` but non-interactively)
  * "batch script" execution, in which case `flux job submit` will need to create a jobspec that requests a new instance with the batch script as initial program.

Because the generation of jobspec is common to `flux job submit` and `flux job run`, we may want to entertain the idea of having a separate generator command that emits jobspec from some kind of user input, which is then fed into these job commands. This could be done invisibly to the user somehow.

Sorry, these initial comments are kind of all over the place, and the existing text in the proposed RFC is almost non-existent, however, hopefully this gets our discussion started and we can make some progress here.

### DIFF
--- a/spec_17.adoc
+++ b/spec_17.adoc
@@ -1,0 +1,43 @@
+ifdef::env-github[:outfilesuffix: .adoc]
+
+17/Flux Job Submission and Execution Interface
+==============================================
+
+This specification describes the base interface for the Flux commands
+and APIs for the execution and submission of parallel jobs.
+
+* Name: github.com/flux-framework/rfc/spec_17.adoc
+* Editor: Mark Grondona <mgrondona@llnl.gov>
+* State: raw
+
+== Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to
+be interpreted as described in http://tools.ietf.org/html/rfc2119[RFC 2119].
+
+== Related Standards
+
+*  link:spec_12{outfilesuffix}[16/KVS Job Schema]
+*  link:spec_14{outfilesuffix}[14/Canonical Job Specification]
+
+== Background
+
+The commands and APIs used to submit and execute parallel jobs are fundamental
+to Flux design and offer the main interface for users to interact with
+the scheduling and resource management aspects of the system. Therefore,
+a standard, base specification of these interfaces is required to prevent
+disruptive changes to these interfaces for users.
+
+== Commands
+
+Commands for submitting, executing, and manipulating Flux jobs SHALL
+be kept under the `flux job` command. The `flux job` subcommands
+described in this RFC are the `flux job run` and `flux job submit` commands.
+
+=== flux job submit
+
+=== flux job run
+
+== APIs
+


garlick:
Thanks for getting this started!

One question: do we need separate subcommands, or just a single command with options for selecting whether to run interactively vs batch, and immediately vs later?

grondo:
> One question: do we need separate subcommands, or just a single command with options for selecting whether to run interactively vs batch, and immediately vs later?

Good question. I'd lean toward keeping them separate only because we started with a single command in Slurm and eventually had to make two commands. Subcommands are cheap and could conceivably share code if they are in the same `flux-job` command driver, so that might make development of two commands just as simple as one command with multiple options.

Running within a time period (immediate vs later) would seem to be specified as part of the jobspec and thus should be generated the same for these two commands.

grondo:
Speaking of other subcommands -- we might also want to provide an alternative to `salloc(1)` which is essentially the same as an interactive `flux job submit` with a shell for an initial program/batch script. Making it a separate command might make a slightly clearer user interface (I'm not sure though, we could also just provide a simple script to get the same effect)

We'd need pty support for this to work, or a way to `waitjob` until the job is running, fetch the URI and start `flux proxy` (requires ssh server on compute nodes)

grondo:
Just occurs to me @garlick, maybe what you were thinking of here is to have some scriptable "plumbing" commands to generate, transform, and submit jobspec, and then `flux job submit` and `flux job run` could be simple wrappers that use those plumbing commands (perhaps even short scripts)

garlick:
Plumbing and porcelain seems like a good way to subdivide this design!

SteVwonder:
I'm struggling a bit and I think it is because my brain wants to map the Flux commands back to SLURM commands, which I shouldn't do.  So instead, let me see if I can map to specific scenarios instead:

1. I'm on a management node and would like to run an application on compute nodes at some point in the future
2. I'm on a compute node and would like to run a large scale application on a different set of compute nodes at some point in the future
3. I'm on a management node and would like an interactive session ASAP
4. I'm on a compute node (either via an interactive session or a batch allocation) and would like to run a large scale application in my current allocation ASAP

Under SLURM:

1. `sbatch my_script.sh`
2. Same as 1
3. `salloc`
4. `srun my_app.exe`

Under the proposed interface, would they be:

1. `flux job submit my_script.sh`
    -  would `flux job submit my_app.exe` work?
2. Same as 1
3. Slick interface TBD, but current interface: `flux job submit` followed by `flux proxy`
4. `flux job run my_app.exe`

Is it right?  If so, what happens if I run `flux job run my_app.exe` from a management node?

trws:
I've actually always wondered, why is salloc a thing? It seems like it's srun (at least with a single target task), but with different internal flags that keep slurm from yelling at you for submitting an interactive job.  Does it actually do something meaningfully differently from a user perspective?

To Stephen's points, we need to make a decision about how we're going to distinguish some of these.  Mostly we need to distinguish whether there's an instance being run in the job or not, since all of the commands, at least by default, target their innermost enclosing instance.   So 2 wouldn't necessarily be the same as 1, and 4 would be run but probably with a flag to say "run on all resources in the enclosing instance."  In a way I think having the abstraction makes things cleaner, we don't have to overload what "run" means like slurm did, but it means we have to think about how something like your case 2 would be done when there is an instance in the job, and how something like 4 would, or even if it could, be done in one without an instance.

grondo:
> I've actually always wondered, why is salloc a thing? It seems like it's srun (at least with a single target task)

Not defending `salloc`, but it is a thing because it is a common thing users want (run a *single* interactive process with access to an allocation). It was actually part of `srun` at the beginning, but I guess was confusing enough that way that they split it out eventually (same with `sbatch`)

> Mostly we need to distinguish whether there's an instance being run in the job or not, since all of the commands, at least by default, target their innermost enclosing instance.  

@garlick, @SteVwonder and I chatted for a bit, and I probably won't be able to summarize the consensus coherently, but I *think* we decided that we at least need separate commands to differentiate running a single parallel program vs. running a "batch script" (for lack of a better term), which itself may invoke one or more parallel programs. In the second case, creation of a child instance with the batch script as its initial program is implied and codified into the generated jobspec.

We also seemed to come to the conclusion that our tools could be built from these basic methods:
(I know the naming is no good, but it is all I have ATM)

 - `submit`: submit a job via fully-specified canonical jobspec to an instance, get back a job identifier
 - `waitjob`: wait for a job to reach one or more states (e.g. "running")
 - `attach`: A function of the chosen job shell, "attach" to I/O, forward signals, process state generated by the remote job shells.

Any of our required tools can be built from these components, so they are probably candidates for the "plumbing" commands.

@SteVwonder had a nice table of simple use cases (perhaps captured above) and we should probably work through those here and capture them in the RFC.

>   So 2 wouldn't necessarily be the same as 1, and 4 would be run but probably with a flag to say "run on all resources in the enclosing instance."  In a way I think having the abstraction makes things cleaner, we don't have to overload what "run" means like slurm did, but it means we have to think about how something like your case 2 would be done when there is an instance in the job, and how something like 4 would, or even if it could, be done in one without an instance.

Thanks for bringing this up, I had forgotten about those cases. I think we had talked before about not allowing the "run on all resources in the enclosing instance" to be implied like it is with `srun app` in Slurm. I could be persuaded otherwise, but it seemed much cleaner to have a different mode for running jobs within a batch script if you want them to run on all resources currently available.

If not, then the special case could be handled by this mode only being implied if the user is the current instance owner.

SteVwonder:
So to summarize a coffee-break, whiteboard meeting that @garlick @grondo and I had, we came to the conclusion that everything boils down to some variant of `flux job submit`.

For example:
- If you want to get an allocation with a new user Flux instance and run a job as the initial program (either application or script), you could call `flux job submit`.
- If you want to directly run an application without creating a new instance and see the output of the application you submitted "live" at your terminal (a la `srun`), you could do something like `flux job submit --no-broker` followed by a `flux job attach`.
- If you want an interactive session (a la `salloc`), you could do `flux job submit` followed by `flux job attach --tty`.

To simplify things for users, we can combine these combinations of commands into a single command.  For example:

- `flux job submit --no-broker`+`flux job attach` could be wrapped up as `flux job run`
- `flux job submit`+`flux job attach -tty` could wrapped up as `flux job interactive`.

The main tables up on the whiteboard at the end of the session looked like:

|  | Management Node | Compute Node |
| ------------- | ------------- | ------------- |
| Batch Execution | `sbatch` | `sbatch` |
| Run Now / Interactive | `srun` OR `salloc`  | `srun`  |

|  | System Enclosing Instance | User Enclosing Instance |
| ------------- | ------------- | ------------- |
| Batch Execution | `flux job submit` | `flux job submit` |
| Run Now / Interactive | `flux job run` OR `flux job interactive`  | `flux job run`  |

Note: There wasn't necessarily consensus on the exact namings or combinations of commands. I took some liberties for the purpose of providing concrete examples.  I think the main consensus was:

- everything seems to boil down to `flux job submit` in the end, so we should take advantage of the composability
- for performance and simplicity reasons, there should be a way to launch a job without also launching a broker

SteVwonder:
@grondo beat me to the chase LOL.  Sorry for any duplication.

Also, tagging https://github.com/flux-framework/flux-core/issues/1570 for future reference.

grondo:
> @grondo beat me to the chase LOL. Sorry for any duplication.

Haha, no, your summary was much more complete than mine.

SteVwonder:
Something else that was mentioned in passing was the need for chaning which instance you are targeting. By default, we want to target the current enclosing instance (regardless if it is the system instance or the user instance). But what if the user wants to submit a new job to the main system instance from within a user instance.  This use-case requires some leg work on the part of the user.  They must first proxy into the parent instance and then submit the job.  It would be nice if they could do something like `flux job submit --target-instance=parent` instead, or if their user instance is 10 instances deep, they could do `flux job submit --target-instance=root`.

trws:
Yeah, having some simplifying commands or arguments for instance targeting would be good, both for enclosing cases and cases where you're running outside the instance you want to target.


That all sounds reasonable, and having different commands for running something directly VS running something that will run job steps sounds good to me.

trws:
Just to put this out there though, do we really want them to be sub-sub commands rather than say "flux run" and "flux submit?" I suppose those could be aliases or something, but the extra token strikes me as odd for some reason.

grondo:
> Just to put this out there though, do we really want them to be sub-sub commands rather than say "flux run" and "flux submit?" I suppose those could be aliases or something, but the extra token strikes me as odd for some reason.

Yeah, I was thinking a lot of people/users would have that same comment.

I guess `flux job `*`subcommand`* could be for the composable plumbing commands, and from there we could build top level `flux run` and `flux submit`. (I'm not opposed to it, I mean)

SteVwonder:
> Just to put this out there though, do we really want them to be sub-sub commands rather than say "flux run" and "flux submit?"

That's a good point.  I think it partially comes down to how many sub-sub-commands do we think `flux job` is going to have?  If the answer is something like 10, then I think sub-sub-commands make sense to avoid "polluting" the sub-command namespace.   If the answer is something like 3-5, then I think popping everything up a level makes sense.

The other piece to consider is vague commands like `cancel` and `modify`.  `flux job cancel` and `flux job modify` seem clear to me, but `flux cancel` and `flux modify` don't.  So the latter would need to be more creatively named.

:+1: to @grondo's point about selectively making some things top-level.

trws:
Good point.  Having a number of commands, like modify, cancel, attach, info, perhaps wait, as flux job * commands seems reasonable, but the main commands of run and submit for some reason feel like special cases.  Maybe they shouldn't, but they do.

SteVwonder:
Have we given any thought into what our job script headers (e.g., `#SBATCH` and `#BSUB`) will look like?   Maybe we do not need them because we have the jobspec?  We've always said that we will provide some commands that automatically generate jobspec for users, but do we also have plans to support providing some combination of the two (jobspec and CLI flags) for the same job?  I guess what I'm really trying to get at is how would a user go about expressing the commonalities between their jobs (e.g., computing account, output file destination, job name) as well as the differences between their jobs (e.g., num nodes, walltime, binary, binary arguments).  Currently users put the commonalities at the top of their job scripts via header commands (`#SBATCH`) and then provide the differences via command line arguments (which take precedence over what's in the script).

trws:
I was actually thinking about that, it would be nice to be able to use the standard yaml header setup, since we're using yaml anyway:

```
---
spec: ...
other stuff: ...
---
script down here
```

The trick would be doing that in a way that it still yields a valid shell script, not sure if there's a standard way to do that but it would make it trivial to handle in the parser if we could.  Alternately, we could accept a spec file that points out to scripts, or possibly even contains them if we want to go the travis route (not sure we do, but it's possible).

spec_17.adoc
```diff
@@ -0,0 +1,43 @@
...
+
+== Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to
+be interpreted as described in http://tools.ietf.org/html/rfc2119[RFC 2119].
+
+== Related Standards
+
+*  link:spec_12{outfilesuffix}[16/KVS Job Schema]
```
SteVwonder: Should this be `link:spec_16`?

garlick:
Maybe this could get an update to call out the plumbing vs porcelain approach, and maybe list the plumbing commands we have now:
* flux jobspec
* flux job submit
* flux job list
* flux job cancel
* flux job priority

And what we're missing
* flux job wait
* flux job attach

grondo:
I think we've moved beyond this and no longer need an RFC. Closing.

PR #134, opened 2018-08-29
RFC 18: KVS Event Log Format

### LINKED ISSUE #133: create RFC for state log format
KVS state logs are a proposed abstraction for representing a log of job state transitions in the KVS.  [RFC 16](https://github.com/flux-framework/rfc/blob/master/spec_16.adoc) proposes multiple state logs for different parts of the system (e.g. job manager, job shell, scheduler).

It was suggested that we could broaden the abstraction to support job state logs as well as other generic log types (such as a debug log).

One proposed format was a simple text format with log entries delimited by newlines, e.g.:
```
<date> SPACE <string> NEWLINE
```
Initial proposal was for `<date>` to be an [ISO 8601 date](https://en.wikipedia.org/wiki/ISO_8601) but it was subsequently questioned whether this was overkill when a numerical representation would suffice.

SteVwonder:
This RFC is a rough first cut at describing the KVS event log format used for job events.  The KVS event log is initially described in RFC16, but the event format was never defined.  This is the log format that @garlick, @grondo, and I settled on during an afternoon brainstorming session.

Closes #133

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,7 @@ SOURCES = \
 	spec_14.adoc \
 	spec_15.adoc \
 	spec_16.adoc \
+	spec_18.adoc
 
 HTML = $(SOURCES:.adoc=.html)
 PDF = $(SOURCES:.adoc=.pdf)
--- a/spec_16.adoc
+++ b/spec_16.adoc
@@ -78,6 +78,9 @@ exiting::
 completed::
   all containers exited
 
+canceled::
+  user requested the job be canceled
+
 rejected::
   scheduler won't run job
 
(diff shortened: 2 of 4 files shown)


grondo:
Thanks! I'm not sure so I'll just ask, do we also want to describe other properties of a "kvs event log" in this RFC?

I was also thinking we should refer to RFC3 for the format of a *topic string*, but that RFC defines a topic string as a `service.method` format, which doesn't apply here. For the event log purpose, we just want to specify that "*topic* is a string name of the event being logged, optionally namespaced. Maybe some wording like:

  "Where *topic* SHALL be a string denoting the name or description of the event being appended to the log. The *topic* string MAY use dot notation to indicate an event that falls under a namespace other than the default. Namespaces for *topic* strings in an event log MAY be arbitrarily nested"

garlick:
I think we agreed to rename "topic" to "event name" and to drop namespacing from the RFC.

If each entry is called an event, we can probably refer to them as such in RFC 16.

garlick:
Should we call "freeform" something like "event parameters" and make it optional?

garlick:
One small suggestion:

> `timestamp` SHALL be the floating point representation of the number of seconds
  since the epoch, with at least millisecond resolution

Maybe this would be slightly more precise if it stated:

> `timestamp` SHALL be the number of seconds since the Unix Epoch (1970-01-01 UTC), expressed with optional sub-second precision in decimal notation.

PR #139, opened 2018-10-01
rfc7: forbid typedefs for fixed length arrays

garlick:
This follows up on a discussion in flux-framework/flux-core#1668


grondo:
Thanks!

PR #140, opened 2018-10-24
rfc5: remove service_name symbol

grondo:
Remove the optional `service_name` extension module symbol. It is
no longer used by the broker to determine a secondary service
name for modules.

(Assuming PR flux-framework/flux-core#1753 goes in)


garlick:
Thanks!  Always felt this was a hacky way to do it.

PR #141, opened 2018-11-04
rfc10: change SHALL to MAY for enclosing instance

garlick:
Problem: RFC 10 requires Flux instances to use the
enclosing instance, if any, as a tertiary level of
content storage.  This feature has not been implemented
and is not strongly motivated.

Change SHALL to MAY for this feature.

This was discussed in flux-framework/flux-core#1784


grondo:
Thanks!

PR #143, opened 2019-01-02
rfc2: update evolving stance on flux project licensing

garlick:
Now that we've decided to make flux-framework LGPL 3 (and received approval), RFC 2 needs an update.


grondo:
Looks good, thanks for doing that!

PR #151, opened 2019-02-06
spec14: clarify UC1.3

### LINKED ISSUE #105: rfc14, problems with "at least" wording in examples
There is a problem with the wording "at least" in some of the rfc14 jobspec examples.  Here is one such case:

> Request 4 nodes with at least 2 sockets each, and 4 cores per socket

Note that in the jobspec, we are actually requesting _exactly_ 2 sockets:

```
    version: 1
    resources:
      - type: slot
        count: 4
        label: nodelevel
        with:
        - type: node
          count: 1
          with:
            - type: socket
              count: 2
              with:
                - type: core
                  count: 4
    tasks:
      - command: [ "flux", "start" ]
        slot: nodelevel
        count:
          per_slot: 1
    attributes:
      system:
        duration: 1 hour
 ```

The scheduler is expected, I think, to allocate whole nodes to each task.  Except that we have asked for nodes with _exactly_ 2 sockets, and each socket must have _exactly_ 4 cores.

So we can improve the example by changing the socket count section to, for instance, the following:

```
  - type: socket
    count:
        min: 2
        max: 10000000000
        operator: "+"
        operand: 1
```

But it is a bit clunky to make people guess a big enough number for "max".  We could perhaps introduce words like "INFINITY":

```
  - type: socket
    count:
        min: 2
        max: INFINITY
        operator: "+"
        operand: 1
```

But it would probably be nicer still to make field optional, so we can specify it like this:

```
  - type: socket
    count:
        min: 2
```

### PR DESCRIPTION
**Pull request by grondo**
This PR is meant to address the main issue brought up in #105, which I think
everyone agreed was valid and should be fixed in the RFC.

The description of Use Case 1.3 is worded with "at least", but the actual
example jobspec specified the exact count of sockets and cores. Update the
example to use a `min` count instead with no `max` (implying "at least").

Adjust the wording for the task `count` key to allow specification of `min`
without `max`, to make the new UC1.3 adhere to the spec.

Closes #105

### DIFF
--- a/data/spec_14/use_case_1.3.yaml
+++ b/data/spec_14/use_case_1.3.yaml
@@ -8,10 +8,12 @@ resources:
       count: 1
       with:
         - type: socket
-          count: 2
+          count:
+            min: 2
           with:
             - type: core
-              count: 4
+              count:
+                min: 4
 tasks:
   - command: [ "flux", "start" ]
     slot: nodelevel
--- a/data/spec_14/use_case_1.3.yaml
+++ b/data/spec_14/use_case_1.3.yaml
@@ -5,6 +5,7 @@ resources:
     label: nodelevel
     with:
     - type: node
+      exclusive: false
       count: 1
       with:
         - type: socket
--- a/data/spec_14/use_case_2.5.yaml
+++ b/data/spec_14/use_case_2.5.yaml
@@ -5,7 +5,8 @@ resources:
     count: 10
     with:
     - type: memory
-      count: 2
+      count:
+        min: 2
       unit: GB
     - type: core
       count: 1
--- a/data/spec_14/use_case_2.6.yaml
+++ b/data/spec_14/use_case_2.6.yaml
@@ -8,7 +8,8 @@ resources:
       count: 1
       with:
         - type: memory
-          count: 4
+          count:
+            min: 4
           unit: GB
 tasks:
   - command: app
--- a/spec_14.adoc
+++ b/spec_14.adoc
@@ -132,16 +132,24 @@ A resource vertex SHALL contain the following keys:
  of two possible values: either a single integer value representing
  a fixed count, or a dictionary which SHALL contain the following keys:
 +
-[horizontal]
+[horizontal width=20%]
    *min*::: The minimum required count or amount of this resource
 
++
+and additionally MAY contain the following keys:
++
+[horizontal width=30%]
    *max*::: The maximum required count or amount of this resource
 
    *operator*::: An operator applied between `min` and `max` which
    returns the next acceptable value
 
    *operand*::: The operand used in conjunction with `operator`
-   to get the next value between `min` and `max`.
+
++
+The default value for `max` SHALL be  _infinite_, therefore a `count`
+which specifies only the `min` key SHALL be considered a request for
+_at least_ that number of a resource.
 
 A resource vertex MAY additionally contain one or more of the
 following keys

### CODE REVIEW

**Comment by dongahn** [comment]
This looks great me. @grondo: do you want to add the semantics of the spec as to when the plain `count` key is used, instead of its sub-keys in this PR?

**Comment by SteVwonder** [comment]
Thanks @grondo!

One thing I noticed, is that (before this PR) UC1.3 and UC1.4 have the exact same jobspec but different text describing them:
```
→ diff use_case_1.3.yaml use_case_1.4.yaml && echo "Files do not differ"
Files do not differ
```
I think UC1.3 is meant to show the node not being exclusively allocated.  Do you mind adding `exclusive: false` under `node` to UC1.3?

Also, if you don't mind, UC2.5 and UC2.6 both have the same wording of `at least` when they should `exactly` (or something similar).  Can you switch from `at least` to `exactly` or change the `count` to use `min`?  Either is fine with me.

**Comment by grondo** [comment]
> do you want to add the semantics of the spec as to when the plain count key is used, instead of its sub-keys in this PR?

I'd be happy to -- though I'm not sure exactly what you want. Currently the description states:

> The `count` SHALL have one of two possible values: either a single integer value representing a fixed count, or a dictionary which SHALL contain the following keys

Would you like the case of the "single integer value representing a fixed count" expanded on, or did you have something else in mind?

**Comment by grondo** [comment]
>I think UC1.3 is meant to show the node not being exclusively allocated. Do you mind adding exclusive: false under node to UC1.3?

Yes, great catch!

> Also, if you don't mind, UC2.5 and UC2.6 both have the same wording of at least when they should exactly (or something similar). Can you switch from at least to exactly or change the count to use min? Either is fine with me.

Ok, will do.

**Comment by dongahn** [comment]
> Would you like the case of the "single integer value representing a fixed count" expanded on, or did you have something else in mind?

Yes, and that plus, what this means to the scheduler in matching and generating the resource set info. If you don't feel comfortable to write this up, I can do this while I'm working on the RFC for R since we will have to link those two RFCs as well.

**Comment by grondo** [comment]
> Yes, and that plus, what this means to the scheduler in matching and generating the resource set info

Ah, a sort of *caveat utilitor* clause for the exact `count` specification? That definitely makes sense to be in here, but I'd rather go over that in a separate PR.

**Comment by grondo** [comment]
Oops I didn't get UC1.3 right. I replaced `count` with `min` instead of adding `min` as a key under `count`. Fix coming up.

**Comment by grondo** [comment]
Ok, fixed, squashed, and forced a push.

>  Can you switch from at least to exactly or change the count to use min? Either is fine with me.

The *at least* in these use cases specified an amount of RAM. I didn't think it would make sense (ever?) to request an *exact* amount of RAM, so I changed the examples to use `min` and kept the wording the same.

**Comment by garlick** [comment]
Is this OK to merge?

Note as discussed offline, the example changes need to be propagated to flux-core's jobspec valid input in `t/jobspec/valid`.  I think the jobspec JSON schema also will need updating to allow the dictionary range without `max`.  I'm happy to take care of both but we should get an issue open over there when this is ready to merge.

**Comment by grondo** [comment]
I had a working branch that updated the valid inputs and was going to ask you about updating the schema.

It did feel like there should be an option to test against the "upstream" rfc automatically or otherwise keep the valid input synchronized without having to do it manually, but I suppose that would make it more difficult to propose a new revision of jobspec without affecting tests in flux-core.

**Comment by garlick** [comment]
Right sounds like there is consensus here so merging.  New PR if anything needs to change.

**Comment by SteVwonder** [comment]
@grondo: which version of asciidoc do you have installed?  I'm running into issues with the `[horizontal width=20%]`:

```
→ make
asciidoc -o spec_14.html spec_14.adoc
asciidoc: ERROR: spec_14.adoc: line 135: illegal style name: horizontal width=20%
asciidoc: ERROR: spec_14.adoc: line 141: illegal style name: horizontal width=30%
make: *** [spec_14.html] Error 1

→ asciidoc --version
asciidoc 8.6.8
```

grondo:
I don't install asciidoc locally and instead use the GitHub rendering to check rendering. Sorry about that! I guess we should probably add asciidoc to travis-ci to make sure we catch these mistakes before merging. (and now that I look back at this I clearly didn't know what I was doing and GitHub has just fallen back to a normal definition list, which is probably what I wanted anyway in this case)

SteVwonder:
Ok.  I'll remove the `width=..` from the horizontal in #154

PR #153, opened 2019-02-07
rfc19: document FLUID generation

garlick:
This is a first stab at converting @grondo's proposal for FLUIDs (now in use for job ID's) from flux-framework/flux-core#470 into an RFC.


grondo:
Nicely done! (Though putting me as Author is going a bit far). I'll go ahead and merge as I didn't see anything of comment.

PR #154, opened 2019-02-08
RFC14: Add `per_resource` option to task's `count`

### LINKED ISSUE #150: Apparently simple resource request not possible in canonical jobspec?
@grondo's original suggestion for a feature to `flux-jobspec`:
> ... isn't the key difference between `flux jobspec srun` and `sbatch` that `sbatch` should insert a `flux start` or `flux broker` before the provided `COMMAND` (which is assumed to be a batch script), and ensure that the tasks run once per node? (or something similar

@SteVwonder's reply:
> ... To make that last piece about one per node work, I think `--nodes` would have to be a required argument. If it is optional and the user only specifies `--cores`, then the tool will have to make a determination/assumption about the number of cores per node......actually, now that I think about this, is it even possible to specify that kind of request in the canonical jobspec? The request of thinking of is along the lines of: "I want N cores. I don't care about the number of nodes, but I want 1 process per node."

@grondo's reply:
> It does feel like we got ourselves into a bind by requiring that the `slot` shape be defined as part of the resource request. There are certain things that are now not possible or very difficult to express (though it could just be my misunderstanding of current jobspec definition). In general, I don't see how you can define a slot at a parent of the resource type you want to request without affect the exact count (and exclusivity) of the parent resource type (if that makes any sense, specific example is what we're discussing here where you want N cores with a slot of a node, but don't care how many nodes. In fact, the exact slot shape can only be determined from a _concrete_ resource set)

> A couple ideas on how to handle this come to mind:
> * add a concept of predefined slots that match resource names, e.g. `node`, `socket`, `core` that allow the slot to be left out of the jobspec resources section, e.g. in the case of `flux jobspec sbatch` you could emit a resources section with no slot and a `task:` with `slot:` label `node:`
> * add a special value for `count:` like `any` or `0` that indicates no preference for the number of this type, and any slot that contains an `any` resource as a direct child would by definition apply to exactly 1 of that resource.

### LINKED ISSUE #156: RFC14: make fails on quartz
```
asciidoc -o spec_14.html spec_14.adoc
asciidoc: ERROR: spec_14.adoc: line 135: illegal style name: horizontal width=20%
asciidoc: ERROR: spec_14.adoc: line 141: illegal style name: horizontal width=30%
source-highlight: missing feature: language inference requires input file
source-highlight: could not find a language definition for yaml
asciidoc: WARNING: example1.yaml: line 19: filter non-zero exit code: source-highlight -f xhtml -s yaml: returned 1
asciidoc: WARNING: example1.yaml: line 19: no output from filter: source-highlight -f xhtml
```

SteVwonder:
Adds `per_resource` option to a task's `count` key.

I noticed that at the top of the first section of examples (i.e., resource only requests), it states:
> In terms of a canonical jobspec, these requests are assumed to be requests to start an instance, i.e. run a single copy of flux start per allocated node.

But then use-cases 1.6 and 1.7 (1.6 and 1.8 in this PR), allocate and run `flux start` on a per-core basis.  Do we want to move those use-cases out of the `resource only requests` section of examples and into the 2nd section of examples?

Closes #150 
Closes #156

### DIFF
--- a/spec_14.adoc
+++ b/spec_14.adoc
@@ -221,19 +221,28 @@ descriptor SHALL contain the following keys:
  number of tasks executed per task slot SHALL be a function of the
  number of resource slots and total number of tasks requested to execute.
 
- *count*::
- The value of the `count` key SHALL be a dictionary supporting at least
- the keys `per_slot` and `total`, with other keys reserved for future
- or site-specific extensions.
-+
-[horizontal]
+ *count*:: The value of the `count` key SHALL be a dictionary supporting at
+ least the keys `per_slot`, `per_resource`, and `total`, with other keys
+ reserved for future or site-specific extensions.
+
   *per_slot*:::
   The value of `per_slot` SHALL be a number indicating the number
   of tasks to execute per task slot allocated to the program.
 
+  *per_resource*::: The value of `per_resource` SHALL be a dictionary which
+  SHALL contain the following keys:
+
+  * *type* The value of the `type` key SHALL be a resource type explicitly
+   declared in the associated task's slot.
+
+  * *count* The value of the `count` key SHALL be a number indicating the number
+     of tasks to execute per resource of type `type` occurring in the task's
+     slot.
+
   *total*:::
   The value of the `total` field SHALL indicate the total number of
   tasks to be run across all task slots, possibly oversubscribed.
+
 
  *attributes*::
  The `attributes` key SHALL be a free-form dictionary of keys which may
--- a/spec_14.adoc
+++ b/spec_14.adoc
@@ -133,13 +133,13 @@ A resource vertex SHALL contain the following keys:
  of two possible values: either a single integer value representing
  a fixed count, or a dictionary which SHALL contain the following keys:
 +
-[horizontal width=20%]
+[horizontal]
    *min*::: The minimum required count or amount of this resource
 
 +
 and additionally MAY contain the following keys:
 +
-[horizontal width=30%]
+[horizontal]
    *max*::: The maximum required count or amount of this resource
 
    *operator*::: An operator applied between `min` and `max` which
--- a/spec_14.adoc
+++ b/spec_14.adoc
@@ -225,6 +225,7 @@ descriptor SHALL contain the following keys:
  least the keys `per_slot`, `per_resource`, and `total`, with other keys
  reserved for future or site-specific extensions.
 
+[horizontal]
   *per_slot*:::
   The value of `per_slot` SHALL be a number indicating the number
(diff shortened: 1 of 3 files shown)


grondo:
> But then use-cases 1.6 and 1.7 (1.6 and 1.8 in this PR), allocate and run flux start on a per-core basis. Do we want to move those use-cases out of the resource only requests section of examples and into the 2nd section of examples?

Sorry for the delay, I thought I had already replied here.

We should think about whether the "resource only" requests description make much sense anymore. The examples launch brokers, so that term is perhaps confusing now?

 - UC1.6 runs one flux-broker per core, but there is also only one core per node. If the example is trying to demonstrate running a broker per core, it would be more enlightening perhaps to make the request for >1 core per node.

 - UC1.8 is similar -- 1 broker is launched per core only because that is the maximum amount of cores we could have available on a host.

I'm not sure there needs to be an example of running an instance with more than one broker per host/node, because that style should not be encouraged (at least I can't think of a reason). Instead we should maybe reword or rework the examples to show how starting an instance with at most 1 broker per node can be done with any resource request by use of your new `per_resource: { "type": "node", "count": 1 }` syntax.

SteVwonder:
>  Instead we should maybe reword or rework the examples to show how starting an instance with at most 1 broker per node can be done with any resource request by use of your new `per_resource: { "type": "node", "count": 1 }` syntax.

:+1: Agreed.  I dropped the new Use Case 1.7 which I added mainly to show a usage of `per_resource` and instead modified the pre-exising Use Cases 1.6 and 1.7 to only launch one broker per node and thus use `per_resource`.

> We should think about whether the "resource only" requests description make much sense anymore.

I changed the section titles to "Node-level Requests" and "General Requests".  I'm not in love with the latter name, but it works.  If someone has a better name (for either), I'm open to suggestions.

spec_14.adoc
```diff
@@ -221,20 +221,29 @@ descriptor SHALL contain the following keys:
  number of tasks executed per task slot SHALL be a function of the
  number of resource slots and total number of tasks requested to execute.
 
- *count*::
- The value of the `count` key SHALL be a dictionary supporting at least
- the keys `per_slot` and `total`, with other keys reserved for future
- or site-specific extensions.
-+
-[horizontal]
```
SteVwonder: Just an FYI, I removed this `[horizontal]` because it made the `type` and `count` sub-list indent really far.  I thought it looked odd.

grondo:
> I changed the section titles to "Node-level Requests" and "General Requests". I'm not in love with the latter name, but it works. If someone has a better name (for either), I'm open to suggestions.

This works for me. Thanks!

Issue #109, opened 2017-10-03
Need specification for "resource set", R


grondo:
This issue is being opened to start a discussion on the use cases, API, and/or specification for the *_R_* as in RFC 15. *R* is the serialized version of any resource set, and is presumably produced by the serializer described in [RFC4](https://github.com/flux-framework/rfc/blob/master/spec_4.adoc#graph), consumed by the resource service in an instance as configuration, and used by the IMP and job shell to determine shape of containment and local resource slots.

In essence, the *R* format will be the way composite resource and resource configuration information will be transmitted to and from instances of Flux.

Ideally, the purpose of this issue is to determine the format of *R* such that a new RFC could be drafted.

To get the discussion started, here are some high level requirements and use cases for *R*:

 - *R* should act as resource configuration input to an instance,
   therefore it may be that configuration of even the system instance
   is written in *R* spec, or the configuration language (RDL?) generates *R*.
   (in fact, one use case might be to directly generate R from hwloc data)

 - Execution service in an instance needs to be able to generate
   *R<sub>local<sub>* from *R* for each rank. So given a rank or even generic
   "resource vertex", there should be a function to generate an
    *R<sub>n<sub>* from *R*, where *R<sub>n<sub>* is a hierarchical subset of *R*.

 - The containment plugins in the IMP will need to query *R<sub>local<sub>*
   for the list of local resources of given type or types on which
   the containment plugins operate. For instance, a memory plugin
   will need to determine the amount and location of RAM contained
   in *R<sub>local<sub>* in order to set up memcg limits. Similarly a Socket/CPU
   plugin would need to iterate over or query the list of *local*
   sockets/cores in *R<sub>local<local>* to add these to the cgroup.

 - The job shell will use jobspec+*R* to determine the local
   'task slots' that map to commands in the 'tasks' section.

Dependency management here might get challenging. The IMP is a user of *R<sub>n<sub>*, but we want to ideally eliminate dependencies in the flux-security project on other flux-framework projects. Possible approaches here might include:
1. The IMP could take a subset of the *R* specification, simple enough to parse with its own parser and offer some simplified interface to IMP plugins that do containment. Containment plugins probably only really need to get a list of local resources from *R*, as long as the logical IDs match the actual system logical IDs. The IMP's interface to R could later be expanded to offer higher level functionality for advanced containment (though I don't have any use cases that I can think of here)
[...]

PR #155, opened 2019-02-09
RFC20: Add Resource Set Specification RFC

dongahn:
This PR adds a new RFC that specifies the format 1 of resource set specification. This is discussed and agreed at https://github.com/flux-framework/rfc/issues/109#issuecomment-460469688.

### DIFF
--- a/spec_20.adoc
+++ b/spec_20.adoc
@@ -1,0 +1,168 @@
+ifdef::env-github[:outfilesuffix: .adoc]
+
+20/Resource Set Specification Version 1
+=======================================
+
+This specification defines the version 1 format of the resource-set
+representation or _R_ in short.
+
+* Name: github.com/flux-framework/rfc/spec_20.adoc
+* Editor: Dong H. Ahn <ahn1@llnl.gov>
+* State: Raw
+
+== Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL"
+in this document are to be interpreted as described in RFC 2119.
+
+== Related Standards
+
+* link:spec_4{outfilesuffix}[4/Flux Resource Model]
+* link:spec_8{outfilesuffix}[8/Flux Task and Program Execution Services]
+* link:spec_14{outfilesuffix}[14/Canonical Job Specification]
+* link:spec_15{outfilesuffix}[15/Independent Minister of Privilege for Flux]
+* link:spec_16{outfilesuffix}[16/KVS Job Schema]
+
+== Overview
+
+Flexible resource representation is important for some of the key
+components of Flux.
+Resource requests are part of Flux jobspec, described in RFC 14.
+This RFC describes the format of a concrete resource-set representation
+referred to as _R_, constructed by the scheduler in response
+to a resource request.
+_R_ is input to the remote execution system, which uses information
+expressed in _R_ to establish containment, binding, mapping,
+and execution of program tasks, apportioned across broker ranks.
+As a program terminates, the execution system releases
+shards of the original _R_, eventually
+adding up to its union, back to the scheduler.
+Finally, when a Flux instance launches a child instance,
+_R_ is passed down from the enclosing instance to the child instance,
+where it primes the child scheduler with a block of allocatable resources.
+
+== Design Goals
+
+The _R_ format is designed with the following goals:
+
+* Allow the resource data conformant to our resource model (RFC 4)
+  to be serialized and deserialized with no data loss;
+* Express the resource allocation information to the program execution
+  service (RFC 8);
+* Use the same format to release a resource subset of _R_ to the scheduler;
+* Allow the consumers of _R_ to deserialize an _R_ object while minimizing
+  the parsing complexity and the data to read;
+
+== Producers and Consumers
+
+* The scheduler for a Flux instance (or instance scheduler) uses
+  this format to serialize each resource allocation
+  as REQUIRED by the instance program execution service and OPTIONALLY
+  REQUIRED by child scheduler instances.
+* The instance scheduler deserializes an _R_ object to build
+  its internal resource data used for scheduling.
+* Users MAY manually write an _R_ object for testing and debugging.
+* User-facing utilities that query a resource status (e.g., what
+  resources are available or idle, or what resources are allocated to a job)
+  MAY use an _R_ object to extract this information;
+* The program execution service emits a valid _R_ object to release
+  a resource subset of an _R_ to the instance scheduler.
+
+== Resource Set Format Definition
+The JSON documents that conform to the _R_ format SHALL be referred
+to as _R_ JSON documents or in short _R_ documents.
+An _R_ JSON document SHALL consist of a dictionary with three
+keys: `version`, `execution` and `scheduling`. It SHALL be valid if and only
+if it contains the `version` key and either or both the `execution`
+and `scheduling` keys. The value of the `execution` key SHALL contain
+sufficient data for the execution system to perform its
+core tasks. The value of `scheduling` SHALL contain sufficient data
+for schedulers.
+
+=== Version
+
+The value of the `version` key SHALL contain 1 to indicate
+the format version.
+
+=== Execution
+
+The value of the `execution` key SHALL contain exactly one key
+with other keys reserved for future extensions: `R_lite`. `R_lite`
+is a strict list of dictionaries each of which SHALL contain three keys:
+
+  *node*:: The value of the `node` key SHALL be the string hostname
+   of the target compute node.
+
+  *rank*:: The value of the `rank` key SHALL be the integer broker rank that
+  manages the target compute node. This SHALL disambiguate the case where multiple
+  brokers manage a same node.
+
+  *children*:: The `children` key encodes the information about certain compute resources
+   contained within this compute node. The value of this key SHALL contain a dictionary
+   with two keys: `core` and `gpu`. Other keys are reserved for future
+   extensions.
+
+[horizontal]
+    *core*::: The `core` key contains a logical compute core IDs string.
+     It SHALL be a comma-separated list of core IDs (e.g., 1,3,4,5) or
+     hyphenated ranges for consecutive integer IDs (e.g., 1,3-5).
+    *gpu*::: This OPTIONAL key contains a logical GPU IDs string.
+     It SHALL be a comma-separated list of gpu IDs or hyphenated ranges
+     for consecutive integer IDs.
+
+=== Scheduling
+
+The `scheduling` key allows RFC4-compliant schedulers to serialize any subset
+of graph resource data into its value and later deserialize this value with
+no data loss. The `scheduling` key contains a dictionary with a single key: `graph`.
+Other keys are reserved for future extensions.
+The `graph` key SHALL conform to the latest version of the JSON Graph Format (JGF).
+Thus, its value is a dictionary with two keys, `nodes` and `edges`,
+that encode the resource vertices and edges as described in RFC 4.
+
+==== Graph Vertices
+
+The value of the `nodes` key defined in JGF is a strict list
+of graph vertices. Each list member is a vertex that contains
+two keys: `id` and `metadata`.
+The `id` key SHALL contain a unique string ID for the containing vertex.
+The value of the `metadata` key is a dictionary that encodes
+the resource pool data described in RFC 4.
+Thus, this dictionary SHALL contain the following
+keys to describe the base data of a resource pool:
+
+* `type`
+* `uuid`
+* `basename`
+* `name`
+* `id`
+* `properties`
+* `size`
+* `unit`
+
+It MAY contain other OPTIONAL resource vertex data.
+
+==== Graph Edges
+The value of the `edges` key defined in JGF SHALL be a strict list of graph edges.
+Each list element SHALL be an edge that connects two graph vertices and
+contains the `source`, `target` and `metadata` keys.
+The value of the `source` key SHALL contain the ID of the source graph vertex.
+The value of the `target` key SHALL contain the ID of the target graph vertex.
+The value of this `metadata` key SHALL contain a dictionary that encodes
+the resource subsystem and relationship data for the containing edge
+as described in RFC 4. It SHALL contain two keys:
+
+  *subsystem*::
+   The value of the `subsystem` key SHALL be a string that indicates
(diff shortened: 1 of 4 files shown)


dongahn:
As discussed at https://github.com/flux-framework/rfc/issues/109#issuecomment-460469688 as well as our last Thu meeting, I created a RFC20 draft for `R`. 

I have only added very basic keys for both `R_lite` as well as `scheduling`. This will further have to be optimized later. But I didn't want to add any premature optimization at this point.

@grondo, @SteVwonder, @garlick and @trws, let me know what you think.

dongahn:
CI failed because I didn't created a link to RFC20 in the top-level README.adoc.

dongahn:
OK. Forced a push. Hopefully this will make Travis happy.

spec_20.adoc
```diff
@@ -0,0 +1,156 @@
...
+* The program execution service emits a valid `R` document to release
+  a resource subset of an `R` to the instance scheduler.
+
+=== Resource Set Format Definition
+
+An `R` JSON document SHALL consist of a dictionary with three
+keys: `version`, `execution` and `scheduling`. `R` SHALL be valid if and only
+if it contains the `version` key and either or both the `execution`
+and `scheduling` keys. The value of the `execution` key SHALL contain
+sufficient data for the execution subsystem CAN perform its
```
SteVwonder: `CAN` isn't one of the formally defined terms in RFC 2119.  So I don't think it should be in all caps here.  I think `to perform` would also read a bit better.

spec_20.adoc
```diff
@@ -0,0 +1,156 @@
...
+* The scheduler for a Flux instance (or instance scheduler) uses
+  this format to serialize each resource allocation
+  as REQUIRED by the instance program execution service and OPTIONALLY
+  REQUIRED by child scheduler instances.
+* The instance scheduler deserializes an `R` document to build
+  its internal resource data used for scheduling.
+* Users MAY manually write an `R` document for testing and debugging.
+* User-facing utilities that query a resource status (e.g., what
+  resources are available or idle, or what resources are allocated to a job)
+  CAN use an `R` document to extract this information;
```
SteVwonder: `CAN` -> `MAY`?

spec_20.adoc
```diff
@@ -0,0 +1,156 @@
...
+for schedulers.
+
+==== Version
+
+The value of the `version` key SHALL contain an non-negative integer that
+indicates the format version.
+
+==== Execution
+
+The value of the `execution` key SHALL contain exactly one key: `R_lite`. `R_lite`
```
SteVwonder: Borrowing from RFC 14, it might be useful to add some language here along the lines of: `with other keys reserved for future extensions` (if that, indeed, is the intent).

spell.en.pws
```diff
@@ -384,3 +384,11 @@ FLUIDs
...
 eventlog
 timelimit
+Ahn
+Bargnesi
+conformant
+deserialize
+deserialized
+deserializes
+JGF
+medadata
```
SteVwonder: This shouldn't be added to the spell list.  `medadata` should actually be `metadata`

spec_20.adoc
```diff
@@ -0,0 +1,156 @@
...
+
+  *children*:: The `children` key encodes the information about certain compute resources
+   contained within this compute node. The value of this key SHALL contain a dictionary
+   with exactly one key, `core`.
+
+[horizontal]
+    *core*::: The `core` key contains a logical compute core IDs string.
+     It SHALL be a comma-separated list of core IDs (e.g., 1,3,4,5) or
+     hyphenated ranges for consecutive integer IDs (e.g., 1,3-5).
+
```
grondo: Do we want to add OPTIONAL `gpu` key here as well since current `R_lite` supports it?

spec_20.adoc
grondo: :+1:

spec_20.adoc
```diff
@@ -0,0 +1,156 @@
...
+=== Resource Set Format Definition
+
+An `R` JSON document SHALL consist of a dictionary with three
+keys: `version`, `execution` and `scheduling`. `R` SHALL be valid if and only
+if it contains the `version` key and either or both the `execution`
+and `scheduling` keys. The value of the `execution` key SHALL contain
+sufficient data for the execution subsystem CAN perform its
+core tasks. The value of `scheduling` SHALL contain sufficient data
+for schedulers.
+
```
grondo: The `execution` or similar key should be required not optional (this will be the part of `R` used by flux-core). The `scheduling` key should be OPTIONAL as a key for scheduler-specific data.

In fact, maybe we should rename `execution` to `core`? (Just a thought)

spec_20.adoc
```diff
@@ -0,0 +1,156 @@
...
+==== Scheduling
+
+The `scheduling` key allows RFC4-compliant schedulers to serialize any subset
+of graph resource data into its value and later deserialize this value with
+no data loss. The `scheduling` key contains a dictionary with a single key: `graph`.
+The `graph` key SHALL conform to the latest version of the JSON Graph Format (JGF).
+Thus, its value is a dictionary with two keys, `nodes` and `edges`,
+that encode the resource vertices and edges as described in RFC 4.
+
```
grondo: In this version of `R`, the purpose of the `scheduling` dict was to an area for advanced schedulers to add data opaque to flux-core. So, the dict should really allow any keys, with `graph` reserved for your JGF spec here. (However, it seems unlikely there will be any other schedulers using this spec for now, so I'm fine with the current language)

spec_20.adoc
dongahn: Agreed.

spec_20.adoc
dongahn: I didn't realize this. Thanks.

spell.en.pws
dongahn: Opps.

spec_20.adoc
dongahn: Thanks.

Re: optional.

One counter example would be when the execution system releases a resource subset to the scheduler, this key shouldn't be necessary in such cases?

The reason I used `execution` was for a potential case where this section is extended like `R_lite++`.

But this should really be adjusted per your development trajectory though. I was thinking the execution system will evolve before being able to fully adapt to the API approach we agreed to do for the future. 

I may be wrong though.

spec_20.adoc
dongahn: Good thought. Agreed.

spec_20.adoc
dongahn: True. But thinking was to make this spec beyound flux-core centric. 

In general there will be thr base scheduler data per RFC4 that needs to be shoved into the graph, I though it would be good specify them.

spec_20.adoc
grondo: Good thoughts. I hadn't thought about how the execution service could create an R subset when it also contains a `scheduling` dictionary. Since that is supposed to be opaque I think it either has to leave that off and require the scheduler to be able to determine which exact resources are being released from R_lite only, or leave the entire `scheduling` section alone and pass it through.

Long term I think you and @trws are correct that it is better to use the one-section approach, but that change can come after this stop-gap spec.

For now can we kick the can down the road and assume the only resource subset than can be released is "by rank"?

Pragmatically, the reason `scheduling` should be optional is that when testing in `flux-core` the dummy scheduler will not use this section, and we still want to create an `R` that adheres to the spec.

spec_20.adoc
grondo: Oh, yeah, sorry I think I wasn't clear. I only meant to change something like:

> The scheduling key contains a dictionary with a single key: graph

to allow other keys besides `graph`

I'm fine with specifying the `graph` key in this RFC.

spec_20.adoc
dongahn: Thanks.

> For now can we kick the can down the road and assume the only resource subset than can be released is "by rank"?

Yeah this seems right. Why don't we target this for now and see if we need to adjust our spec as we continue to develop the functionality...

> Pragmatically, the reason scheduling should be optional is that when testing in flux-core the dummy scheduler will not use this section, and we still want to create an R that adheres to the spec.

Agreed. 

BTW, my reason for making `execution` optional was the same. I wanted to create an `R` document for scheduler tests like the new reader with this key adsence.

In general, does the current spec meet your requirement or do you need to make the execution key still mandatory?

spec_20.adoc
dongahn: Ah yes. I think that more defined keys should be added there at that level later. In addition other arbitrary should be allowed.

For eample I will probably need the `properties` key to have a properties map for edges. 

I just wanted to keep things simple and grow the spec as we develop this within our resource matching service.

For now, yes I think it makes sense to change 

> The scheduling key contains a dictionary with a single key: graph

to allow other arbitrary keys at this level.

spec_20.adoc
```diff
@@ -0,0 +1,156 @@
...
+Some of the key components of Flux MUST be able to produce and/or
+consume the data about
+a set of resources as a fundamental scheduling and execution unit.
+For instance, a scheduler MUST be able to express a resource set
+to the remote execution subsystem once an allocation is created.
+Using this data,
+the execution subsystem then MUST be able to extract a per-rank,
+local resource set for every flux-broker rank in order to perform
+its work such as the containment, binding, mapping, and execution
+of program processes.
```
garlick: Suggestion:  IMHO the opening paragraph/section should provide human digestible background, and the precise RFC 2119 language should be reserved for requirements/goals and the format definition.  I find these RFCs easier to digest when there is some expository writing at the top, and some really precise "programmer instructions" further in, and depending on which hat one has on, one or the other can be somewhat glossed over.

For the intro, maybe something like:
> Flexible resource representation is an important Flux value proposition.  Resource _requests_ are part of Flux jobspec, described in RFC 16.  This RFC describes the format of concrete resource sets, referred to as _R_, constructed by the scheduler in response to a resource request.
>
> _R_ is input to the core execution system, which uses information expressed in _R_ to establish containment, binding, mapping, and execution of program tasks, apportioned across broker ranks.  As a parallel program terminates, the execution system releases shards of the original _R_, eventually adding up to its union, back to the scheduler.
>
> Finally, when Flux launches Flux, _R_ is passed down from enclosing to child Flux instance, where it primes the child scheduler with a block of allocatable resources.

spec_20.adoc
grondo: Ah I see, I think it is fine if the `execution` key is optional for now, with the caveat that without `execution` key the `R` can't create a valid Flux job.

spec_20.adoc
```diff
@@ -0,0 +1,156 @@
...
+keys: `version`, `execution` and `scheduling`. `R` SHALL be valid if and only
+if it contains the `version` key and either or both the `execution`
+and `scheduling` keys. The value of the `execution` key SHALL contain
+sufficient data for the execution subsystem CAN perform its
+core tasks. The value of `scheduling` SHALL contain sufficient data
+for schedulers.
+
+==== Version
+
+The value of the `version` key SHALL contain an non-negative integer that
```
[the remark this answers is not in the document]
grondo: :+1:  Good catch!

spec_20.adoc
dongahn: This sounds good to. I will incorporate this.

spec_20.adoc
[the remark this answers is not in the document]
dongahn: Yes. Thanks.

spec_20.adoc
```diff
@@ -0,0 +1,156 @@
...
+* The instance scheduler deserializes an `R` document to build
+  its internal resource data used for scheduling.
+* Users MAY manually write an `R` document for testing and debugging.
+* User-facing utilities that query a resource status (e.g., what
+  resources are available or idle, or what resources are allocated to a job)
+  CAN use an `R` document to extract this information;
+* The program execution service emits a valid `R` document to release
+  a resource subset of an `R` to the instance scheduler.
+
+=== Resource Set Format Definition
```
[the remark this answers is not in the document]
dongahn: Good catch. Unintended. Will add a section.

spec_20.adoc
garlick: Oops:  `s/RFC 16/RFC 14/`

dongahn:
@SteVwonder, @grondo, @garlick: Okay, I believe I incorporate all of your comments. Please let me know if you want me to address more issues.

spec_20.adoc
```diff
@@ -26,62 +26,69 @@ in this document are to be interpreted as described in RFC 2119.
...
+to a resource request.
+_R_ is input to the remote execution system, which uses information
+expressed in _R_ to establish containment, binding, mapping,
+and execution of program tasks, apportioned across broker ranks.
+As a program terminates, the execution system releases
+shards of the original _R_, eventually
+adding up to its union, back to the scheduler.
+Finally, when a Flux instance launches a child instance,
+_R_ is passed down from the enclosing to the child, where it primes
+the child scheduler with a block of allocatable resources.
```
garlick: Probably my bad but "passed down from the enclosing to the child" should probably read "passed down from the enclosing instance to the child instance".

spec_20.adoc
```diff
@@ -26,62 +26,69 @@ in this document are to be interpreted as described in RFC 2119.
...
+
+== Design Goal
```
garlick: Goal should be plural.

spec_20.adoc
```diff
@@ -26,62 +26,69 @@ in this document are to be interpreted as described in RFC 2119.
...
   resources are available or idle, or what resources are allocated to a job)
-  CAN use an `R` document to extract this information;
-* The program execution service emits a valid `R` document to release
-  a resource subset of an `R` to the instance scheduler.
-
-=== Resource Set Format Definition
-
-An `R` JSON document SHALL consist of a dictionary with three
-keys: `version`, `execution` and `scheduling`. `R` SHALL be valid if and only
+  MAY use an _R_ document to extract this information;
```
garlick: Is _R_ a "document"?  Maybe "object" is a better word choice.

dongahn: Thanks. Since what an _R_ document is is defined later, I will use "object" here. Glad you mentioned this, though, as I struggled a bit with what _R_ actually defines from the previous revision. In this revision, I used the following definition. Hopefully it makes sense. 

- R is defined to be "the resource-set representation" 
- The R format: the format of this representation
- An R JSON document or in short an R document: a document that conforms to the format in JSON.

dongahn:
Addressed @grondo and @garlick's comments and squashed the commits. This should be ready to go in. Thanks everyone for reviewing it. It should give me a firm target to work on. Don't need to worry about other emit types...

grondo:
Thank you, @dongahn!

dongahn:
Thanks.

Issue #131, opened 2018-08-23
create RFC for job manager states


garlick:
States known by job manager, separate from per-scheduler or per-job shell states which go in their own state log are (proposed):

* Submitted
* Held
* Allocated
* Running
* Completed
* Failed
* Canceled
* Inactive

dongahn:
Great start!

Do you plan to add what component (e.g., job-ingest, scheduler and job shell) the job manager will interact with to transition these job states?

Do you expect the state log will only be used and updated by the job manager?

SteVwonder:
> Do you plan to add what component (e.g., job-ingest, scheduler and job shell) the job manager will interact with to transition these job states?

That's a good idea.  Maybe we can include the "job lifecycle" diagram in the RFC.  Although maybe all the interactions is a separate RFC.  Thoughts @garlick and @grondo?

> Do you expect the state log will only be used and updated by the job manager?

For these specific states, the job manager will be the authority.  It will be the only entity updating the "top-level state log".  Other modules like the scheduler and job shell will have their own state logs that they can append to.  For example, if a job is spawned using SCR, the job shell might record that the job cycles between `checkpointing`, `running`, `failed`, and `restarting`, as SCR helps the job prepare for and recover from failure.

SteVwonder:
> I do hope that the scheduler will only have to deal with one type of event set (which would be a superset of the job manager states).

Yeah, that is the idea.   The scheduler should only need to "subscribe" to events/state changes from the job manager.  And if the scheduler needs to record finer granularity state changes besides `Submitted` -> `Allocated`, it can do so in its own state log.

For user-facing tools, they will start at the job manager state log level, and if more information is required as to what phase of "running" the job is in, the tool can query the job shell/exec module state log.

SteVwonder:
> It seems it will also be useful to map out the valid state transitions for the scheduler somewhere.

That does make a lot of sense to include in this RFC. A simple state machine diagram would be nice.  We had a sketch of one on a whiteboard that I could replicate with tikz pretty easily.

garlick:
I am planning on submitting a job state RFC soon but wanted to touch base here.

Is there a preferred way to include a state diagram image in an asciidoc RFC?  I see RFC 4 has some examples. I was thinking of using graphviz to generate it, although tikz was mentioned above.  Any preference?

I was going suggest the following properties in the RFC
[...]

PR #157, opened 2019-02-10
rfc21: add new rfc for job states

garlick:
Here's a first cut at an RFC for job states.

There is some discussion leading up to it in #131.  I guess the main takeaway from there is that this is an over-simplified first cut, which may change as we build up the new exec system.

Anticipating questions about synchronization - I agree that the proposed job states will not be sufficient synchronization points for some critical use cases like debuggers.  There will need to be something more intensive available.  Maybe watching the job eventlog?  Or maybe something different that can span the main and guest KVS namespaces.  TBD, but my main point is there needs to synchronization above and beyond what is proposed here.

Sorry for the lame diagram.   I can revisit later and prettify with `tikz` / `dot2tex`.  I seem to have mucked up my tex install and was going down a rat hole, so I just went with plain old graphviz for now.


garlick:
> can you add an "active" label to the box around the "active" states subgraph?

good idea - let me just add that now since it's trivial in dot.

grondo:
This is really nicely done. I especially like how you evolved the job exception system. This is familiar territory for programmers, and seems to easily allow addition of future exception types of sub-types easily, without requiring the exceptional condition to somehow be encoded in the job state! Great improvement IMO

grondo:
Is there a self transition `CLEANUP -> CLEANUP` missing in the state diagram for `free` or other exceptions that occur during this state? I'm not sure it adds any value, but I'm wondering in general about additional execptions occuring during this phase (required epilog fails to run, node goes down during free with some purge process incomplete, another cancel request, etc)

garlick:
> Is there a self transition `CLEANUP -> CLEANUP` missing in the state diagram for `free` or other exceptions that occur during this state? I'm not sure it adds any value, but I'm wondering in general about additional execptions occuring during this phase (required epilog fails to run, node goes down during free with some purge process incomplete, another cancel request, etc)

My thinking was the first exception is the one most likely to be the root cause of the failure, and a user waiting on CLEANUP state could fetch (or be handed) that one immediately.  Any exceptions after that would be logged to the eventlog and so could be either fetched after INACTIVE is reached, or if we create a mechanism for non instance owners to watch the eventlog, maybe they could track it that way.

In general though I was hoping to avoid cycles in the state machine, to keep the interface for synchronization at this level as simple as possible.

grondo:
> In general though I was hoping to avoid cycles in the state machine, to keep the interface for synchronization at this level as simple as possible.

Yeah, I did see that stated in Design Criteria but I wasn't thinking clearly enough to realize that is a cycle. Essentially, you answered my question exception in CLEANUP, log an event and stay in CLEANUP. Makes sense to me to leave that off the state diagram.

grondo:
> My thinking was the first exception is the one most likely to be the root cause of the failure

Just one clarification here -- the case I was thinking of is a failure *during* CLEANUP (see examples above). A job has exited "normally" (no exception, no non-zero exit codes recorded), but some tasks performed during CLEANUP log an exception.

Outside the scope of this discussion here: but if a user `wait()`s on a job in CLEANUP that had an exception in during CLEANUP state, but otherwise experienced normal termination, the wait should report normal termination (however that is done), and not propagate the exception I assume.

garlick:
Yes that is what I assumed you meant, and I was thinking exceptions after a global status for the job is available would not affect the "success" of the job.

Maybe one other point is whether or not a non-zero global status (e.g. from the _program_ not the job shells) triggers an exception at this level or is just a normal "result".   I didn't take a stand on that one in the RFC but I was leaning towards the latter.

grondo:
> Maybe one other point is whether or not a non-zero global status (e.g. from the program not the job shells) triggers an exception at this level or is just a normal "result". I didn't take a stand on that one in the RFC but I was leaning towards the latter.

I agree with you here, fwiw.

Since you have developed an extensible exception system here, maybe (in the future) there would be a way to have the job shell emit an exception on non-zero global exit or even other custom conditions (without saying at this time exactly how these exceptions are propagated...)

dongahn:
> This is really nicely done. I especially like how you evolved the job exception system. This is familiar territory for programmers, and seems to easily allow addition of future exception types or sub-types easily, without requiring the exceptional condition to somehow be encoded in the job state! Great improvement IMO

I agree!

Just a comment relevant to future extensions.

@SteVwonder and I have been working with the UQPipeline team to define a new HPC workflow exception model. It seems what workflow teams generally want is to capture the notion of both "soft" and "hard" exceptions.

With the current transition of all of exception events going to `CLEANUP` it may confine us when the exception recovery action (i.e., in particular the exception type is "software") should be just "continue" (after the exception is handled). 

Have you thought about this use case? 

If you want to capture this in the current exception system, this will undoubtedly add cycles to the transition and complexity. So, one possibility is to make distinction between "soft" and "hard" exceptions and explain that the current _CLEANUP_ transition is only for "hard" type?

garlick:
@dongahn, are you proposing that we generalize the `exception` event to allow for "soft" exceptions that don't trigger a job state change?  That way a workflow manager could watch a job's eventlog for the `exception` event and react to both hard and soft exceptions (with the source of soft exceptions TBD - maybe give application and/or job shell a way to raise them?)

We could add a `severity` attribute to `exception` events, say, with `severity=0` triggering the job teardown, and other severities left to a future "exception model" RFC?

I'm not sure if you're questioning the linear state flow proposed here.  I think there are definitely some issues related to grow and shrink that will have to be addressed that affect your exception model (like grow to replace a bad resource), and we will have to decide if that means cycles in the job state machine, e.g. RUN -> SCHED -> RUN, or whether that can be a detail within RUN.  I'm proposing that we punt on this advanced stuff for now (and secretly hoping we can keep state machine simpile :-)

garlick:
(Sorry Dong, I just reread your message and the question I asked at the beginning of my response is exactly what you suggested in your last paragraph - I guess i glossed over that on first read).

grondo:
Another thought that occurred to me -- no cycles means a job can only have dependencies satisfied *once*. That is, once you leave the DEPEND state, a user or system service cannot add another dependency to a job. (feature?)

Also, just to clarify, a job which has met dependencies but doesn't yet have an outstanding `alloc` request (due to priority or rejected alloc request) is still in the SCHED state, correct?

dongahn:
> @dongahn, are you proposing that we generalize the `exception` event to allow for "soft" exceptions that don't trigger a job state change? 

Yes, that's more or less what I was thinking about. 

> We could add a `severity` attribute to `exception` events, say, with `severity=0` triggering the job teardown, and other severities left to a future "exception model" RFC?

Yes, this would be good to add it to this PR so that we can at least suggest our intention here on our future work. 

> That way a workflow manager could watch a job's eventlog for the `exception` event and react to both hard and soft exceptions (with the source of soft exceptions TBD - maybe give application and/or job shell a way to raise them?)

Not an ask for this PR. But...

At least, I think we should probably end up wanting applications and/or workflow managers to be able to detect soft exceptions by putting some filters/markers to the various subsystem activities that Flux can monitor and/or manage. For example, they can put a filter to stdout/stderr stream manager to detect mesh-tangling events or similar.

We would also need to make our upcoming job dependency system to be exception-model aware so that we can retry the job, or start another job or start one type of jobs of several other recovery jobs if that's a part of handling. (Though we should start simple).

Finally, the model we propose is "cooperative" where applications, workflow managers, and resource managers are handling exceptions in a cooperative and chained manager. So, we would ultimately need a mechanism by which an exception can be handled by different components or different types of components in a specific order -- that is when two or more components want to handle the same type of exception...

We will soon write down our thoughts in a position paper. So we should be better verse in the future.

garlick:
> Another thought that occurred to me -- no cycles means a job can only have dependencies satisfied _once_. That is, once you leave the DEPEND state, a user or system service cannot add another dependency to a job. (feature?)

This assumes dependency resolution is a phase that occurs before a job requests resources, so whether dependencies are dynamic (computed some way?) or static (part of jobspec?), at some point a dependency subsystem says "resolved" and the job enters SCHED state, and new dependencies cannot be meaningfully added.  Is that a mistaken assumption?

> Also, just to clarify, a job which has met dependencies but doesn't yet have an outstanding `alloc` request (due to priority or rejected alloc request) is still in the SCHED state, correct?

Right, whether alloc requests are submitted as soon as the job enters SCHED state, or deferred based on some flow control algorithm would be a detail not exposed in the state machine.

garlick:
> > We could add a `severity` attribute to `exception` events, say, with `severity=0` triggering the job teardown, and other severities left to a future "exception model" RFC?

> Yes, this would be good to add it to this PR so that we can at least suggest our intention here on our future work.

Great!  OK, I will add this.

garlick:
@dongahn, on your other comment, I wonder if we could get pretty far by adding a `flux_job_raise()` API call that could be called by a workflow manager, job shell, stdio badness detector, or whatever, to raise both hard and soft exceptions?  In fact it's an obvious superset of `flux_job_cancel()` so poof! that goes away.

Then provide a mechanism for all exceptions to be logged to the main job eventlog via the job manager, and a way (either direct KVS watch or some other mechanism) for other parts of the system or user tools to monitor job events, or specifically, exception events?

dongahn:
> This assumes dependency resolution is a phase that occurs before a job requests resources, so whether dependencies are dynamic (computed some way?) or static (part of jobspec?), at some point a dependency subsystem says "resolved" and the job enters SCHED state, and new dependencies cannot be meaningfully added. Is that a mistaken assumption?

Just a question: What happens if the user adds a dependency right after the job is submitted when the job already move to the SCHED state? Maybe he/she needs some dynamic entity from the script which is used to submit this job to specify a dependency. For the first few jobs, the job can go into this state pretty quickly? Maybe we can use small cycles in a judicious manner...

grondo:
> new dependencies cannot be meaningfully added. Is that a mistaken assumption?

I think symbolic dependencies go a long way toward making a static dependency system a possibility. The specific case I was thinking of is treating "earliest start time" as a depedency. A user may want to adjust something like this after submission, though I admit allowing that may be a misfeature. 

Also as I write that, treating begin time as a dependency might not be what users want because they don't even get seen by the scheduler until dependencies are met, so they lose any queue time priority the job may have had. (Though this is the same problem with other dependencies, so maybe not a real issue). Users can just run an `at` script to submit their jobs in the future, so not sure my points are adding any value here. Sorry!

grondo:
> What happens if the user adds a dependency right after the job is submitted when the job already move to the SCHED state?

I think what we're assuming here is that this is not allowed. The dependencies have to be described in the jobspec. (We don't have a good story for how *any* job parameter is adjusted after submission. If we're trying to keep things simple, jobs are immutable and require cancellation/resubmission to change any parameter...)

garlick:
> Just a question: What happens if the user adds a dependency right after the job is submitted when the job already move to the SCHED state? Maybe he/she needs some dynamic entity from the script which is used to submit this job to specify a dependency. For the first few jobs, the job can go into this state pretty quickly? Maybe we can use small cycles in a judicious manner...

It would be ignored, but I think you could get the desired effect by submitting the job with an unresolved symbolic dependency, then add the dependencies you compute after submission, then resolve the first symbolic dependency.  I guess it depends (ha hah) on how we build out the dependency system...  I wasn't imagining that dependencies would be added like that.

By the way the design I had in mind was for the job manager to send a request to a dependency subsystem which would block until all dependencies are resolved.  I was leaving all the details up to that service, but the one thing I was assuming is that once the service responds, that's the end of it.

garlick:
> Also as I write that, treating begin time as a dependency might not be what users want because they don't even get seen by the scheduler until dependencies are met, so they lose any queue time priority the job may have had.

I guess it depends on how the scheduler is implemented.  If it's got its own queue that it keeps ordered by submit time, then an "older" job could jump in front of a newer one that happened to reach the scheduler first due to having its dependencies resolved first.

grondo:
> By the way the design I had in mind was for the job manager to send a request to a dependency subsystem which would block until all dependencies are resolved.

FWIW, I like this design. My one question is how/if that affects queue priority of jobs with dependencies. In a more typical design the "scheduler" sees all jobs even those with dependencies. A job sitting in the queue for a long time with unmet dependency may be able to immediately bubble to the top of the queue once all dependencies are met. In this system, the "scheduler" only is even notified about a job once all dependencies are met.

Feel free to ignore this for now, but I wanted to bring up the question now since we are on the subject.

grondo:
> I guess it depends on how the scheduler is implemented. If it's got its own queue that it keeps ordered by submit time, then an "older" job could jump in front of a newer one that happened to reach the scheduler first due to having its dependencies resolved first.

This could certainly work, however I guess the point I was trying to make is that any scheduler that takes into account queue time of existing jobs can't account for any jobs waiting on dependencies in this system, whereas I assume most current schedulers can and do. The net effect I guess is that once a job that was long waiting on dependencies is submitted to the scheduler, there is a large queue time factor *suddenly* added to the internal fair-share calculation. I have no idea what the effect would be, probably just massive reordering on queue on every job dependency resolution.

grondo:
BTW, this is also a good argument for immutable jobs. People can't try to game the system by submitting a bunch of jobs with "dependencies" so they can have a bunch of "ballast" with long "queue times" and then altering the jobs and "submitting" them by satisfying dependencies...

garlick:
> My one question is how/if that affects queue priority of jobs with dependencies. In a more typical design the "scheduler" sees all jobs even those with dependencies. A job sitting in the queue for a long time with unmet dependency may be able to immediately bubble to the top of the queue once all dependencies are met. In this system, the "scheduler" only is even notified about a job once all dependencies are met.

In this design the scheduler has no notion of when a symbolic dependency will be fulfilled anyway.   I don't see how it can do any planning involving jobs with unmet dependencies unless it also knows that running some set of other jobs will fulfill those dependencies.

OTOH, once a job's dependencies are met, the arriving alloc request could cause the scheduler to throw out its plan and make a new one.  It'll still be carrying its compelling submit time and/or priority even if the alloc request arrived late.   Although, maybe there's a missed opportunity to create a reservation earlier, and thus the high priority job could be delayed.

garlick:
Ah, we could have "dependency oblivious" schedulers at first, and maybe develop some way later on for the scheduler to interact with the dependency system to get smarter, like retrieve the whole DAG of dependencies by jobid, become aware of jobs that are blocked in DEPEND, etc..

grondo:
> In this design the scheduler has no notion of when a symbolic dependency will be fulfilled anyway. I don't see how it can do any planning involving jobs with unmet dependencies unless it also knows that running some set of other jobs will fulfill those dependencies.

I'm sorry I had deleted that comment, and didn't mean to discuss scheduler planning but prioritization. I only meant to make the same point you are making, that with this system the scheduler is not able to make any plan/prioritization/etc. with jobs with outstanding dependencies (then that was all). I apologize for bringing it up.

grondo:
BTW, to get a little back on track: I really like the proposal for `flux_job_raise()` above.

And since it didn't seem like there was any substantive changes required for this RFC, I say it is ready to merge once rebased.

garlick:
> that with this system the scheduler is not able to make any plan/prioritization/etc. with jobs with outstanding dependencies

I hadn't really thought about that point before, so I'm glad you brought it up!

garlick:
> And since it didn't seem like there was any substantive changes required for this RFC, I say it is ready to merge once rebased.

Thanks!  I was going to add a `severity` attribute to the `exception` event to help @dongahn pave the way for his HPC exception manager stuff, so I'll go ahead with that first, then ready.

grondo:
Thanks, merging. Other updates/changes can come as future PRs, which will be a nice way to track changes.

garlick:
Thanks for the good discussion!

dongahn:
> @dongahn, on your other comment, I wonder if we could get pretty far by adding a `flux_job_raise()` API call that could be called by a workflow manager, job shell, stdio badness detector, or whatever, to raise both hard and soft exceptions? In fact it's an obvious superset of `flux_job_cancel()` so poof! that goes away.

Yes, it seems this is an excellent low-level API! We will need other APIs to program the excepting conditions as well.

> Then provide a mechanism for all exceptions to be logged to the main job eventlog via the job manager, and a way (either direct KVS watch or some other mechanism) for other parts of the system or user tools to monitor job events, or specifically, exception events?

Yes, and plus, the dependency service may also need to make use of the exception events reported to the event log to satisfy certain dependencies.

PR #158, opened 2019-02-11
RFC14: Link to RFC20 and clarify the semantics of count

dongahn:
This PR links the jobspec RFC to the new RFC 20 and also addresses https://github.com/flux-framework/rfc/pull/151#issuecomment-461153941.

### DIFF
--- a/spec_14.adoc
+++ b/spec_14.adoc
@@ -24,6 +24,7 @@ be interpreted as described in http://tools.ietf.org/html/rfc2119[RFC 2119].
 
 * link:spec_4{outfilesuffix}[4/Flux Resource Model]
 * link:spec_8{outfilesuffix}[8/Flux Task and Program Execution Services]
+* link:spec_20{outfilesuffix}[20/Resource Set Specification Version 1]
 
 == Goals
 
@@ -149,7 +150,13 @@ and additionally MAY contain the following keys:
 +
 The default value for `max` SHALL be  _infinite_, therefore a `count`
 which specifies only the `min` key SHALL be considered a request for
-_at least_ that number of a resource.
+_at least_ that number of a resource, and the scheduler SHALL generate
+the _R_ that contains the maximum number of the resource that is
+available and subject to the operator and operand. By contrast,
+if a fixed count is given to the `count` key, the scheduler SHALL
+match any resource that contains _at least_ `count` of the resource,
+but its _R_ SHALL contain exactly `count` of the resource
+(potentially leaving excess resources unutilized).
 
 A resource vertex MAY additionally contain one or more of the
 following keys
--- a/spell.en.pws
+++ b/spell.en.pws
@@ -397,3 +397,4 @@ gpu
 sched
 svg
 invariants
+unutilized


spec_14.adoc
```diff
@@ -149,7 +150,13 @@ and additionally MAY contain the following keys:
...
 The default value for `max` SHALL be  _infinite_, therefore a `count`
 which specifies only the `min` key SHALL be considered a request for
-_at least_ that number of a resource.
+_at least_ that number of a resource, and the scheduler SHALL generate
+the _R_ that contains the maximum number of the resource that is
+available and subject to the operator and operand. By contrast,
+if a fixed count is given to the `count` key,
+the scheduler SHALL generate the _R_ that contains exactly
+that number of the resource while it SHALL still match a resource that
+has _at least_ that number of the resource.
```
grondo: > ... generate the _R_ that contains exactly that number of the resource while it SHALL still match a resource that has _at least_ that number of the resource.

I'm probably just not understanding the intent, but is the second part of this sentence redundant? I'm reading it something like `x == 42 && x >= 42`.

dongahn: Maybe I wasn't clear enough. The intent is to avoid a possible misunderstanding that some folks have had: if a single scalar is given to the `count` key, does the scheduler perform an exact match?  IOW, will the scheduler only match a socket that contains *exactly* 4 cores (not a socket that contains 16 cores) if `count=4` is given whereas it will match a socket that contains 16 cores if `count.min=4` is given? The answer is no of course. 

I wanted to capture the main difference between the two requests: *R*s being generated by the scheduler are different.

grondo: Oh, on re-reading I think I see what you mean. It would have been clearer to me if the sentence were ordered as:

"By contrast, if a fixed count is given to the `count` key the scheduler SHALL match any resource that contains _at_least_ `count` of the resource, but _R_ SHALL contain exactly `count` of the resource (potentially leaving excess resources unutilized)"

Just a suggestion.

grondo: @dongahn, thanks I understand now. What do you think re-wording something like above?

dongahn: :-) Works for me.

dongahn:
OK. Reworded and rebased. Thanks.

grondo:
Sorry I guess "unutilized" is not in the dictionary. Perhaps there is a better term, or we can add it to the spelling dictionary because I do think it is a [real word](https://en.oxforddictionaries.com/definition/unutilized).

spec_14.adoc
```diff
@@ -149,7 +150,13 @@ and additionally MAY contain the following keys:
...
 The default value for `max` SHALL be  _infinite_, therefore a `count`
 which specifies only the `min` key SHALL be considered a request for
-_at least_ that number of a resource.
+_at least_ that number of a resource, and the scheduler SHALL generate
+the _R_ that contains the maximum number of the resource that is
+available and subject to the operator and operand. By contrast,
+if a fixed count is given to the `count` key, the scheduler SHALL
+match any resource that contains _at least_ `count` of the resource,
+but its _R_ SHALL contain exactly `count` of the resource
+(potentially leaving excess resource unutilized).
```
grondo: Should it be "leaving excess resource*s* unutilized"? Or is singular correct here in your opinion?

spec_14.adoc
```diff
@@ -149,7 +150,13 @@ and additionally MAY contain the following keys:
...
```
grondo: Should the parenthetical statement use plural "resources" or is the singular correct here?

dongahn: Oops. My bad.

dongahn:
OK. Forced a push. Also somehow the spell checker in the Travis CI didn't have the word, `unutilized` so I added that to the `spell.en.pws`.

grondo:
Ok, I tried out adding `merge-when-passing` label to your PR and mergify seems to have dinner the merging

PR #162, opened 2019-02-21
rfc20: allow multiple ranks per R_lite entry

### LINKED ISSUE #161: rfc20: allow multiple ranks per entry in R_lite
I'd like to extend the format of the `execution.R_lite` key in RFC 20 to allow a more compact representation of resource sets where many ranks have the same sets of resource identifiers allocated.

The gist is to allow either an alternate key `ranks` instead of `rank`, which will be a string encoding a set of ranks to which the current entry applies, or to allow the existing `rank` key in `R_lite` to be either an integer or a string. If `rank` is a string, then it must be a set of ranks in idset format.

Which would you prefer?

As an example, here's an *R* for a 1024 node cluster with 32 cores per node using the `ranks` key:

```json
{
  "execution": {
    "R_lite": [
      {
        "children": {
          "core": "0-31"
        },
        "ranks": "0-1023"
      }
    ]
  },
  "version": 1
```

grondo:
As discussed in #161, allow multiple ranks per R_lite "entry" object to allow for condensed representation.

Also makes `node` an optional member of *R_lite* entries (not sure hostnames belong in R_lite anyway), and reference an TBD (to be drafted) RFC for the *idset* format.

Closes #161

### DIFF
--- a/spec_20.adoc
+++ b/spec_20.adoc
@@ -23,6 +23,7 @@ in this document are to be interpreted as described in RFC 2119.
 * link:spec_14{outfilesuffix}[14/Canonical Job Specification]
 * link:spec_15{outfilesuffix}[15/Independent Minister of Privilege for Flux]
 * link:spec_16{outfilesuffix}[16/KVS Job Schema]
+* link:spec_22{outfilesuffix}[22/Idset String Representation]
 
 == Overview
 
@@ -103,12 +104,10 @@ the following two keys:
    extensions.
 
 [horizontal]
-    *core*::: The `core` key contains a logical compute core IDs string.
-     It SHALL be a comma-separated list of core IDs (e.g., 1,3,4,5) or
-     hyphenated ranges for consecutive integer IDs (e.g., 1,3-5).
-    *gpu*::: This OPTIONAL key contains a logical GPU IDs string.
-     It SHALL be a comma-separated list of gpu IDs or hyphenated ranges
-     for consecutive integer IDs.
+    *core*::: The `core` key SHALL contain a logical compute core IDs string
+     in RFC 22 *idset format*.
+    *gpu*::: The OPTIONAL `gpu` key SHALL contain a logical GPU IDs string
+     in RFC 22 *idset format*.
 
 An `R_lite` dictionary entry MAY also contain any of the following optional
 keys:
--- a/spec_20.adoc
+++ b/spec_20.adoc
@@ -94,9 +94,10 @@ and `expiration`. `R_lite`
 is a strict list of dictionaries each of which SHALL contain at least
 the following two keys:
 
-  *rank*:: The value of the `rank` key SHALL be the integer broker rank that
-  manages the target compute node. This SHALL disambiguate the case where multiple
-  brokers manage a same node.
+  *rank*:: The value of the `rank` key SHALL be a string list of
+   broker rank identifiers in *idset format* (See RFC 22). This list
+   SHALL indicate the broker ranks to which other information in
+   the current entry applies.
 
   *children*:: The `children` key encodes the information about certain compute resources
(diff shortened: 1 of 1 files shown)


grondo:
Wait to merge this one until we have a new RFC for the *idset format*. I'll update the commits here to reference the actual RFC and then we can merge. This PR is up early for discussion because the change to *R_lite* seemed to be a bit controversial.

dongahn:
> Wait to merge this one until we have a new RFC for the _idset format_. I'll update the commits here to reference the actual RFC and then we can merge.

Ah. Yes, this makes sense. In fact I was going to make some suggestion about that portion.

BTW, after you add a new RFC for the idset format,  could you also make the mod as part of this PR?

<     *core*::: The `core` key contains a logical compute core IDs string.
     It SHALL be a comma-separated list of core IDs (e.g., 1,3,4,5) or
     hyphenated ranges for consecutive integer IDs (e.g., 1,3-5).

>  *core*::: The `core` key contains a logical compute core IDs string.
     Its SHALL be in the idset format defined in RFC xx.

grondo:
> BTW, after you add a new RFC for the idset format, could you also make the mod as part of this PR?

That is already there :wink:

dongahn:
FYI -- this will probably be conflicted with PR #164. So whatever goes in later will have to resolve the conflict.

grondo:
Ok, I've rebased, added references to RFC 22 (thanks @garlick), and finally made the `execution.{starttime,expiration}` keys optional.

spec_20.adoc
```diff
@@ -94,9 +94,10 @@ and `expiration`. `R_lite`
 is a strict list of dictionaries each of which SHALL contain at least
 the following two keys:
 
-  *rank*:: The value of the `rank` key SHALL be the integer broker rank that
-  manages the target compute node. This SHALL disambiguate the case where multiple
-  brokers manage a same node.
+  *rank*:: The value of the `rank` key SHALL be a string list of
+  broker rank identifiers in *idset format* (See RFC 22), and indicates
```
dongahn: indicates -> indicate ?

dongahn:
@grondo: This looks good to me. I can merge this as-is, as I can see the word "indicates" can just be used as is (in which case you just don't make it a requirement clause but a descriptive clause). Let me know.

spec_20.adoc
grondo: The subject here is "the value" (which is a  string list of ids), and because value is singular I *think* that indicates is correct.

However, if the sentence is confusing maybe I can reword it. Maybe split into two sentences:

"The value of the `rank` key SHALL be a string list of broker rank identifiers in *idset format* (See RFC 22). This list describes the broker ranks to which other information in the current entry applies."

Is that any better?

dongahn: Yes, that's better. I initially thought you wanted to say:

The value of the `rank` key SHALL be a string list of broker rank identifiers in *idset format* (See RFC 22), and SHALL indicate ...

But the second "SHALL" here can be omitted because the first clause already has SHALL, in this case you want to use "indicate" instead of "indicates"...

In any case, I think you can still also use "indicates". It is just that, in that case, you don't make this clause a requirement but descriptive. 

All minor points...

grondo: Ah, I see! Ok I've updated to my rewording above (before I saw your suggestion to make the clause a requirement). Now that you mention it, maybe the use of SHALL as you suggest is more clear -- I'll switch to that.

grondo:
Thanks! I switched the clause to a requirement as you suggested @dongahn.
Can you give it another look?

dongahn:
Perfect! I will merge it as soon as Travis turns green. Thanks @grondo!

grondo:
Thanks!

PR #163, opened 2019-02-21
rfc22: add idset string specification

garlick:
This RFC defines the (simple!) idset string format, so that other RFCs don't have to.  @grondo suggested it and I thought it was a good idea.


grondo:
Thanks! Just one change: don't use apostrophe to pluralize "ID", i.e. `IDs` not `ID's`

garlick:
Thanks!  Fixed and squashed.

Issue #160, opened 2019-02-15
RFC 20: need a way to express time limit in R?


garlick:
Should there be a way to express an allocation time limit in _R_?

Practically speaking, some part of the system needs to throw a `timelimit` exception on a job once it's outlived its wallclock limit.   AFAICS the only thing we have to go on at this point is `duration` from jobspec, which is optional I think?

Another way to do it is when the scheduler places _R_ in the KVS before responding to a job manager allocation request, it could also write a wallclock expiration time to  another key like `job.active.<jobid>.R_expiration`.

dongahn:
> Should there be a way to express an allocation time limit in _R_?

Yes, I do think we need it. 

> Practically speaking, some part of the system needs to throw a `timelimit` exception on a job once it's outlived its wallclock limit. AFAICS the only thing we have to go on at this point is `duration` from jobspec, which is optional I think?

Yes. The walltime should be determined as a function of the `duration` key in the jobspec and system-imposed policy. Generally speaking, we will need a way to create a system configuration to specify system imposed policy parameters. We had a related discussion at https://github.com/flux-framework/rfc/issues/101#issuecomment-329020512. Though the discussion was on the `exclusive` key, this really applies to a broader topic. 

> Another way to do it is when the scheduler places _R_ in the KVS before responding to a job manager allocation request, it could also write a wall-clock expiration time to another key like `job.active.<jobid>.R_expiration`.

We can add this to the `R_lite` key. But as you suggested, I like this separate key approach better since we will ultimately go to an API approach  and at that point `R_lite` will go away. @grondo?

SteVwonder:
> AFAICS the only thing we have to go on at this point is `duration` from jobspec, which is optional I think?

I think there is also no guarantee that the `duration` in jobspec is a single integer.  It could be a range for some really advanced schedulers.

Having a `timelimit` directly in R could be really helpful when/if Flux instances are temporarily loaning resources to one another.  Then they would only have to send over a single payload with an R.

dongahn:
> Generally speaking, we will need a way to create a system configuration to specify system imposed policy parameters.

For lack of the full system config support, this was my stopgap on flux-sched.

https://github.com/flux-framework/flux-sched/blob/master/resource/config/system_defaults.hpp#L28

garlick:
I think configuration is a separate issue (not that it isn't an important one).  We could always use a broker attribute for the default time limit in the mean time.

[...]

PR #164, opened 2019-02-21
rfc20: Add starttime and expiration keys

dongahn:
This PR addresses Issue #160. This allows the system to raise a `timelimit` exception on a job once it's outlived its wallclock limit.

### DIFF
--- a/spec_20.adoc
+++ b/spec_20.adoc
@@ -87,8 +87,9 @@ the format version.
 
 === Execution
 
-The value of the `execution` key SHALL contain exactly one key
-with other keys reserved for future extensions: `R_lite`. `R_lite`
+The value of the `execution` key SHALL contain exactly three keys
+with other keys reserved for future extensions: `R_lite`, `starttime`,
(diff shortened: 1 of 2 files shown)


grondo:
Thanks @dongahn! I'm happy to merge this first so you don't have to resolve conflicts with #162. However, is it ok if `starttime` and `expiration` are OPTIONAL? (e.g. these won't make much sense when a job has no expiration, or in the *R* for a system instance for example)

I can make the change in my PR along with making the `node` key of *R_lite* entries optional.

dongahn:
Thanks @grondo.

> However, is it ok if `starttime` and `expiration` are OPTIONAL? (e.g. these won't make much sense when a job has no expiration, or in the _R_ for a system instance for example)

It should be okay as far as we describe the semantics of not having those keys. 

> I can make the change in my PR along with making the `node` key of _R_lite_ entries optional.

Yes, this is preferred since I have a meeting to go to now and then will have to knock down a contract item for the remainder of the day.

PR #167, opened 2019-03-11
rfc16: updates related to exec system, guest namespace

grondo:
Just some small updates to the Job KVS Schema to bring the RFC in line with emerging reality:

 - Update job lifecycle to indicate that exec system notifies job manager of "job shell exit" events which triggers the free of resources back to scheduler.
 - Remove copies of `jobspec` and `R` from guest kvs namespace. It seems these copies are not needed once the kvs proxy service is in place.
 - Note that the exec system will copy/link guest namespace into the main KVS namespace before notifying the job manager that the execution is "finished".

### DIFF
--- a/spec_16.adoc
+++ b/spec_16.adoc
@@ -193,12 +193,6 @@ and uses this information to launch _job shells_ and subsequently tasks.
 The _exec system_ creates the job's guest namespace and links it to
 `jobs.active.<jobid>.guest`.  Its initial contents are populated with
 
-`exec.R`::
-copy of `jobs.active.<jobid>.R`
-
-`exec.jobspec`::
-copy of `jobs.active.<jobid>.jobspec`
-
 `exec.eventlog`::
 An eventlog for the use of _job shells_, TBD.
 
@@ -222,9 +216,6 @@ a name unique to the service producing the data.
 
 The application MAY store application-specific data in the guest KVS
 namespace under `application`.
-
-When the application is another Flux instance, `exec.R` MAY be used
-to initialize the resource set managed by the instance.
 
 === Content Consumed/Produced by Tools
 
--- a/spec_16.adoc
+++ b/spec_16.adoc
@@ -196,6 +196,11 @@ The _exec system_ creates the job's guest namespace and links it to
 `exec.eventlog`::
 An eventlog for the use of _job shells_, TBD.
 
+Once all _job shells_ have exited and all outstanding writes to
+the guest namespace have stopped, the _exec system_ links the guest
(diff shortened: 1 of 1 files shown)


garlick:
LGTM thanks!

spec_16.adoc
```diff
@@ -193,12 +193,6 @@ and uses this information to launch _job shells_ and subsequently tasks.
 The _exec system_ creates the job's guest namespace and links it to
 `jobs.active.<jobid>.guest`.  Its initial contents are populated with
 
-`exec.R`::
```
dongahn: @grondo: thank you for putting this together. 

Just as a clarification, I thought once we agreed that we wanted to put the instance *R* into a known location so that each instance knows its overall resource set. The nested scheduler can deserialize that before doing its nested scheduling.

I do think that there is an advantage in keeping this data in the guest KVS somewhere (e.g., better providence, better performance of the nested services that operate on the instance resource set, etc) than needing to fetch the data on the fly from the parent's proxy service. 

I understand you may want to do this for higher throughput. 

So maybe we can either optionally require this or pencil in an alternative where a nested service puts this into a *known* location. 

Also for such a location, I'm not convinced that `exec.R` is the best location. I rather like `resource.R` which you proposed before.

grondo: Ah, there is still a known location `jobs.{active,inactive}.[id].R` and I don't think there will be noticeable performance hit fetching from the proxy service instead of a guest KVS namespace.

For providence, when jobs are inactive the guest namespace is copied into the main kvs namespace anyway, so I don't think providence is helped.

Perhaps a nested instance in its init script can simply fetch `R` once from the proxy service of the parent and then write to `resource.R` in *its* main namespace, then services in the nested instance can all access from that well known location?

grondo: If you want I can open an issue on this or perhaps find one that is already open. I think you have some very valid questions and a good use case above.

dongahn: > Perhaps a nested instance in its init script can simply fetch R once from the proxy service of the parent and then write to resource.R in its main namespace, then services in the nested instance can all access from that well known location?

I like this idea a lot. I think the only thing needed more then is to make sure we do this for the top instance as well so that we can write our code that works from the top to bottom.

I will open a ticket so that we can leave the trace of this discussion there instead of a PR.

grondo: Thanks!

>  I think the only thing needed more then is to make sure we do this for the top instance as well so that we can write our code that works from the top to bottom.

Ah yes, I had originally suggested we do this from `flux hwloc reload` dynamically but pushed this off until we had decided on a non-transitional format for `R`.

PR #169, opened 2019-03-14
rfc21: define the events that drive state transitions in more detail

garlick:
Add detail regarding the events that the job manager logs to the primary job eventlog.

Back out the requirement that the event context use a `key=value [key=value...]` format, and use JSON instead so that we don't set ourselves on a path of reinventing yet another half-baked serialization format.

The transition to JSON was discussed in flux-framework/flux-core#2076

This is a rather big update but I would say the original draft was pretty drafty, and now that we've got some implementation out of the way, we can set more details down here.  There are still some changes required to support partial release of resources and reporting of global exit status, but I thought we are better off documenting what we have now and adding that later.


grondo:
LGTM! Nice work.

PR #171, opened 2019-03-18
rfc21: change title, add clean event

garlick:
Update RFC 21 to add a `clean` event, used to indicate that CLEANUP state activities are over.

Also since the RFC now contains quite a lot of detail about job events, update the title to
_21/Job States and Events_.

Finally, I noticed we didn't mention that a job can have multiple exceptions.  Make that explicit.


garlick:
Thanks!

Issue #172, opened 2019-03-18
rfc 14: allow attributes to have any value type


garlick:
As noted in flux-framework/flux-core#2073 in the context of passing debug input to the exec system, RFC 14 JCR limits user and system attributes to strings only, but the textual description appears to allow a value to be any type.  I'm not sure if this was at one time a real requirement, or just an error.  It seems like allowing any JSON object is a powerful way for users to provide structured input to jobs.  IMHO we should remove the restriction.

Either way, we should clarify the text.  If we want to relax it, we need to either fix the JCR or replace with json-schema (resolving #142).

The json-schema validator in flux-core restricts attributes to strings.   Looks like the cpp libjobspec used in sched also has this restriction.  So those (and their tests) will require attention as well.

grondo:
I'm trying to recall the use cases for these attributes -- as far as I recall

 * `attributes.user` was reserved for arbitrary user metadata which could be passed along in the jobspec and/or stored in the "permanent" job record. The use cases for that are pretty open ended, so it seems clear we wouldn't want to limit the values to strings only.

 * The main use case for `attributes.system` is as a conduit for user-selected optional system components, e.g. similar to Slurm spank plugins. In this case I can't think of cases where we would absolutely need non-string values, however, as shown in flux-framework/flux-core#2073 it might be useful to at least allow it for future flexibility.

I'm sure there are other use cases I'm missing, but even from these two examples we seem to be in agreement that the restriction should be removed. We have a nice way to extend our validation (at least conceptually) of submitted jobspec, so there isn't a compelling reason for the restriction IMO.

garlick:
Closed by #173

PR #173, opened 2019-03-19
rfc14: allow attributes to have any value type

garlick:
As discussed in #172, this PR
* Specifically calls out that system and user attributes can have any value type
* Replaces JCR specification with json-schema from flux-core, which now allows any type

Change the Makefile to generate html using asciidoctor and coderay, which makes a much prettier rendering, with inline source includes expanded and highlighted.

Sadly:  github rendering of includes doesn't even produce a clickable link anymore.  We really need to consider those other document publishing options, time permitting.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,22 +23,19 @@ SOURCES = \
 	spec_22.adoc
 
 HTML = $(SOURCES:.adoc=.html)
-PDF = $(SOURCES:.adoc=.pdf)
+
+# N.B. 'apt-get install codray' to enable inline source highlighting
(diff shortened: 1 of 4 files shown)


garlick:
Sorry for the forced pushes - I noticed the JCR (now json-schema) footnote was rendering improperly on github, tried to use a different footnote style, failed, and finally eliminated the additional reference to the footnote, which seems non-portable.  Now at least it's rendering OK.

grondo:
Thank you for taking care of this!

garlick:
NP!  Thanks.

PR #176, opened 2019-03-22
rfc23: describe a common duration string format

grondo:
This PR adds a description of Flux Standard Duration:tm:, a common string duration format (as in GNU `sleep(1)`)


garlick:
Seems straightforward, thanks!

PR #177, opened 2019-03-22
rfc14: make duration attribute a number of seconds

### LINKED ISSUE #175: RFC 14: define jobspec duration format
Peeling this off of #149 (document minimal jobspec).

RFC 14 does not really define the `duration` attribute in any detail.  It says:
> The value of the duration attribute is a string representing time span. The scheduler will make an effort to allocate the requested resources for the time specified in duration.

The jobspec valid inputs (from flux-core sharness test) are:
```
$ grep duration valid/*
valid/example1.yaml:    duration: 1 hour
valid/example2.yaml:    duration: 1 hour
valid/use_case_1.1.yaml:    duration: 1 hour
valid/use_case_1.2.yaml:    duration: 1 hour
valid/use_case_1.3.yaml:    duration: 1 hour
valid/use_case_1.4.yaml:    duration: 1 hour
valid/use_case_1.5.yaml:    duration: 4 hours
valid/use_case_1.6.yaml:    duration: 1 hour
valid/use_case_1.7.yaml:    duration: 1h
valid/use_case_2.1.yaml:    duration: 1 hour
valid/use_case_2.2.yaml:    duration: 1 hour
valid/use_case_2.3.yaml:    duration: 1 hour
valid/use_case_2.4.yaml:    duration: 1 hour
valid/use_case_2.5.yaml:    duration: 1 hour
valid/use_case_2.6.yaml:    duration: 1 hour
```
The json-schema simply requires
```json
"duration": { "type": "string" }
```

In #149, @grondo proposed
> Actually we could increase both ease of parsing and readability if we allowed duration to be a string with the same format as GNU sleep(1), i.e.
>
>> NUMBER[SUFFIX] - Duration in NUMBER seconds. SUFFIX may be 's' for seconds (the default), 'm' for minutes, 'h' for hours or 'd' for days. NUMBER may be an arbitrary floating point number.
> 
> Slurm-style duration as above is easily converted to this format, though for flux tools the sleep(1) syntax is much more user-friendly IMO.

Do we want to tighten up RFC 14 to _require_ this format?

grondo:
Problem: RFC 14 specifies that the system attribute "duration" is
a "string representing a time span", but does not specify how that
string is to be interpreted.

Since this RFC describes a canonical jobspec format, specifically
state that the duration is a number of seconds, leaving no question
as to how it should be interpreted.

Update all use case examples that used arbitrary strings for duration.

Fixes #175

### DIFF
--- a/data/spec_14/example1.yaml
+++ b/data/spec_14/example1.yaml
@@ -16,4 +16,4 @@ tasks:
       per_slot: 1
 attributes:
   system:
-    duration: 1 hour
+    duration: 3600.
--- a/data/spec_14/example2.yaml
+++ b/data/spec_14/example2.yaml
@@ -13,4 +13,4 @@ tasks:
       per_slot: 1
 attributes:
   system:
-    duration: 1 hour
+    duration: 3600.
--- a/data/spec_14/use_case_1.1.yaml
+++ b/data/spec_14/use_case_1.1.yaml
@@ -13,4 +13,4 @@ tasks:
       per_slot: 1
 attributes:
   system:
-    duration: 1 hour
+    duration: 3600.
--- a/data/spec_14/use_case_1.2.yaml
+++ b/data/spec_14/use_case_1.2.yaml
@@ -17,4 +17,4 @@ tasks:
       per_slot: 1
 attributes:
   system:
-    duration: 1 hour
+    duration: 3600.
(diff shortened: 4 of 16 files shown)


garlick:
LGTM, tahnks!

grondo:
Thanks!

PR #178, opened 2019-03-22
rfc14: clarify duration attribute range and meaning

grondo:
As mentioned in flux-framework/flux-core#2096, the `duration` system attribute should have a minimum of `0`, where `0` (zero) itself is reserved for a special meaning of either "not set" or "unlimited".

After I wrote that I though perhaps "not set" could be represented by literally not setting the `duration` attribute. Should I change the wording here to state that zero means "unlimited" duration?

### DIFF
--- a/spec_14.adoc
+++ b/spec_14.adoc
@@ -278,9 +278,11 @@ at all possible.
 Some common system attributes are:
 
  *duration*::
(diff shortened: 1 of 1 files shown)


garlick:
Looks good, thanks!

Issue #180, opened 2019-04-04
rfc18: convert events to JSON


garlick:
As discussed in flux-framework/flux-core#2106, now that job manager events use JSON for the context portion of RFC 18 format events, it seems evident that it would have been better to define each event as a one-line JSON object so that we don't need any one-off parsing code.

JSON events will have to be encoded single-line; that is, no embedded newline characters since that is the separator for events in the event log.

Note that the eventlog is not valid JSON, although interfaces that accept "streaming JSON" like `jq` may still be able to deal with it, and representing it as a JSON array may be useful internally.

RFC 21 will also need minor updates after this change.

PR #181, opened 2019-04-05
RFC 18: require events to be JSON objects

garlick:
As discussed in #180, it's useful to require all events appended to Flux KVS Event Logs to be JSON objects, rather than the hybrid format we ended up with in the current job manager.  This PR makes that change to the event log RFC.

### DIFF
--- a/spec_18.adoc
+++ b/spec_18.adoc
@@ -32,24 +32,37 @@ enables a Flux Event Log to be used for synchronization.
 
 == Event Log Format
 
-An event log SHALL be append-only.  An event appended to a KVS event log
-MUST take the following form:
+A Flux KVS Event Log SHALL consist of events separated by newlines.
+Each event SHALL be an independent JSON object, serialized without
+embedded newlines.  The Event Log as a whole is not valid JSON.
 
+The following keys are REQUIRED in an event object:
+
+timestamp::
+(number) The time stamp indicating when the event was created,
+represented as seconds since the Unix Epoch (1970-01-01 UTC).
+It MUST be greater than zero, and MAY include sub-second precision.
+
+name::
+(string) The name of the event.
+
+The following keys are OPTIONAL in an event object:
+
+context::
+(object) Application-specific event data.
+
+== Example
+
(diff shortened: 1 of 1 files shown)


spec_18.adoc
```diff
@@ -32,24 +32,37 @@ enables a Flux Event Log to be used for synchronization.
...
+[source]
+----
+{"timestamp":1552593348.073045,"name":"submit,"context":{"priority":16,"userid":5588,"flags":0}}
+{"timestamp":1552593547.411336,"name":"priority","context":{"priority":0,"userid":5588}}
+{"timestamp":1552593348.088391,"name":"alloc",context:{"note":"rank0/core[0-1]"}}
+{"timestamp":1552593348.093541,"name":"free"}
+{"timestamp":1552593348.089787,"name":"start"}
+{"timestamp":1552593348.092830,"name":"release","context":{"ranks":"all","final":true}}
+{"timestamp":1552593348.090927,"name":"finish","context":{"status":0}}
+{"timestamp":1552593348.104432,"name":clean}
```
chu11: quotes around clean?

spec_18.adoc
```diff
@@ -32,24 +32,37 @@ enables a Flux Event Log to be used for synchronization.
...
-The elements (i.e., timestamp, name, and context) within a single event are
-spaced-delimited, and events themselves are newline terminated.
+== Example
+
+An example Flux Event Log:
+
+[source]
+----
+{"timestamp":1552593348.073045,"name":"submit,"context":{"priority":16,"userid":5588,"flags":0}}
+{"timestamp":1552593547.411336,"name":"priority","context":{"priority":0,"userid":5588}}
```
chu11: the timestamp is supposed to be an integer?  So no decimal place?

garlick: I meant that to be a "number" with optional subsecond precision, good catch.

garlick:
LMK if this is good now and I'll squash.

chu11:
LGTM, lets squash

PR #183, opened 2019-04-09
add style requests for enumerations, update style

trws:
This mainly came out of looking at some issues with bindings generated for rust, where our enumerations couldn't be used reasonably because they were all anonymous and had unnamable types.  The tweaks to deal with that are pretty simple, and may help with allowing us to have strongly typed enums in our internal interfaces where feasible as well.

Since I was in here, I also tweaked some of the language, and added a reference to the clang-format file we have to enforce our style.


garlick:
Thanks @trws!

The commit titles here should include an "rfc7: " prefix and be limited to 50 chars (see Patch Requirements in RFC 1).

garlick:
What does the `_b` suffix accomplish?

When a function is declared that accepts a bitmask of enum values, it _should_ be declared as an int, right?

Incidentally an example in the RFC might be useful here.  E.g.
```c
enum flux_msg_type {
    FLUX_MSGTYPE_REQUEST    = 0x01,
    FLUX_MSGTYPE_RESPONSE   = 0x02,
    FLUX_MSGTYPE_EVENT      = 0x04,
    FLUX_MSGTYPE_KEEPALIVE  = 0x08,
    FLUX_MSGTYPE_ANY        = 0x0f,
    FLUX_MSGTYPE_MASK       = 0x0f,
};
```

trws:
In terms of safety in the language, nothing.  It serves a similar purpose to using “_t” on typedefs, it identifies to the developer how that type is used, even if there’s only one value in it or they might not otherwise be sure.  It also serves as a convenient marker for binding generators that operate on patterns, for example generate bitwise operators and implicit conversion to int for enums with the _b suffix and generate strongly typed enumerations for the rest.

Good point on the example, I’ll add in a couple.

On Apr 10, 2019, at 11:26 AM, Jim Garlick <notifications@github.com<mailto:notifications@github.com>> wrote:


What does the _b suffix accomplish?

When a function is declared that accepts a bitmask of enum values, it should be declared as an int, right?

Incidentally an example in the RFC might be useful here. E.g.

enum flux_msg_type {
    FLUX_MSGTYPE_REQUEST    = 0x01,
    FLUX_MSGTYPE_RESPONSE   = 0x02,
    FLUX_MSGTYPE_EVENT      = 0x04,
    FLUX_MSGTYPE_KEEPALIVE  = 0x08,
    FLUX_MSGTYPE_ANY        = 0x0f,
    FLUX_MSGTYPE_MASK       = 0x0f,
};

garlick:
Question:  is the `flux_msg_type` example not also an example of an enum deserving of the `_b` suffix?

trws:
You're probably right @garlick, I was looking at it thinking it would always represent exactly one type rather than a bit mask of all applicable types.  That's actually the distinction I intend to draw, whether it's an enum that represents all valid values as explicitly named variants.  I probably need to find another example for a non `_b` type.

garlick:
How about
```c
enum flux_requeue_mode {
    FLUX_RQ_HEAD = 1,   /* requeue message at head of queue */
    FLUX_RQ_TAIL = 2,   /* requeue message at tail of queue */
};
```
(To follow your advice about derived names, might want to change to `FLUX_REQUEUE_HEAD` etc.)

SteVwonder:
@trws: do you mind adding @garlick's suggestion and doing a quick rebase on this PR?  Looks like the only conflict is on the spelling dictionary, which should be an easy one to resolve.  We want to get this in before @gonsie drops her asciidoc -> restructuredText PR.

If you don't have the time, feel free to give me permissions to your fork and I can do the rebase.

SteVwonder:
Thanks @trws.  Can you throw a `rfc7:` prefix in your commit messages?  I think this is GTG after that.

SteVwonder:
Thanks @trws!  Setting the mergify label

SteVwonder:
@mergifyio rebase

SteVwonder:
Forgot to approve the PR.  Sorry about that 🤦‍♂

PR #185, opened 2019-05-02
rfc16: Remove jobs.active and jobs.inactive

chu11:
Remove jobs.active and jobs.inactive directories, instead, all
job information will be listed under "jobs".

Adjust accordingly in rfc21 as well.

### DIFF
--- a/spec_16.adoc
+++ b/spec_16.adoc
@@ -82,7 +82,7 @@ The Flux instance has a default, shared namespace that is accessible
 only by the instance owner.
 
 All job data is stored under a `jobs` directory in the primary
-namespace.  Each job has a directory under `jobs.<jobid>`, where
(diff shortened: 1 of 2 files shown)


chu11:
per some discussion in https://github.com/flux-framework/flux-core/pull/2107 and later https://github.com/flux-framework/flux-core/issues/2093

chu11:
While working on this problem, I noticed that `flux-core` implemented the `jobs` (plural) directory as `job` (singular).  We could go with either `jobs` or `job`, but with the removal of the `active` & `inactive` directories, I feel that the singular `job` feels more correct.  We can go either way on this of course.

garlick:
Good catch on `job` vs `jobs`!  Easier to change the RFC than the code, so let's go with `job`.

garlick:
Perfect, thanks!

PR #189, opened 2019-06-28
RFCs 14 & 25: Jobspec Version 1 + ENV and CWD attributes

### LINKED ISSUE #100: duration attribute in the canonical job spec
@morrone: I am working on a time-related scheduling logic which makes use of attributes like duration, start time, etc in my resource query work. Currently, jobspec has "duration" attribute whose value must be a string and one exampled used is "1 hour." As such, it doesn't require a certain format. It would make my life easier if the allowed time format is specified as part of this spec. Did you have thoughts on this?

### LINKED ISSUE #149: need to document minimal jobspec (V1)
We still need to decide on how to document this (wiki vs rfc).  If we go the RFC route, my suggestion would be to do something along the lines of: RFC 14.1 documents version 1 of the jobspec and RFC 14.2 documents version 2 of the jobpsec, etc.  Rather than having the various jobspec versions each get an entirely different RFC number.

Just to document what we have so far.  The version we had on the whiteboard looked like
```yaml
version: 1
resources:
  - type: node
    count: ... # positive integer
    with:
      - type: slot
        count: ... # positive integer
        label: task
        with:
          - type: core
            count: ... # positive integer
tasks:
  - command: ... # array w/ command and args
    slot: task
    count:
      per_slot: ... # positive integer
attributes:
  system:
    duration: days-hours:minutes:seconds # SLURM-style walltime specification
```

where the `node` resource is optional, the ordering of the resource types is fixed (i.e., `[node->]slot->core`), only 1 slot is allowed, only 1 command is allowed, and only `per_slot` is supported for task count.  #146 discusses some additional attributes that could be added.

### LINKED ISSUE #188: RFC14: Add `environment` key to jobspec
My proposal is to add it under the `attributes` key in the `tasks` list.  Thus, each task in the jobspec can have it's own environment. I'm guessing it is easiest to have it as a dictionary with the env variable name as the key and the environment variable value as the value, but I can also see representing it as a large string.

It could also potentially go under the `system` key in the top-level `attributes` section.  This would place it alongside `duration` and apply to every task in the jobspec.  In the name of simplicity and explicitness, I'm thinking that it is better to have it in the `tasks` list.  This way, each task can have a different environement, and we don't have to deal with reconciling the two "layered" environments.  As with other design decisions around jobspec, we can always make a higher-level, more concise format that compiles down into the more verbose jobspec.  On the other hand, having the same environment duplicated in every task will get big, quick.

Thoughts?

Example:
```yaml

version: 1
resources:
  - type: node
    count: 4
    with:
      - type: slot
        count: 1
        label: default
        with:
          - type: core
            count: 2
tasks:
  - command: app
    slot: default
    count:
      per_slot: 1
    attributes:
      environment:
        PATH: "/foo/bar:/john/paul:/george/ringo"
        PS1: ">"
        HOME: "/home/sherbein"
attributes:
  system:
    duration: 3600.
```

SteVwonder:
Still TODO:

- [x] resolve the discussion in #2214 and integrate any changes
- [x] validate the rfc25 examples against the V1 jobspec schema
- [x] add missing `.` back into `data/spec_14/use_case_1.3.yaml`
- [x] make RFC 2% a "delta" of 14 (rather than a complete copy)
- [x] Per https://github.com/flux-framework/flux-sched/pull/502, make `exclusive: false` under a `slot` invalid
- [x] Nix the exclusive option
- [ ] ~~Integrate lessons learned and requirements from `flux run` design~~
- [ ] ~~Add `slot>node` as a valid configuration (`node>slot` is not)~~
- [ ] ~~Add examples and jsonschema for `per_resource`~~
- [ ] ~~Discussion about warning that should be emitted (task vs slot vs execution target counts)~~

I'm happy to split the environment and cwd changes out into a separate PR if desirable.

Closes #100
Closes #188 
Closes #149

### DIFF
--- a/spec_14.adoc
+++ b/spec_14.adoc
@@ -287,7 +287,7 @@ Some common system attributes are:
 === Example Jobspec
 
 Under the description above, the following is an example of a fully compliant
-version 1 jobspec. The example below declares a request for 4 "nodes"
+canonical jobspec. The example below declares a request for 4 "nodes"
 each of which with 1 task slot consisting of 2 cores each, for a total
 of 4 task slots. A single copy of the command `app` will be run on each
 task slot for a total of 4 tasks.
@@ -306,7 +306,7 @@ include::data/spec_14/example2.yaml[]
 
 === Schema
 
-A jobspec conforming to version 1 of the language definition SHALL
+A jobspec conforming to the canonical language definition SHALL
 adhere to the following ruleset, described using JSON Schema
 footnote:[https://json-schema.org/latest/json-schema-core.html[JSON Schema: A Media Type for Describing JSON Documents]; H. Andrews; 2018].
 
--- a/spec_14.adoc
+++ b/spec_14.adoc
@@ -260,7 +260,7 @@ which, if present, must have values.  Values MAY have any valid YAML type.
 
  *user*::
  Attributes in the `user` dictionary are unrestricted, and may be used
- as the application demands.  Flux may provide addition tools that can
+ as the application demands.  Flux may provide additional tools that can
  identify jobs based on `user` attributes.
 
  *system*::
--- a/spec_14.adoc
+++ b/spec_14.adoc
@@ -291,7 +291,7 @@ Some common system attributes are:
  representing the environment variable name and the `value` is a string
  representing the environment variable value to set.  A `null` `value`
  represents unsetting the environment variable given by `key`.  The values
- provided here can be overriden per-rak by providing the
+ provided here can be overridden per-rank by providing the
  `attributes.environment` dictionary under the target task.
 
  *cwd*::
--- a/spec_14.adoc
+++ b/spec_14.adoc
@@ -284,6 +284,20 @@ Some common system attributes are:
  will make an effort to allocate the requested resources for the number
  of seconds specified in `duration`.
 
+ *environment*::
+ The value of the `environment` attribute is a dictionary containing the names
+ and values of environment variables that should be set (or unset) when spawning
+ tasks.  For each entry in the `environment` dictionary, the `key` is a string
+ representing the environment variable name and the `value` is a string
+ representing the environment variable value to set.  A `null` `value`
+ represents unsetting the environment variable given by `key`.  The values
+ provided here can be overriden per-rak by providing the
+ `attributes.environment` dictionary under the target task.
+
+ *cwd*::
+ The value of the `cwd` attribute is a string containing the absolute
+ path to the current working directory to use when spawning the task.
+
 === Example Jobspec
 
 Under the description above, the following is an example of a fully compliant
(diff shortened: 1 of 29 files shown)


SteVwonder:
Just pushed a change that adds integrates the changes from https://github.com/flux-framework/flux-core/issues/2214  and add the missing `.` back in that @dongahn caught.

The jsonschema isn't perfect.  For example, it doesn't check that there is exactly one and only one slot in the jobspec.  While technically doable for this restrictive of a jobspec, it isn't very "clean".  I can add that if people want.

I just need to run the RFC24 examples through the V1 json schema, and then I think this is GTG.

spec_24.adoc
```diff
@@ -0,0 +1,367 @@
...
+
+=== Attributes
+
+The value of the `attributes` key SHALL be a dictionary of dictionaries.
+The `attributes` dictionary MAY contain one or both of the following keys
+which, if present, must have values.  Values MAY have any valid YAML type.
+
+ *user*::
+ Attributes in the `user` dictionary are unrestricted, and may be used
+ as the application demands.  Flux may provide addition tools that can
```
dongahn: s/addition/additional

spec_14.adoc
```diff
@@ -284,10 +284,24 @@ Some common system attributes are:
  will make an effort to allocate the requested resources for the number
  of seconds specified in `duration`.
 
+ *environment*::
```
dongahn: I realized RFC 14 also has a typo: s/addition/additional

spec_14.adoc
```diff
@@ -284,10 +284,24 @@ Some common system attributes are:
...
+ The value of the `environment` attribute is a dictionary containing the names
```
dongahn: Can you also change the type of `duration` to FP to match with V1?

dongahn:
@SteVwonder: Is this PR still WIP or do you want the folks to start a real review?

SteVwonder:
I believe this PR is now ready for a full review.

The json-schema for jobspec V1 now works, and I verified that all of the examples in RFC24 pass the validation against it.  I also verified that all of the examples in RFC24 pass validation against the canonical jobspec json-schema.

Thanks @dongahn for that feedback.  Added some commits to address those issues.

Note: I'd like to squash many of these commits after this PR is approved but before it is merged.  So please don't merge immediately after approving.

garlick:
One quick thought:  looks like RFC 24 started from a clone of RFC 14, which is a reasonable approach.  But what if we just made RFC 24 just a set of constraints on RFC 14?   Just wondering if that would be more concise, and we we would be relieved of then need to keep 14 and whatever the current jobspec version is in sync?  (If I'm misinterpreting 24 sorry - I was trying to read on my phone and may not have gotten the whole picture)

SteVwonder:
> But what if we just made RFC 24 just a set of constraints on RFC 14?

Yeah, that makes a lot of sense.  I'll work on that on I have the python and schedutil refactor PRs up.

SteVwonder:
This is ready for another review pass.

PR #190, opened 2019-07-01
add non-infinite and non-NaN conditions to fsd

trws:
Clarify allowed fsd floats to correspond to changes in flux-framework/flux-core#2216.

### DIFF
--- a/spec_23.adoc
+++ b/spec_23.adoc
@@ -29,10 +29,10 @@ said to support "Flux Standard Duration."
 
 == Implementation
 
-A duration in Flux Standard Duration SHALL be of the form `N[SUFFIX]`
-where `SUFFIX` SHALL be optional and, if provided, MUST be a single character
-from the set `s,m,h,d`. The value `N` MUST be a positive, floating-point number,
-and SHALL be interpreted as:
+A duration in Flux Standard Duration SHALL be of the form `N[SUFFIX]` where
+`SUFFIX` SHALL be optional and, if provided, MUST be a single character from the
+set `s,m,h,d`. The value `N` MUST be a positive, non-infinite, floating-point
+number excluding `NaN`, and SHALL be interpreted as:
 
   * _seconds_ if `SUFFIX` is not provided, or is `s`.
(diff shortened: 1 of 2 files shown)


garlick:
What about zero?  Maybe change to

> The value `N` MUST be a non-negative, non-infinite, floating-point number excluding `NaN`.

I guess our implementation will allow forms like `3.1E10s` which is maybe fine.  Could we say

> `N` SHALL be in one of the forms allowed by POSIX `strtod(3)`.
footnote:[https://pubs.opengroup.org/onlinepubs/009695399/functions/strtod.html[strtod, strtof, strtold - convert a string to a double-precision number]; The Open Group Base Specifications Issue 6; IEEE Std 1003.1, 2004 Edition]

trws:
Good points.



> What about zero? Maybe change to
> 
> > The value `N` MUST be a non-negative, non-infinite, floating-point number excluding `NaN`.

That sounds more right, I'll rework.

> I guess our implementation will allow forms like `3.1E10s` which is maybe fine. Could we say
> 
> > `N` SHALL be in one of the forms allowed by POSIX `strtof(3)`.
> > footnote:[https://pubs.opengroup.org/onlinepubs/009695399/functions/strtod.html[strtod, strtof, strtold - convert a string to a double-precision number]; The Open Group Base Specifications Issue 6; IEEE Std 1003.1, 2004 Edition]
> 
> Edit: changed strtod->strtof since we're not specifically calling out double precision.

I'd prefer to reference the C standard either c89 `strtod` or c99 `strtof` rather than posix, but completely agree on linking to the relevant standard.

trws:
Updated version pushed.  I went with the C99 definition for now, if we want to require POSIX extensions, we can switch it, but as far as I know we don't explicitly support locales so I'm not sure it's relevant.

garlick:
Dict update needed:
```
spec_23.adoc: IEC
spec_23.adoc: strtod
spec_23.adoc: strtof
spec_23.adoc: strtold
```
o/w looks good.

garlick:
Thanks @trws!

Issue #192, opened 2019-07-31
need RFC for stdio representation in KVS


garlick:
The stdin, stdout, and stderr of a job should be part of the job's permanent record in the KVS by default.  Since output may be parsed long after it was written, the representation should be documented.  The representation should have the following characteristics:
* The KVS value is a sequence of chunks
* Chunks may be appended, with intermediate values still a valid sequence
* A chunk may contain binary data (0x00 - 0xff)
* Chunk metadata includes _timestamp_, _task rank_, and _stream_ id

garlick:
Pondering this a bit this morning, I was reminded of flux-framework/flux-core#932 (use UTF-8 encoding), which I think we dismissed as wrong headed due to the fact that this would communicate that the output is "text" when it might not be.  (Sadly the issue wasn't updated after we talked about it, so maybe I've forgotten where that ended up)

Should an encoding include the equivalent of a `content-type` field, with a default of "text" and a user-controlled option for "octet-stream" or similar?  That would let us represent text directly as a JSON string, including UTF-8 code points if used.  Avoiding base64 for text would be more efficient and less obfuscated for future alien decoding.  Yet if a user wants to use stdio for non-textual data, they could do that by providing an option to the shell, and then aliens would know whether the bytes are human readable or not...

For actual implementation, the shell will encounter a jansson error if it tries to encode invalid UTF-8 as a string.  Would this be fatal to the job?

This implies a "header" portion in the KVS representation.  Hmm, even if I'm totally off base with imposing an interpretation on the data, maybe a header is still a good idea, as it would allow us to version the encoding without repeating it in each blob...

grondo:
Yeah, that seems to make sense. Realistically, we can't expect users to provide the content-type of their stdio for jobs, and we'd have to protect against jobs inadvertently sending binary data as output as you hinted above (the equivalent of accidentally cat-ing a binary file to your terminal). However, the pty option could probably turn on the correct encoding, which is the main case we're worried about now.

I guess the main question is can we get it done in the time allotted for this work? If not, then we have to decide if the benefit is worth pushing back the deadline. Whereas, we already have several working base64 based solutions since it is simple, and we could update to the superior solution down the road after we have something working.

garlick:
[...]

PR #194, opened 2019-08-12
rfc24: add new RFC for KVS stdio format

garlick:
Here is a first go at defining a KVS eventlog format for standard I/O as discussed in #192.

### DIFF
--- a/spec_24.adoc
+++ b/spec_24.adoc
@@ -1,0 +1,159 @@
+ifdef::env-github[:outfilesuffix: .adoc]
+
+24/Flux Job Standard I/O Version 1
+==================================
+
+This specification describes the format used to represent
+standard I/O streams in the Flux KVS.
+
+* Name: github.com/flux-framework/rfc/spec_24.adoc
+* Editor: Jim Garlick <garlick.jim@gmail.com>
+* State: raw
+
+== Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to
+be interpreted as described in http://tools.ietf.org/html/rfc2119[RFC 2119].
+
+== Related Standards
+
+*  link:spec_16{outfilesuffix}[16/KVS Job Schema]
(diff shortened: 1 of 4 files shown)


spec_24.adoc
```diff
@@ -0,0 +1,193 @@
...
+
+The following keys are OPTIONAL in the event context object:
+
+data::
+(string) The input data, encoded as described by the header.
+
+eof::
+(boolean) End of stream indicator.
+
+An input event SHOULD contain a `data` and/or `eof` key.
```
grondo: I think we already discussed this, but I can't remember the conclusion. How would we direct input to a specific rank? Would this be in the `options` field of the header eventlog entry for `guest.input`? Or should we allow a `rank` key in input events. (I think either way would work well)

(Or is this detail maintained completely outside of the `guest.input` eventlog format?)

garlick: I don't think we reached a conclusion, and my intent was to leave that open at this point, to be filled in when we get further along designing input modes.  Maybe I should add a note to that effect in both the options section and the input event description?

grondo: One thought is that solely using the exec.input header options to describe input mode would constrain stream names to be known and static at the time the `exec.input` header event is generated. Whereas, If a `rank` field was included in input events, a job could start with stdin of the front end tool directed into the event log for only rank 0, and later something else could come along and write some other stream to rank N or the remaining ranks.

I guess another way to put it - why do input events need a different format than output events?

garlick: > One thought is that solely using the exec.input header options to describe input mode would constrain stream names to be known and static at the time the exec.input header event is generated. 

It's constrained anyway since the encoding type for each stream is in the header, so if want it unconstrained, we would need to do something different like include encoding type in each input event (or maybe just the first one for the stream?)

I don't have a strong opinion on this one...

garlick:
Just pushed some changes per our offline discussion
* define four streams: stdin, stdout, stderr, stdlog
* retain separation of input and output keys
* merge input and output events into a "data" event, so the same format is used for both input and output, and input can specify a task idset

garlick:
Incremental changes squashed.

grondo:
Thanks, LGTM!

PR #195, opened 2019-08-14
rfc24: drop footer event

garlick:
While switching the `guest.output` from the ad-hoc JSON array format I used earlier to the RFC 24 eventlog format, I realized the `footer` event is a pain to generate, unnecessarily constrains the eventlog to single writer (or requiring synchronization among writers), and isn't especially useful.

Drop `footer` event from the spec.

Also: tidy up a list of stream names.


grondo:
Looks good!

PR #196, opened 2019-08-14
rfc24: add log event

garlick:
Add a new event type called "log" that is intended to capture flux-shell log messages as part of the job's output key.  Drop the "stdlog" stream, which was the first attempt at enabling logging, as log messages don't necessarily follow stream semantics.

### DIFF
--- a/spec_24.adoc
+++ b/spec_24.adoc
@@ -36,7 +36,7 @@ be interpreted as described in http://tools.ietf.org/html/rfc2119[RFC 2119].
 * Input and output should be labeled with a timestamp.
 * Each stream may independently indicate end-of-stream.
 * Support standard UNIX I/O streams: `stdin`, `stdout`, `stderr`.
-* Support an additional output stream called `stdlog`.
+* Support shell debug logging.
 
 == Implementation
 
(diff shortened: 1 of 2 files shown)


spec_24.adoc
```diff
@@ -36,7 +36,7 @@ be interpreted as described in http://tools.ietf.org/html/rfc2119[RFC 2119].
 * Input and output should be labeled with a timestamp.
 * Each stream may independently indicate end-of-stream.
 * Support standard UNIX I/O streams: `stdin`, `stdout`, `stderr`.
-* Support an additional output stream called `stdlog`.
+* Support shell debug logging.
```
grondo: Maybe clearer: "Support shell error and debug logging."

spec_24.adoc
```diff
@@ -126,3 +125,13 @@ Example:
   }
 }
 ----
+
+=== Log Event
+
+The log event supports debug logging from the Flux shells.
```
grondo: Maybe just semantics, but should this be debug and error logging?

Just throwing this out here, and I don't think it is really needed, but should a log event also include a log level or severity?

garlick:
Crap! I set merge when passing with a fixup! pending.  Tried to remove it but too late!  Oh well.

PR #199, opened 2019-08-22
rfc24: Add stream counts to header

chu11:
in order to determine that the output is complete, we need to know how many EOFs to expect, so we need to know the count of streams will send an EOF.

Multiple ways to do this of course.  I thought that perhaps a "count" object with a count of individual streams inside it was best.


spec_24.adoc
```diff
@@ -59,6 +59,10 @@ encoding::
 (object) A dictionary mapping stream names (e.g. `stdin`, `stdout`, `stderr`)
 to encoding type (e.g. `base64`, `UTF-8`).
 
+count::
+(object) A dictionary mapping stream names (e.g. `stdout`, `stderr`) to a count of
+the number of streams that for that stream.
```
grondo: Maybe: "A dictionary mapping stream names (e.g. `stdout`, `stderr`) to the number of expected open streams with that name"

chu11:
re-pushed with tweak

chu11:
> The idea of adding the footer back to the output eventlog was also discussed yesterday when we chatted. Your addition here seems ok to me, but I'll let @garlick comment as well.

Only thinking about this for 5 seconds, the footer seems to make things easier.  And possibly safer?  Could there be scenarios (e.g. shell crashing) where an EOF won't be logged?  But the footer will be logged?

garlick:
The reason I removed the footer was it seemed to close the door on each shell updating the eventlog independently, rather than routing through the leader shell, or forcing the shells to execute a barrier on exit.

Should we just have job-info send an ENODATA response after the last event if it knows an eventlog under `guest` can no longer change?  Or similarly if the primary eventlog reaches the `clean` event?

chu11:
>The reason I removed the footer was it seemed to close the door on each shell updating the eventlog independently, rather than routing through the leader shell, or forcing the shells to execute a barrier on exit.

Good point.

> Should we just have job-info send an ENODATA response after the last event if it knows an eventlog under guest can no longer change?  Or similarly if the primary eventlog reaches the clean event?

Initial thought, it's seems a pretty decent idea and feasible to implement.  Perhaps better than hang / wait forever when an event will not occur?

Just thinking about this and implementation:

- When watching the main eventlog, if we receive the "clean" event, return ENODATA to the caller.  Easy to implement.

- if watching an eventlog in the guest namespace, we can assume end of the eventlog if the guest namespace receives a ENOTSUP error (namespace no longer exists) and can return ENODATA to the caller.

- if the guest namespace has been removed and the eventlog has already been moved to the main namespace, we don't need to do a "watch" on the eventlog.  We can just do a lookup on the eventlog, stream the data to the caller, and return ENODATA at the end.

- `flux job wait-event` and similar calls have to handle the error (e.g. "event FOOBAR never occurred") if the event never occurred and they receive ENODATA.

PR #205, opened 2019-09-20
rfc18: establish some basic eventlog update rules

garlick:
It came up in flux-framework/flux-core#2374 that some basic rules of engagement for eventlogs had not been documented.  So this adds a little bit of clarification.


grondo:
This seems like a good clarification.

PR #206, opened 2019-09-27
rfc24: Add redirect event

chu11:
Per discussion in https://github.com/flux-framework/flux-core/pull/2395 and https://github.com/flux-framework/flux-core/issues/2399

### DIFF
--- a/spec_24.adoc
+++ b/spec_24.adoc
@@ -134,6 +134,42 @@ Example:
 }
 ----
 
+=== Redirect Event
+
+The redirect event indicates that a stream's data has been redirected
+away from the log.  The caller should not expect any additional data
+events in the log for that stream.
+
+The following keys are REQUIRED in the event context object:
+
+stream::
+(string) The stream name (e.g. `stdout`, `stderr`).  All valid stream
+names MUST appear as keys in the header `encoding` object.
+
+rank::
+(string) An idset string (RFC 22) representing the rank(s) that are
+redirecting output.
+
+The following keys are OPTIONAL in the event context object:
+
+path::
+(string) Indicates the path data has been redirected to, if the data
+has been redirected to a file.
+
+Example:
+
+[source,json]
+----
+{
+  "timestamp":1552593350.4,
+  "name":"redirect",
+  "context":{
+    "stream":"stdout",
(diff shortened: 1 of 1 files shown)


spec_24.adoc
```diff
@@ -134,6 +134,38 @@ Example:
...
+away from the log.  The caller should not expect any additional data
+events in the log for that stream.
+
+The following keys are REQUIRED in the event context object:
+
+stream::
+(string) The stream name (e.g. `stdout`, `stderr`).  All valid stream
+names MUST appear as keys in the header `encoding` object.
+
+The following keys are OPTIONAL in the event context object:
```
grondo: Just a thought, we may want to add an optional `rank` key to indicate the tasks to which the redirect applies. This would allow some tasks to be redirected and not others in the future. (e.g. redirect all tasks except rank 0 to `/dev/null`)

chu11: yeah, I went back and forth on that.  Now that I think about it, I suppose if the redirect applies to all ranks, the rank be == `all`?  So we should make the rank a requirement.

chu11: oh wait, ranks are idset strings.   Hmmm.  Should we allow idset string or "all"?

grondo: Would it be too difficult to generate the correct idset in the shell output plugin?

chu11: Nah, that shouldn't be too hard.

I was mostly thinking about how to recognize that all tasks have been redirected, but we can can count idsets just like EOFs and we'd be fine.

I was also hoping to output a single message if both stdout/stderr were redirected to the same file, but I suppose that can be two messages.

chu11:
re-pushed adding ranks

chu11:
re-pushed with different filename in example, to pass spellcheck

spec_24.adoc
grondo: > I was also hoping to output a single message if both stdout/stderr were redirected to the same file, but I suppose that can be two messages.

Ah, that is a good point. Maybe stream could be a list? Eh, maybe two events is ok like you say.

grondo:
Ok, merging.

PR #208, opened 2019-10-03
rfc26: extend canonical jobspec to support job dependencies

### LINKED ISSUE #209: ci: validate jobspec examples with their jsonschemas
Any opposition to copying `validate.py` over into this repo and running the various jobspec examples through their matching jsonschemas?  I've been doing it manually (when I remember to), but I think it would be good for Travis to verify this.

SteVwonder:
Adds the OpenMP-inspired `in`, `out`, and `inout` keys as valid
dependency types.

Based on discussion in
https://github.com/flux-framework/flux-core/issues/1698

Closes #209

### DIFF
--- a/data/spec_14/use_case_2.8.yaml
+++ b/data/spec_14/use_case_2.8.yaml
@@ -1,0 +1,30 @@
+version: 999
+resources:
+  - type: slot
+    count: 1
+    label: default
+    with:
+      - type: node
+        count: 1
+tasks:
+  - command: [ "flux", "start" ]
+    slot: default
+    count:
+      per_slot: 1
+attributes:
+  system:
+    duration: 3600.
+    cwd: "/home/flux"
+    dependencies:
+      - type: in
+        scope: user
+        scheme: fluid
+        value: hungry-hippo-white-elephant
+      - type: in
+        scope: user
+        scheme: string
+        value: foo
+      - type: out
+        scope: user
+        scheme: string
+        value: bar
--- a/spec_14.adoc
+++ b/spec_14.adoc
@@ -25,6 +25,7 @@ be interpreted as described in http://tools.ietf.org/html/rfc2119[RFC 2119].
 * link:spec_4{outfilesuffix}[4/Flux Resource Model]
 * link:spec_8{outfilesuffix}[8/Flux Task and Program Execution Services]
 * link:spec_20{outfilesuffix}[20/Resource Set Specification Version 1]
+* link:spec_26{outfilesuffix}[26/Job Dependency Specification]
 
 == Goals
 
@@ -297,6 +298,9 @@ Some common system attributes are:
  *cwd*::
  The value of the `cwd` attribute is a string containing the absolute
  path to the current working directory to use when spawning the task.
+
+ *dependencies*:: The value of the `dependencies` attribute SHALL be a
+ list of dictionaries following the format specified in RFC 26.
 
  *job*::
  The `job` attribute is an optional dictionary containing job
--- a/spec_14.adoc
+++ b/spec_14.adoc
@@ -625,6 +625,23 @@ overridden environment (i.e., unset FOO and set BAR=2).
 Jobspec YAML::
 +
 [source,yaml]
+
 ----
 include::data/spec_14/use_case_2.7.yaml[]
+----
+
+'''
+Use Case 2.8:: Specify dependencies
++
+Specific Example:: Depend on two previously submitted jobs.  The first job's
+Flux ID (fluid) is known (`hungry-hippo-white-elephant`).  The second job's
+fluid is not known but its `out` dependency (`foo`) is known.  Also provide an
+`out` dependency (`bar`) that other jobs can depend on.
++
+Jobspec YAML::
++
+[source,yaml]
+
+----
+include::data/spec_14/use_case_2.8.yaml[]
 ----
--- a/spec_26.adoc
+++ b/spec_26.adoc
@@ -1,0 +1,171 @@
+ifdef::env-github[:outfilesuffix: .adoc]
+
+26/Job Dependency Specification
+==============================
+
+An extension to the canonical jobspec designed to express the dependencies
+between one or more programs submitted to a Flux instance for execution.
+
+* Name: github.com/flux-framework/rfc/spec_26.adoc
+* Editor: Stephen Herbein <herbein1@llnl.gov>
+* State: raw
+
+== Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to
+be interpreted as described in http://tools.ietf.org/html/rfc2119[RFC 2119].
+
+== Related Standards
+
+* link:spec_14{outfilesuffix}[14/Canonical Job Specification]
+* link:spec_19{outfilesuffix}[19/Flux Locally Unique ID]
+* link:https://www.openmp.org/wp-content/uploads/OpenMP-API-Specification-5.0.pdf[OpenMP Specification]
+* link:https://tools.ietf.org/html/rfc3986[IETF RFC3986: Uniform Resource Identifier (URI)]
+
+== Goals
+
+* Express the dependencies between jobs to the dependency service.
+* Provide a mechanism for specifying dependencies as a directed acyclic graph
+  (DAG).
+* Provide a mechanism for specifying more advanced, runtime dependencies.
+
+== Job Dependency Definition
+
+A dependency SHALL be a dictionary containing the following keys (whose
+definitions are detailed in the sections below):
+
+* *type*
+* *scope*
+* *scheme*
+* *value*
+
+=== Type
+
+The value of the `type` key SHALL be one of the following (the semantics of
+which are inspired by the OpenMP specification):
+
+ * *out* This key only affects future submitted jobs. If the value of this key
+         is the same as the value in an `in` or `inout` dependency of a future
+         job within scope, then the future job will be a dependent job of the
+         current generated job.
+
+ * *in* If the value of this key is the same as the value in an `out` or `inout`
+        dependency of a previously submitted job within scope, then the
+        generated job will be a dependent job of that previous job.
+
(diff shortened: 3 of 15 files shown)


SteVwonder:
@frankd412, @dmdom, and @koning: I believe this is relevant to your workflows.  Please let us know if you have any thoughts on this strawman for how job dependency will be specified in Flux's jobspec.

spec_14.adoc
```diff
@@ -298,6 +299,37 @@ Some common system attributes are:
...
+        at least one of the list items is the same as the value of a list item
+        in a `dependencies.out` or `dependencies.inout` key of a previously
+        submitted job, then the generated job will be a dependent job of that
+        previous job.
+
+ * *out* The value of the `out` key SHALL be a list of strings.  If the value of
+         at least one of the list items is the same as the value of a list item
+         in a `dependencies.in` or `dependencies.inout` key
+         of a previously submitted job, then the generated job will be a
+         dependent job of that previous job.
```
trws: I'm reading this as "out jobs will depend on in jobs" is that intentional?

spec_14.adoc
```diff
@@ -298,6 +299,37 @@ Some common system attributes are:
...
  path to the current working directory to use when spawning the task.
 
+ *dependencies*:: The value of the `dependencies` attribute SHALL be a
+ dictionary that SHALL contain the following keys (whose semantics are inspired
+ by the OpenMP specification):
+
+ * *in* The value of the `in` key SHALL be a list of strings.  If the value of
+        at least one of the list items is the same as the value of a list item
+        in a `dependencies.out` or `dependencies.inout` key of a previously
+        submitted job, then the generated job will be a dependent job of that
```
trws: Does "a previously submitted job" cross jobs for all users of the instance, or is it specific to the current user?  Whichever it is, is there a way for a user to request the other?

spec_14.adoc
SteVwonder: > is that intentional?

Good catch.  No, it's a copy-and-paste carry over from the `inout` verbiage.

That is the behavior in OpenMP though.  I'm guessing for the same reason that you said `out`'s depend on other `out`s: an assumed data dependency?  Assuming so, I'll go ahead and remove it.

spec_14.adoc
SteVwonder: Great question.  Symbol collision across user's could get quite hairy. fluids are assumed to be globally unique, so I don't think the problem exists for them.

For abstract symbols (and other URIs with a high probability of collision across users), I'm thinking one of the following:
1) Let everything be global and if the user's want their own sandbox/namespace, use a sub-instance.
2) Leverage URIs for this
3) Add something else to the jobspec

1 is obviously the least work for us, but I feel like it is artificially restrictive.  Really bad for the ATS and UQP workflows that want to submit a bunch of large jobs that each have a single core, dependent job to do post-processing.  Those don't pack into a nested instance too nicely.

Strawman example for 2: `string://stephen@foo/bar` would refer to the symbol `bar` in `stephen`'s namespace in flux instance `foo`. `string:foo` could default to the submitting user's namespace in the enclosing instance.  `string://global@foo/bar`  would refer to the symbol `bar` in the `global` (or whatever name we choose to reserve for this) namespace of `foo`.  This breaks down somewhat since Flux instances are currently path'd by URIs as well.  Nested URIs would be kinda :stuck_out_tongue_closed_eyes:.  Maybe another reason to move forward with our `where-the-flux` service ;)

For 3, maybe we smush another dictionary in there with `global` and `user` as keys. For example:
```yaml
dependencies: 
  in: 
    global: 
      - string:bar
      - fluid:hungry-hippo-white-elephant
    user: "string:foo"
  out: 
    user: "string:bar"
```

trws:
Essentially yes. Because they were conceived as data dependencies there is an assumption of mutual exclusion between any tasks that have out on a given dependency and all other dependence types with that dependence. If we are using them as execution dependencies, we don’t need that property by default and could add it optionally if desired, which seems more natural to me for this kind of system.

Sent from my iPhone

On Oct 3, 2019, at 8:26 PM, Stephen Herbein <notifications@github.com> wrote:

﻿

@SteVwonder commented on this pull request.

In spec_14.adoc<https://github.com/flux-framework/rfc/pull/208#discussion_r331329713>:

> + * *out* The value of the `out` key SHALL be a list of strings.  If the value of
+         at least one of the list items is the same as the value of a list item
+         in a `dependencies.in` or `dependencies.inout` key
+         of a previously submitted job, then the generated job will be a
+         dependent job of that previous job.


is that intentional?

Good catch. No, it's a copy-and-paste carry over from the inout verbiage.

That is the behavior in OpenMP though. I'm guessing for the same reason that you said out's depend on other outs: an assumed data dependency? Assuming so, I'll go ahead and remove it.

spec_14.adoc
SteVwonder: Quick brain dump of everything @trws and I just discussed:

- We can use URIs as shorthand, but we could also "explode" them into a dictionary
- The dependence type should be a key within that dictionary (along with scope)
- The default scope should be `user`
- Adding a `system` scope that users can only add `in` dependencies on for DATs and downtime
- Add text saying we plan on adding other types like `inoutset`, `runtime` and `all`
- Correcting the `inout` definition to be the same as `out`
- All of these dependencies (even the `runtime`) should be "push" not "pull" (i.e., no polling)
  - We could still support `file://path` by treating it like a symbol (you need a job that has an `out` dependence and another that has an `in` dependence).  We could also check in between those jobs if the file actually exists and throw an exception if not
- We can add `exit-code` or `state` as an additional key in the dictionary to get parity with Slurm and other scheduler's dependency models

SteVwonder:
I believe this is ready for another review @trws.

I'm still not 100% sure I have the right definition for `out`.  Does adding an `out` dependency ever make the job depend on a previous job (i.e., one with a matching `inout`)?  When I think about it in terms of matching `in` dependencies up with `out` dependencies, it doesn't make sense for it to create a dependency.  When I think about it in terms of using `inout` as a mutual exclusion type it does make sense for it to create a dependency.  I think that latter idea is wrong though, so I went with the former definition for now.  Let me know if you disagree.

trws:
I think we need to decide whether it does or not.  In OpenMP `out` creates exclusion with both `in` and `inout` in both directions, but I think for flux we were discussing using out as a unidirectional dependency.  An out -> in clearly creates a dependency, but if we want mutual exclusion then outs can't run concurrently with each other or with inouts or with ins.

SteVwonder:
> I think for flux we were discussing using out as a unidirectional dependency. An out -> in clearly creates a dependency, but if we want mutual exclusion then outs can't run concurrently with each other or with inouts or with ins.

After letting it sit for a few days, I agree on the "unidirectional dependency".  I think mutual exclusion should be reversed for `inout`.  I believe the current definition is what we want.

SteVwonder:
@trws (and anyone else interested in reviewing), I believe this is ready for another review.

spec_26.adoc
```diff
@@ -0,0 +1,171 @@
...
+
+The value of the `scope` key SHALL be one of the following:
+
+ * *user* All jobs previously submitted by the user are contained within
+    this scope.  A user can create a dependency of any type within this scope.
+
+ * *system* All jobs previously submitted by the instance owner are contained
+    within this scope.  The instance owner can create a dependency of any type
+    within this scope. A non-instance owner can only create a dependency with
+    the type `in` within this scope.
```
trws: This is the only thing that jumps out at me as a remaining question.  In a top-level instance, can one user depend on a job submitted by another user? It seems like the system level only contains jobs from the owner, so not other users, and user would be for the current user unless I misunderstand?

SteVwonder: > It seems like the system level only contains jobs from the owner, so not other users, and user would be for the current user unless I misunderstand?

Correct.

>  In a top-level instance, can one user depend on a job submitted by another user?

Not in this design.    I've been noodling on this, but I'm not sure of the best way to do this without breaking the URI shorthand.  Is this a common use-case on our (or other) systems?

data/spec_14/use_case_2.8.yaml
```diff
@@ -0,0 +1,30 @@
...
+      per_slot: 1
+attributes:
+  system:
+    duration: 3600.
+    cwd: "/home/flux"
+    dependencies:
+      - type: in
+        scope: user
+        scheme: fluid
+        value: hungy-hippo-white-elephant
```
grondo: Just happened to notice a typo here `s/hungy/hungry/`

spec_26.adoc
grondo: I'm not sure, but I was curious to see if slurm allowed setting a dependency on another user's job.
Answer: it seems to allow it. Not surprising since dependencies are based on jobids only.

SteVwonder: > edit: it seems you could eventually allow it without breaking uris by allowing the userid/username to be specified with the user scope e.g. user:grondo, or maybe by allowing an any scope, etc...

Yeah, that definitely seems like a viable option.  Now that you mention it, we could also do that in the `value` field: 

```yaml
      - type: in
        scope: user
        scheme: fluid
        value: grondo:hungy-hippo-white-elephant
```

trws: Fluids are logically global anyway right? Do they require further specification, or could they just specify the fluid, possibly with an option of another user to use as scope for a sanity check?

grondo: I was thinking the same thing. One nice thing about making it per-user though is that you could more easily catch typos in FLUIDs that might match the an incorrect global job. Unlikely, though, so always matching job ids in a global context is fine with me.

SteVwonder: All good points.  How does this sound?

- Switch the `system` label to be `global` (maybe this is unnecessary,  but it doesn't seem obvious to me that other users' jobs would end up in the `system` scope; `global` seems to capture that better).
- Instance owner can create `in` and `out` dependencies with `global` scope while non-owners can only create `in` dependencies with a `global` scope.
- By default all job's fluids are inserted into the `global` scope by the dependency module, allowing users to create `in` dependencies on other users' jobs.

SteVwonder: Just pushed my proposed changes above.

SteVwonder:
Just pushed changes based on the latest round of feedback and rebased on master.

grondo:
Setting merge-when-passing.

SteVwonder:
Thanks @grondo!

PR #210, opened 2019-10-14
rfc24: update log event context

grondo:
This PR proposes a couple minor changes to the RFC 24 `log` event context.

Probably uncontroversial -- optional `file` `line` and `component` keys are added to the context. Addition of these members to log messages may aid in post mortem debugging.
(In case it isn't clear, the `component` would be used to indicate a facility or more likely plugin name which generated the log message)

Perhaps more controversial, I removed the reference to RFC 5424 severity levels and renamed the field to `level` instead of `severity`. The syslog severity system seems more tailored to system logs and not simple process logging, for which we'd like a more straightforward logging system that goes from "Fatal" errors to "Trace" logs. (For the shell my current implementation uses `0=Fatal, Error, Warning, Info, Debug, 6=Trace`.

### DIFF
--- a/spec_24.adoc
+++ b/spec_24.adoc
@@ -174,12 +174,25 @@ Example:
 
 The log event supports error and debug logging from the Flux shells.
 
+The following keys are REQUIRED in the log event context object:
+
 rank::
 (integer) The shell rank.
 
-severity::
+level::
 (integer) An Internet RFC 5424 severity level in the range of 0 (LOG_EMERG)
 to 7 (LOG_DEBUG).
 
 message::
 (string) Textual log message, encoded with UTF-8.
+
+The following keys are OPTIONAL in the event context object:
+
+file::
+(string) Source file from which the log message was generated.
+
+line::
+(integer) Source line from which the log message was generated.
+
+component::
+(string) A shell component or plugin name which generated the log message.


chu11:
> we'd like a more straightforward logging system that goes from "Fatal" errors to "Trace" logs. (For the shell my current implementation uses 0=Fatal, Error, Warning, Info, Debug, 6=Trace.

Do we have these logging levels defined anywhere?  Or are they up to the user where they can put anything into "level" that they want?

grondo:
They'd be defined in `shell.h`

grondo:
I'd actually be fine with keeping the syslog defined severities, but some of them don't make sense in the context of user programs, and are counter to the more typical logging levels of user programs. However, we could just remap them internally, but then I thought why bother?

chu11:
I think level is fine, just thinking the RFC could be clearer stating `0=Fatal, Error, Warning, Info, Debug, 6=Trace`.  Or do you not want to be that strict, and the user can have some flexibility defining what level means?

garlick:
If the level is exposed to the user as an int, it might be good to stick with the RFC 5424 levels since it would be easy for someone familiar with other parts of the code base to assume the LOG_DEBUG or whatever is valid, and then not get the desired effect.  I guess that's easily fixed with an enum or by not exposing the levels in the functions/macros that generate logs.  OK, now that I write that, it seems like either way would be fine so I defer to the guy who is doing the work.

grondo:
> I think level is fine, just thinking the RFC could be clearer stating 0=Fatal, Error, Warning, Info, Debug, 6=Trace. Or do you not want to be that strict, and the user can have some flexibility defining what level means?

I was thinking 99% of cases would write a `log` event using a logging interface provided in `shell.h`, for which 99% of those cases are going to use wrappers like `shell_trace()` `shell_debug()` `shell_info()` etc., which encode the level.

It probably doesn't make sense to have other use cases have another definition of logging level, so if we went this route I think you are right and we should define the levels in the RFC (at the time it felt like exposing a little too much of the shell internals)


@garlick makes a good point about accidential use of `LOG_DEBUG`. I'm not 100% sure where this would come up, but one mitigation would be to just remap the RFC 5424 severities to shell logging levels (hopefully without the need to include `syslog.h`). i.e. add a never used logging level for LOG_CRIT or something.

Or, we could just keep the RFC 5424 severit. If you do that you still have to define what they actually mean in terms of shell logging levels so probably still some additional text needed in the rfc.

garlick:
> I was thinking 99% of cases would write a log event using a logging interface provided in shell.h, for which 99% of those cases are going to use wrappers like shell_trace() shell_debug() shell_info() etc., which encode the level.

That allays my concern about user surprise, so I'm fine with departing from the syslog levels here if its helpful or more clear to do so internally.

grondo:
> That allays my concern about user surprise, so I'm fine with departing from the syslog levels here if its helpful or more clear to do so internally.

There is a potential use case for agents external to the job shell writing a "log" event to the output eventlog, and that is where I think your worry makes sense. Especially if say, a scheduler wanted to log a message about preempting a job to that job's output (`Congratulations, this standby job has been selected for immediate termination!`) -- in that context the RFC 5424 severity levels and "shell" levels could easily be mixed up.

grondo:
Ok, I reinstated the log level requirement to use RFC 5424 severity level.

For the common case the shell interface will make this invisible to the "user" (shell components writing log messages). The shell interface will likely remap the severity levels something like:

| RFC 5424 Level | Shell Level | Level string |
------------------------ | ----------------- | ------------------
  `LOG_EMERG` | `SHELL_LOG_FATAL` | "FATAL"
  `LOG_ALERT` | `SHELL_LOG_ERROR` | "ERROR"
  `LOG_CRIT` | `SHELL_LOG_ERROR` | "ERROR"
  `LOG_ERR`  |  `SHELL_LOG_ERROR` | "ERROR"
  `LOG_WARNING` | `SHELL_LOG_WARN` | "WARN"
  `LOG_NOTICE` | `SHELL_LOG_NOTICE` | ""
  `LOG_INFO` |  `SHELL_LOG_INFO` | "INFO"
  `LOG_DEBUG` | `SHELL_LOG_DEBUG` | "DEBUG"


Slightly different than common practice, but it should work ok. The default level is `SHELL_LOG_NOTICE`, and increased verbosity gets you "INFO" (i.e. verbose) messages and then "DEBUG" messages.

Unfortunately, to suppress warnings and errors, you would have to give the equivalent of 4 `--quiets` - though perhaps this is not a use case anyway.

grondo:
Maybe it makes more sense to make `LOG_EMERG` == `LOG_ALERT` == `LOG_CRIT` == "Fatal" for the shell. External loggers shouldn't be using these severity levels anyway, and this removes the "three levels of nonfatal errors" problem from above.

Internally the shell will log fatal when the error actually is fatal.

spec_24.adoc
```diff
@@ -174,12 +174,25 @@ Example:
...
 
+The following keys are REQUIRED in the log event context object:
+
 rank::
 (integer) The shell rank.
 
-severity::
+level::
 (integer) An Internet RFC 5424 severity level in the range of 0 (LOG_EMERG)
 to 7 (LOG_DEBUG).
```
garlick: Since RFC 5424 remains, the commit message needs to be updated:
> * RFC 5424 severity levels don't make much sense for program
   logging levels. Relax this requirement and just use log "level"
   from 0 (Fatal) to 6 (Trace).

grondo: Oh, thanks for catching that.

PR #211, opened 2019-10-24
rfc24: log event context updates

grondo:
As discussed in flux-framework/flux-core#2477, change the `rank` key in RFC 24 log event context from REQUIRED to OPTIONAL. If  the `rank` key is missing it is assumed the shell rank is unknown at the time of the log message.

This PR also adds another optional key: `program`, which is assumed to be `flux-shell` when not present, but could be used in the future if outside entities write log messages to a job output event log. It is assumed the that `program` value will be used as the prefix for log messages in display programs like `flux-job attach`.

### DIFF
--- a/spec_24.adoc
+++ b/spec_24.adoc
@@ -177,9 +177,6 @@ The log event supports error and debug logging from the Flux shells.
 
 The following keys are REQUIRED in the log event context object:
 
-rank::
-(integer) The shell rank.
-
 level::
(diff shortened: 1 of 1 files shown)


chu11:
LGTM except for the typo "unkown"

grondo:
Ack! Sorry about that. Fixed, squashed and pushed the result.

grondo:
Thanks!

PR #214, opened 2019-11-27
rfc14: Add system.job-name field description

chu11:
My preliminary proposal for a `job-name` field in the jobspec.

### DIFF
--- a/spec_14.adoc
+++ b/spec_14.adoc
@@ -298,6 +298,18 @@ Some common system attributes are:
  The value of the `cwd` attribute is a string containing the absolute
  path to the current working directory to use when spawning the task.
 
+ *job*::
+ The `job` attribute is an optional dictionary containing job
+ metadata.  This metadata may be used for searching and filtering of
+ jobs.  Every `value` in the dictionary must be a string.  The
+ application is free to create keys of any name, however the following
+ are reserved for special use:
+
+   *name*:::
+   The `name` key contains the name of the job.  The default name of a
+   job is the first argument of the command run by the user, or it can be
+   set by the user to an arbitrary value.
+
 === Example Jobspec
 
 Under the description above, the following is an example of a fully compliant


grondo:
In flux-framework/flux-core#2401, I had proposed reserving a dictionary at `attributes.system.job.` for generic job attributes such as `name`. Eventually tools could support display or filtering on any of these attributes.

Reading the edits to the RFC though, a single `job-name` attribute does seem to fit better with some of the existing attributes `cwd`, `environment`, but what do you think about `system.job.name` vs `system.job-name`? It is almost irrelevant at this point, but may allow for easier extending in the future.

chu11:
> what do you think about system.job.name vs system.job-name

My initial feeling was that the "job-name" is sort of a special case, something expected by the system, where as the arbitrary metadata in `system.job` is, well of arbitrary?

Now that I think about it, what would be the difference between having users put metadata in `system.job` vs. `user`?  Isn't `user` where arbitrary user metadata would go?

grondo:
> Now that I think about it, what would be the difference between having users put metadata in system.job vs. user? Isn't user where arbitrary user metadata would go?

I think we said before that `user` is for use by the user and/or application, Flux components wouldn't even look there. The attributes in `system.attributes.job.` may be available for filtering so would probably have to conform to some specification, e.g. may only be of type string, number, or boolean.

chu11:
>  I guess my question more specifically is do you have an objection to job.name vs job-name?

No objection on my part.  I'll re-work.

grondo:
fyi - Some other attributes we might need under `system.job.` that come to mind are `job.bank` `job.wckey`, etc, which might be used by the bank & accounting system.

chu11:
re-worked doing the `system.job.name` approach.

chu11:
> fyi - Some other attributes we might need under system.job. that come to mind are job.bank job.wckey, etc, which might be used by the bank & accounting system.

Ahh, that makes the need far clearer.

chu11:
> Just one nit: the job dictionary probably should be called out as optional, since the section it's in leaves that somewhat ambiguous. Then obviously any keys within it are optional.

Ahhh, I tweaked by changing "The value of the `job` attribute is a dictionary containing additional job metadata. " to "The `job` attribute is an optional dictionary containing job metadata."

garlick:
Thanks!  I'll go ahead and set merge-when-passing since you addressed @grondo's comment earlier.

Issue #213, opened 2019-11-26
RFC 14: clarify and/or simplifiy command specification 


chu11:
When reading the definition of a command:

`The value of the command key SHALL be a string OR list representing an executable and its arguments.`

I originally read this as "The command key SHALL be string OR list, each of which could be representing an executable and its arguments".  Presumably arguments in the string are separated by spaces.

but, seeing some code in `flux-core/shell`, it interprets this sentence more like: "the command key SHALL be a string with a single command OR a list representing an executable and its arguments".  In other words, commands with arguments must be in list form, strings are arrays of length one.

It need not matter, but it should be clarified which is the correct interpretation with code adjusted accordingly.

OR

perhaps things could be simplified.  Perhaps the command should always be a list?  As far as I can tell, it is only used as a list in the current `flux-core`.  Some examples would have to be updated as a result.

grondo:
Currently a string is treated as a list of one entry.

It would probably simplify the parsing code if the `command` could always be assumed to be a list, passed directly to exec.

chu11:
> It would probably simplify the parsing code if the command could always be assumed to be a list, passed directly to exec.

Sounds like a good idea to me.

PR #215, opened 2019-12-02
rfc14: Update commands to always be a list

chu11:
Per discussion offline and in issue #213, update commands to always be a list, instead of possibly a list or string.

Update schema and examples accordingly.

### DIFF
--- a/data/spec_14/example1.yaml
+++ b/data/spec_14/example1.yaml
@@ -10,7 +10,7 @@ resources:
           - type: core
             count: 2
 tasks:
-  - command: app
+  - command: [ "app" ]
     slot: default
     count:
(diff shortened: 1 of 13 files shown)


grondo:
This seems like an improvement to me, but I think we should get consensus, especially from @garlick, @SteVwonder and @trws.

grondo:
> Shall we interpret @SteVwonder's thumbs ups in #213 as approval?

Ah I had missed that. Yeah, let's do it!

trws:
FWIW I'm totally good with that for canonical, might want to do the fallback thing for initial user input for convenience but once it's in the system it may as well be simple.

PR #219, opened 2020-01-08
rfc6: drop group RPC section

garlick:
The mrpc class is deprecated in flux-framework/flux-core#2612,
so we no longer need group matchtag bits.  It was a hack anyway.


grondo:
Seems reasonable. Good to merge?

garlick:
Sure!

PR #227, opened 2020-02-22
Rendering of Readme

gonsie:
This PR changes the title at the of README so that it will look better in the RTD website. 

The index.rst file is used by sphinx to make sure each page appears somewhere. We make the TOC tree hidden since we’ll use the README to list each sub-page for this repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
-rfc
-===
+RFC Index
+=========
 
 This is the Flux RFC project.
 
--- a/index.rst
+++ b/index.rst
@@ -7,8 +7,7 @@ Flux RFC Index
 ==============
 
 .. toctree::
-   :maxdepth: 1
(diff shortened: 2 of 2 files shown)


SteVwonder:
@gonsie: which of your docs-1 branches is it again that has the RFC repo pulled in?  I'd like to give this a test locally with that flux-docs branch (after updating the rfc submodule to point to this PR).

PS - Unrelated and random, but it looks github does not automatically redirect forks after a rename (i.e., gonsie/flux-docs.git doesn't exist).

gonsie:
`rfc-integration` is the branch you want. There are still a few fixes before it can be megered... maybe I’ll try the wip PR title.

GitHub remembers renames, people can still push to flux-framework/docs without issue. Changing existing fork names requires work.

PR #239, opened 2020-04-30
pip requirements file

gonsie:
This is needed to setup readthedocs. We could name the file something else (maybe `.requirements.txt` or `.github/requirements.txt` if we want to keep the top-level for the repo from having too many files.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,6 @@
+sphinx
+sphinx-rtd-theme
+sphinxcontrib-spelling
+recommonmark
+pyyaml
+jsonschema


grondo:
I'm fine with a `requirements.txt` file in the top level.
I'll try this out since I need to figure out working with this repo anyway.

gonsie:
It can be used with the command `pip install -r requirements.txt` (you may need to use `pip3`)

grondo:
Great, thanks! Worked for me. And if it works for me it will work for anyone!

grondo:
Ok to set merge-when-passing?

gonsie:
Yes please!

grondo:
Thank you!

PR #244, opened 2020-05-05
Index readme

### LINKED ISSUE #234: README: add build instructions
With python virtualenv:
```
virtualenv -p python3 sphinx-rtd 
source sphinx-rtd/bin/activate
pip install sphinx sphinxcontrib-spelling recommonmark sphinx_rtd_theme
git clone git@github.com:flux-framework/rfc
cd rfc
make html
```

Credit goes to @gonsie

### LINKED ISSUE #242: Readme Links to specs
(This was uncovered by @garlick ‘s comment on flux-framework/flux-docs#38).

Right now the README file in this repo is doing 2 things:
1. It is used on github to index and link directly to the spec files (users can click and be directed to the correct file)
2. It is rendered to readthedocs (via the recommonmark markdown parser) as a main page.

Unfortunately there is an issue with the links to the spec files. For the github linking to work, they need to be the full file name (e.g., spec_1.rst), but this doesn’t get parsed correctly when the HTML files are built (links point to .rst files when they should point to .html files). 

The only way I can figure out how to get the links to render properly is to use RST `:doc:` links, which don’t work correctly on github. 

**My suggested fix:**
Have the README.md file only for github and use a separate index.rst file for the readthedoc index page. The danger is a duplication of information... but I the README could just have a github-clickable-link to each spec for fast navigation and the index file could have the nice paragraphs of text.

This would also fix #234. If we moved the nice spec detail list to the index file, the README file could include build/install/contribution instructions, which would only appear on github.

Thoughts?

gonsie:
This PR move the detailed RFC listing to the index file, which will appear as the top level page for RFC on readthedocs. This is entry page for users.

The README file is removed from sphinx processing. The README is now only used for indexing directly to source files on GitHub and providing information for developers.

This is a major change to the README, so I’d like a few people to review it before merging.

Closes #242 and Closes #234

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@ This is the Flux RFC project.
 We collect specifications for APIs, file formats, wire protocols, and
 processes.
 
+The full RFC specs can be found on [readthedocs](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest).
 Details about each of the active RFC documents can be found [in the index](index.rst)
 
 Table of Contents
--- a/README.md
+++ b/README.md
@@ -6,158 +6,51 @@ This is the Flux RFC project.
 We collect specifications for APIs, file formats, wire protocols, and
 processes.
 
-Active RFC Documents
---------------------
+Details about each of the active RFC documents can be found [in the index](index.rst)
 
-#### [1/C4.1 - Collective Code Construction Contract](spec_1.rst)
+Table of Contents
+-----------------
 
-The Collective Code Construction Contract (C4.1) is an evolution of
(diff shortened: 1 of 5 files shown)


garlick:
It makes sense to me to drop the verbose RFC listing for github rendering.  You had to scroll way down to see it anyway, making it kind of useless.  That was only there because it was the only user facing page we had.  Good change IMHO.

gonsie:
Great idea @SteVwonder. I’ve added that change. I think it’s ready to merge.

PR #248, opened 2020-06-05
rfc20: add optional attributes key under scheduling

dongahn:
Problem: rfc20 is current written:

"The scheduling key allows RFC4-compliant schedulers
to serialize any subset of graph resource data into
its value and later deserialize this value
with no data loss. The scheduling key contains
a dictionary with a single key: graph.
Other keys are reserved for future extensions."

As such, we can't add addition keys under the
scheduling key.

As part of resource restruction work within fluxion,
we need to embed the queue name into RV1 so that
when fluxion qmanager is reloaded, it can
enqueue each currently running jobs into
the correct queue.

W/ this proposal, we can stick the queue name
to `scheduling.attributes.system.queue`.

### DIFF
--- a/spec_20.rst
+++ b/spec_20.rst
@@ -99,13 +99,17 @@ Resource Set Format Definition
 
 The JSON documents that conform to the *R* format SHALL be referred
 to as *R* JSON documents or in short *R* documents.
-An *R* JSON document SHALL consist of a dictionary with three
-keys: ``version``, ``execution`` and ``scheduling``. It SHALL be valid if and only
+An *R* JSON document SHALL consist of a dictionary with four
+keys: ``version``, ``execution``, ``scheduling`` and ``attributes``.
+It SHALL be valid if and only
 if it contains the ``version`` key and either or both the ``execution``
 and ``scheduling`` keys. The value of the ``execution`` key SHALL contain
 sufficient data for the execution system to perform its
 core tasks. The value of ``scheduling`` SHALL contain sufficient data
-for schedulers.
+for schedulers. Finally, the value of ``attributes`` SHALL provide
+optional information including but not being limited
+to data specific to the scheduler used to create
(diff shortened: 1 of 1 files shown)


dongahn:
Made changes per Issue #248

spec_20.rst
```diff
@@ -100,12 +100,16 @@ Resource Set Format Definition
...
-keys: ``version``, ``execution`` and ``scheduling``. It SHALL be valid if and only
+keys: ``version``, ``execution``, ``scheduling`` and ``attributes``.
+It SHALL be valid if and only
 if it contains the ``version`` key and either or both the ``execution``
 and ``scheduling`` keys. The value of the ``execution`` key SHALL contain
 sufficient data for the execution system to perform its
 core tasks. The value of ``scheduling`` SHALL contain sufficient data
-for schedulers.
+for schedulers. Finally, the value of ``attributes`` SHALL provide
+provide opional information including but not being limited
```
grondo, line 110: typo: double "provide" here.

grondo: Also, missing `t` on `opional` (though I bet spellcheck caught that one)

dongahn: Good catch! Thanks.

spec_20.rst
```diff
@@ -100,12 +100,16 @@ Resource Set Format Definition
 The JSON documents that conform to the *R* format SHALL be referred
 to as *R* JSON documents or in short *R* documents.
 An *R* JSON document SHALL consist of a dictionary with three
```
SteVwonder, line 102:
```suggestion
An *R* JSON document SHALL consist of a dictionary with four
```

dongahn: Ah... good catch. Just forced a push.

dongahn:
Thanks @SteVwonder and @grondo!

PR #251, opened 2020-06-24
rfc27: Resource Allocation Protocol Version 1

garlick:
Here's a proposed RFC for a slight refinement of the protocol between the scheduler and job manager, heretofore only described at the top of [modules/job-manager/alloc.c](https://github.com/flux-framework/flux-core/blob/master/src/modules/job-manager/alloc.c), as promised in issue flux-framework/flux-core#2608 and discussed in PR flux-framework/flux-core#2960

I changed the `note` (string) key in the `alloc` response to `annotations` (object) in the ANNOTATE response only, leaving `note` (string) as is for SUCCESS and DENIED responses. Only ANNOTATE would update the job's hash of scheduler annotations.

### DIFF
--- a/spec_21.rst
+++ b/spec_21.rst
@@ -214,7 +214,7 @@ Resources have been allocated by the scheduler.
 The following keys are OPTIONAL in the event context object:
 
 note
-   (string) Scheduler annotation for resource allocation.
+   (string) A scheduler-dependent message that describes the allocation.
 
 Example:
 
--- a/spec_21.rst
+++ b/spec_21.rst
@@ -32,6 +32,8 @@ Related Standards
 
 -  :doc:`22/Idset String Representation <spec_22>`
 
+-  :doc:`27/Flux Resource Allocation Protocol Version 1 <spec_27>`
+
 
 Background
 ----------
@@ -213,14 +215,14 @@ Resources have been allocated by the scheduler.
 
 The following keys are OPTIONAL in the event context object:
 
-note
-   (string) A scheduler-dependent message that describes the allocation.
+annotations
+   (object) A dictionary of scheduler-dependent key-value pairs as described in RFC 27
 
 Example:
 
 .. code:: json
 
-   {"timestamp":1552593348.088391,"name":"alloc","context":{"note":"rank0/core[0-1]"}}
+   {"timestamp":1552593348.088391,"name":"alloc","context":{"annotations":{"sched.resource_summary:"rank0/core[0-1]"}}}
 
 
 Free Event
--- a/spec_21.rst
+++ b/spec_21.rst
@@ -120,6 +120,22 @@ INACTIVE
    Job data in KVS is now read-only (captive state).
 
 
+Virtual States
+~~~~~~~~~~~~~~
+
+In the interest of encouraging consistent language, we define the following
+"virtual states" as shorthand for the union of two or more actual job states:
+
+PENDING
+  The job is in DEPEND or SCHED states.
+
+RUNNING
+  The job is in RUN or CLEANUP states.
+
+ACTIVE
+  The job is in DEPEND, SCHED, RUN, or CLEANUP states.
+
+
 Exceptions
 ~~~~~~~~~~
 
--- a/spec_27.rst
+++ b/spec_27.rst
@@ -1,0 +1,529 @@
+.. github display
+   GitHub is NOT the preferred viewer for this file. Please visit
+   https://flux-framework.rtfd.io/projects/flux-rfc/en/latest/spec_27.html
+
+27/Flux Resource Allocation Protocol Version 1
+==============================================
+
+This specification describes Version 1 of the Flux Resource Allocation
+Protocol implemented by the job manager and a compliant Flux scheduler.
+
+-  Name: github.com/flux-framework/rfc/spec_27.rst
+
+-  Editor: Jim Garlick <garlick@llnl.gov>
+
+-  State: raw
+
+
+Language
+--------
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to
+be interpreted as described in `RFC 2119 <http://tools.ietf.org/html/rfc2119>`__.
+
+
+Related Standards
+-----------------
+
+-  :doc:`14/Canonical Job Specification <spec_14>`
+
+-  :doc:`16/KVS Job Schema <spec_16>`
+
+-  :doc:`20/Resource Set Specification Version 1 <spec_20>`
+
+-  :doc:`21/Job States and Events <spec_21>`
+
+-  :doc:`23/Flux Standard Duration <spec_23>`
+
+
+Background
+----------
+
+The Flux job manager's role is managing the queue of pending job requests
+and transitioning jobs through the job states defined in RFC 21, actively
+initiating actions in each state.  A Flux scheduler's role is passively
+fulfilling resource allocation requests that the job manager makes on
+behalf of jobs.
+
+A scheduler implementation registers the generic service name ``sched``
+and provides several well known service methods.  The job manager requests
+resources from the scheduler with a ``sched.alloc`` request when a job enters
+SCHED state.  It releases resources with a ``sched.free`` request when a job
+enters CLEANUP state.
+
+The simplest imaginable scheduler satisfies ``sched.alloc`` requests in order
+until it is out of resources, then blocks until ``sched.free`` requests
+release enough resources to satisfy the next ``sched.alloc`` request.
+More complex schedulers consider multiple ``sched.alloc`` requests and
+satisfy them out of order to prioritize or balance measures of success
+such as resource utilization or fairness.
+
+Abstract resource allocation requests are expressed as a *jobspec* object
+(RFC 14).  Concrete resources assignments are expressed a an *R* object
+(RFC 20).  These objects are stored in the KVS per the job schema (RFC 16).
+The ``sched.alloc`` request implicitly refers to *jobspec* in the KVS by job ID,
+while the ``sched.alloc`` response and ``sched.free`` request implicitly refer
+to *R* in the KVS by job ID.
+
+This RFC describes the RPC messages outlined above.  It also describes the
+initialization messages used to establish parameters for scheduler operation
+and identify resources that are already allocated at scheduler startup.
+It does not cover the mechanism by which a scheduler discovers the initial
+inventory of resources.
+
+
+Design Criteria
+---------------
+
+- Support multiple scheduler implementations, minimizing repeated code
+  in schedulers.
+
+- Allow the maximum number of outstanding allocation requests sent by
+  the job manager to be controlled by the scheduler.
+
+- Allow the scheduler module to be reloaded with recovery of resource
+  allocations of running jobs.
+
+- Allow the scheduler module to abort (return from its module thread
+  unexpectedly) without impacting running work.
+
+- Send allocation requests to scheduler in priority, submit time order.
+
+- Inform scheduler of job priority and submit time so it can reorder requests
+  internally, combining these factors with others.
+
+- Support job cancellation in SCHED state.
+
+- Support job priority change in SCHED state.
+
+- The resource allocation protocol should not present obstacles to scaling
+  to O(1M) jobs in SCHED state.
+
+- The protocol should not inhibit scaling job throughput to O(100) jobs per
+  second.
+
+- Capture scheduler specific job annotations for display by the job listing
+  tool (e.g. start time estimates).
+
+
+Implementation
+--------------
+
+To escape scalability limitations of the Flux "tag pool", ``sched.alloc`` and
+``sched.free`` RPCs use the job ID to match requests and responses, and set the
+RFC 6 matchtag message field to zero.  It follows that:
+
+- The job ID MUST appear in the ``sched.alloc`` and ``sched.free`` request
+  and response message payloads.
+
+- There SHALL NOT be more than one ``sched.alloc`` or ``sched.free`` request
+  in flight for each job, since otherwise a request could not be uniquely
+  matched to a response using the job ID.
+
+- The errnum field in ``sched.alloc`` and ``sched.free`` response messages
+  MUST be set to zero, even if the response indicates an error.  Otherwise,
+  the message payload could not include the job ID since RFC 6 defines the
+  payload of an error response to be optional error text.
+
+- The job manager SHALL treat a conventional Flux error response to
+  ``sched.alloc`` or ``sched.free`` with a nonzero errnum field as a
+  scheduler fatality, and SHALL not send further requests to the scheduler
+  until it receives a new ``job-manager.sched-ready`` request (see Finalization
+  below).
+
+The other RPCs behave conventionally.
+
+A detailed description of these RPCs follows.
+
+
+Hello
+~~~~~
+
+Before any other RPCs are sent to the job manager, the scheduler SHALL
+send an empty request to ``job-manager.sched-hello`` and wait for a response.
+
+The response payload SHALL consist of a JSON object with one key: ``alloc``,
+whose value is an array of zero or more jobs with allocated resources.
+Each array entry SHALL be a JSON object with the following REQUIRED keys:
+
+id
+  (integer) job ID
+
+priority
+  (integer) job priority in the range of 0 through 31
+
+userid
+  (integer) job owner
+
+t_submit
+  (double) job submit time in seconds since the UNIX epoch
+
+Example:
+
+.. code:: json
+
+   {
+     "alloc": [
+       {
+         "id": 1552593348,
+         "priority": 20,
+         "userid": 5588,
+         "t_submit": 593011542.2359374
+       },
+       {
+         "id": 1552599944,
+         "priority": 20,
+         "userid": 5588,
+         "t_submit": 593069342.2334354
+       }
+     ]
+   }
+
+For each job in the response, the scheduler SHALL mark its assigned resources
+*allocated* internally.  It MAY look up *R* in the KVS by job ID according
+to the job schema (RFC 16).
+
+If an error response is returned to the ``job-manager.sched-hello`` request,
+the scheduler SHALL log the error and exit its module thread.
+
+
+Ready
+~~~~~
+
+Once the scheduler has processed the ``job-manager.sched-hello`` handshake,
+it SHALL notify the job manager that it is ready to accept allocation requests
+by sending a request to ``job-manager.sched-ready``.
+
+The request payload SHALL consist of a JSON object with the following
+REQUIRED key:
+
+mode
+  (string) selected concurrency mode
+
+The mode string SHALL be one of the following:
+
+single
+  The job manager SHALL send a ``sched.alloc`` request only when there are
+  no outstanding ``sched.alloc`` requests.  This mode is only useful for simple
+  schedulers that run jobs strictly in the job manager queue order.
+
+unlimited
+  The job manager SHALL send a ``sched.alloc`` request for all jobs in SCHED
+  state, with no limit on concurrency.
+
+Example:
+
+.. code:: json
+
+   {"mode":"unlimited"}
+
+The response payload SHALL be empty.
+
+After responding to the ``job-manager.sched-ready`` request, the job manager
+MAY immediately begin sending ``sched.alloc`` and ``sched.free`` requests.
+
+If an error response is returned to the ``job-manager.sched-ready`` request,
+the scheduler SHALL log the error and exit its module thread.
+
+
+Alloc
+~~~~~
+
+The job manager SHALL send a ``sched.alloc`` request when a job enters SCHED
+state, and concurrency criteria established by the initialization handshake
+are met.  The request payload consists of a JSON object with the following
+REQUIRED keys:
+
+id
+  (integer) job ID
+
+priority
+  (integer) job priority in the range of 0 through 31
+
+userid
+  (integer) job owner
+
+t_submit
+  (double) job submit time in seconds since the UNIX epoch
+
+Example:
+
+.. code:: json
+
+   {
+     "id": 1552593348,
+     "priority": 20,
+     "userid": 5588,
+     "t_submit": 593011542.2359374
+   }
+
+Upon receipt of the ``alloc`` request, the scheduler SHALL look up *jobspec*
+in the KVS by job ID according to the job schema (RFC 16).
+
+The response payload is a JSON object with the following REQUIRED keys:
+
+id
+  (integer) job ID
+
+type
+  (integer) response type in the range of 0 through 3
+
+There are four response types:
+
+SUCCESS (0)
+  Resources have been allocated
+
+ANNOTATE (1)
+  The scheduler wishes to annotate the job (see below)
+
+DENY (2)
+  The job cannot be scheduled
+
+CANCEL (3)
+  The alloc request was canceled by a ``sched.cancel`` request (see below).
+
+The ``alloc`` request MAY receive multiple responses.
+
+Alloc Success
+^^^^^^^^^^^^^
+
+If resources can be allocated, the scheduler SHALL ensure that *R* has
+been successfully committed to the KVS per the job schema (RFC 16)
+before responding.
+
+In addition to the above REQUIRED keys, the SUCCESS response includes
+the OPTIONAL key:
+
+annotations
+  (object) key value pairs
+
+Example:
+
+.. code:: json
+
+   {
+     "id": 1552593348,
+     "type": 0,
+     "annotations": {
+       "sched.resource_summary":"rank[0-1]/core0"
+     }
+   }
+
+If present, the job manager SHALL update the job's annotation dictionary
+as described in the next section.  The scheduler MAY delete annotations
+such as ``sched.t_estimate`` that are not relevant now that the allocation
+request has been satisfied.
+
+The job manager posts an ``alloc`` event in response to the successful
+allocation of resources.  A snapshot of job's annotation dictionary, after
+the above update, is included in the ``alloc`` event context per RFC 21,
+thus preserving it in job record when the allocation is successful.
+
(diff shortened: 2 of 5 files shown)


spec_27.rst
```diff
@@ -0,0 +1,486 @@
...
+
+While a job is in SCHED state, the scheduler MAY send multiple ANNOTATION
+type responses to the ``sched.alloc`` request to provide information of
+interest to users while the job is waiting for resources.
+
+In addition to the above REQUIRED keys, the ANNOTATE response includes
+the REQUIRED key:
+
+annotations
+  (object) key value pairs that the job manager
```
grondo, line 320: unfinished statement? "key value pairs tha the job manager"

spec_27.rst
```diff
@@ -0,0 +1,486 @@
...
+
+Example:
+
+.. code:: json
+
+   {
+     "id": 1552593348,
+     "type": 1,
+     "annotations": {
+       "sched.t_estimate": 593016000.0,
```
grondo, line 343: This job will start on my 16th birthday! :cake:

spec_27.rst
```diff
@@ -0,0 +1,486 @@
...
+   {
+     "id": 1552593348
+   }
+
+The scheduler SHALL NOT respond directly to the ``sched.cancel`` request.
+Instead, if a ``sched.alloc`` request is pending for the specified job,
+it SHALL respond to the ``sched.alloc`` request with a CANCEL response
+as described above.  If the specified job does not have a pending
+``sched.alloc`` request, the request SHALL be ignored by the scheduler.
+
```
grondo, line 465: This section overlaps with `Alloc Canceled` above, but worded slightly differently, or am I just misreading? (reading via Github diff I may have missed something obvious)

I any event, should the `Alloc Canceled` and `Cancel` sections be combined or at least follow eachother in the RFC since they (seem to be) closely related?

spec_27.rst
```diff
@@ -0,0 +1,486 @@
...
+- The job ID MUST appear in the ``sched.alloc`` and ``sched.free`` request
+  and response message payloads.
+
+- There SHALL NOT be more than one ``sched.alloc`` or ``sched.free`` request
+  in flight for each job, since otherwise a request could not be uniquely
+  matched to a response using the job ID.
+
+- The errnum field in ``sched.alloc`` and ``sched.free`` response messages
+  MUST be set to zero, even if the response indicates an error, since
+  otherwise the message payload could not include the job ID.
```
grondo, line 120: It may not be clear here why a message payload containing an errnum cannot also contain a job ID. Perhaps explicitly stating why this would be so would be helpful. (e.g., an errnum response cannot also contain a JSON payload)

spec_27.rst
```diff
@@ -0,0 +1,486 @@
...
+     "t_submit": 593011542.2359374
+   }
+
+Upon receipt of the ``alloc`` request, the scheduler SHALL look up *jobspec*
+in the KVS by job ID according to the job schema (RFC 16).
+
+The response payload is a JSON object with the following REQUIRED keys:
+
+id
+  (integer job ID
```
chu11, line 257: typo - no end paren on `integer`

spec_27.rst
```diff
@@ -0,0 +1,486 @@
...
+.. code:: json
+
+   {
+     "id": 1552593348,
+     "type": 2,
+     "note": "more nodes requested than configured"
+   }
+
+After the DENIED response, the ``sched.alloc`` request is complete and may be
+retired by the job manager and scheduler.
```
chu11, line 375: in the current `job-manager` i believe "denied" also stores the `note` in the eventlog.  Should that be listed here as well just like the success case?

spec_27.rst
garlick: oops!  Thanks!

spec_27.rst
garlick: So much for the inscrutibility of my contrived examples!  But I guess I nailed that one!

spec_27.rst
garlick: Good idea - I'll reduce the duplication there and move `Cancel` up above the `Free` section.

spec_27.rst
garlick: How's this:

> - The errnum field in ``sched.alloc`` and ``sched.free`` response messages
  MUST be set to zero, even if the response indicates an error.  Otherwise,
  the message payload could not include the job ID since RFC 6 defines the
  payload of an error response to be optional error text.

spec_27.rst
garlick: Yep, thanks!

spec_27.rst
grondo: Perfect :+1:

garlick:
> A high level anal-ness point on some of the alloc response types names. The response types of "success", "annotation", "denied", and "cancel" don't seem right, they seem to be inconsistent in tense.
> 
> I'll propose "success", "annotate", "deny", and "cancel".

OK, sounds good!

garlick:
Just pushed a fixup commit incorporating the suggested changes (thanks!)

garlick:
Oops missed one - one more fixup :-)

garlick:
OK, one more fixup for stuff that I found on a post-lunch read.  Now I need to figure out why travis is failing.

chu11:
LGTM!

As an aside, even if we want to do more than this (e.g. store annotations for recovery), I think this is a good starting point for implementation.  If there are issues, they can be addressed later.

garlick:
Thanks!

Apparently I have spelling errors but travis isn't reporting them out:
```
Spelling checker messages written to /home/travis/build/flux-framework/rfc/_build/output.txt

[91mWarning, treated as error:[39;49;00m
Found 4 misspelled words
```

I have sphinx working at home nicely in a venv, but `make spelling` fails with
```
Warning, treated as error:
/home/garlick/proj/flux-rfc/py3-sphinx/lib/python3.6/site-packages/sphinx/ext/autosummary/templates/autosummary/base.rst:3:Error in "currentmodule" directive:
maximum 1 argument(s) allowed, 3 supplied.
```
Anybody run into this?

garlick:
Fixed the spelling errors by finding them directly with ispell.  So...shall I squash?

spec_27.rst
```diff
@@ -0,0 +1,486 @@
...
+
+The Flux job manager's role is managing the queue of pending job requests
+and transitioning jobs through the job states defined in RFC 21, actively
+initiating actions in each state.  A Flux scheduler's role is passively
+fulfilling resource allocation requests that the job manager makes on
+behalf of jobs.  The job manager requests resources from the scheduler with
+an ``sched.alloc`` request when a job enters SCHED state.  It releases
+resources with a ``sched.free`` request when a job enters CLEANUP state.
+
+The simplest FIFO scheduler answers ``sched.alloc`` requests in order until it
```
dongahn, line 51: Slightly preference to use FCFS instead of FIFO since it was the term we have used in Fluxion. Regardless, it would be good to spell it out, FCFS (First Come, First Served). 

Slightly preference to use "satisfy" as "answer" can be either "yes" or "no."

spec_27.rst
```diff
@@ -0,0 +1,486 @@
...
+is out of resources, then blocks until ``sched.free`` requests release enough
+resources to satisfy the next ``sched.alloc`` request.  More complex schedulers
+accept multiple ``sched.alloc`` requests and answer them out of order to
```
dongahn, line 54: Instead of "accept", perhaps "MAY consider" would be better? 

"answer" -> "satisfy"

FCFS schedulers can accept multiple requests as well, which is in fact what Fluxion scheduler does it currently.

spec_27.rst
```diff
@@ -0,0 +1,486 @@
...
+maximize fairness, utilization, or some other measure.
```
dongahn, line 55:
```suggestion
balance fairness and resource utilization along with some other important measures.
```

spec_27.rst
```diff
@@ -0,0 +1,486 @@
...
+to *R* in the KVS by job ID.
+
+This RFC describes the ``sched.alloc`` and ``sched.free`` RPC messages outlined
+above.  It also describes the ``job-manager.sched-hello`` and
+``job-manager.sched-ready`` initialization messages used to establish
+parameters for scheduler operation and identify resources that are already
+allocated at scheduler startup.  It does not cover the mechanism by which
+a scheduler discovers the initial inventory of resources.
+
+*grow* and *shrink*, where jobs incrementally acquire and release resources
```
dongahn, line 71: Perhaps we don't need this sentence? It should surely come, but in general any future reference tends to make a doc appear incomplete. IMO, this RFC should be as complete as possible within its intended scope: version 1.

spec_27.rst
```diff
@@ -0,0 +1,486 @@
...
+will be covered in a future revision of this protocol.
+
+
+Design Criteria
+---------------
+
+- Support multiple scheduler implementations, minimizing repeated code
+  in schedulers.
+
+- Allow the maximum number of outstanding allocation requests sent by
```
dongahn, line 81: Hmmm. I thought queue-depth isn't supported.

spec_27.rst
```diff
@@ -0,0 +1,486 @@
...
+  the job manager to be controlled by the scheduler.
+
+- Allow the scheduler module to be reloaded on a live system, without impacting
+  running work.
+
+- Allow the scehduler module to abort (return from its module thread
```
dongahn:
```suggestion
Allow the scheduler module to abort (return from its module thread
```

spec_27.rst
```diff
@@ -0,0 +1,486 @@
...
+Design Criteria
+---------------
+
+- Support multiple scheduler implementations, minimizing repeated code
+  in schedulers.
+
+- Allow the maximum number of outstanding allocation requests sent by
+  the job manager to be controlled by the scheduler.
+
+- Allow the scheduler module to be reloaded on a live system, without impacting
```
dongahn, line 84:
```suggestion
- Allow the scheduler module to be reloaded with its running job and resource state fully recovered on a live system, avoiding any impact on
```

spec_27.rst
```diff
@@ -0,0 +1,486 @@
...
+- Send allocation requests to scheduler in priority, submit time order.
+
+- Inform scheduler of job priority and submit time so it can reorder requests
+  internally, combining these factors with others.
+
+- Support job cancellation in SCHED state.
+
+- Support job priority change in SCHED state.
+
+- Scale to O(1M) jobs in SCHED state.
```
dongahn, line 99: Can we be a bit more specific about what this means? Memory overheads? Making the performance logarithmic?

garlick:
(went ahead and squashed the fixups)

spec_27.rst
```diff
@@ -0,0 +1,507 @@
...
+
+The response payload SHALL be empty.
+
+After responding to the ``job-manager.sched-ready`` request, the job manager
+MAY immediately begin sending ``sched.alloc`` and ``sched.free`` requests.
+
+If an error response is returned to the ``job-manager.sched-ready`` request,
+the scheduler SHALL log the error and exit its module thread.
+
+TODO: add a mode that allows just enough ``sched.alloc`` requests to be
```
dongahn, line 223: Do we need these TODOs as part of this RFC? Perhaps more modes can be added to a future protocol revision?

spec_27.rst
```diff
@@ -0,0 +1,507 @@
...
+^^^^^^^^^^^^^
+
+If resources can be allocated, the scheduler SHALL ensure that *R* has
+been successfully committed to the KVS per the job schema (RFC 16)
+before responding.
+
+In addition to the above REQUIRED keys, the SUCCESS response includes
+the OPTIONAL key:
+
+note
```
dongahn, line 299: Does it mean that if you want to annotate a job with key value pairs, you are required to send an additional response? I don't know how one additional required response affect the other design goal of O(100) jobs per second and O(1M) for high throughput workload...

spec_27.rst
```diff
@@ -0,0 +1,507 @@
...
+
+The job manager SHALL store annotations while the job is in SCHED state,
+and make them available for job listings (mechanism out of scope for this RFC).
+
+The key MAY be one of the following:
+
+sched.t_estimate
+  (double) estimated absolute start time in seconds since UNIX epoch
+
+sched.reason
```
chu11, line 341: Just occurred to me. Instead of `sched.reason`, would be clearer with `annotations.reason`? (adjusting elsewhere of course)

spec_27.rst
dongahn: Should their be any restrictions on the value types?

spec_27.rst
```diff
@@ -0,0 +1,507 @@
...
+If present, the note SHALL be added to the ``exception`` event context
+generated by the job manager when processing the allocation failure.
+
+After the DENY response, the ``sched.alloc`` request is complete and may be
+retired by the job manager and scheduler.
+
+Alloc Cancel
+^^^^^^^^^^^^
+
+When the scheduler receives a ``sched.cancel`` request for a job (see below),
```
dongahn, line 394: I don't know if this can be a problem in real life. But I think there can be a race condition where the scheduler's success response for `sched.alloc` is in flight when the `cancel` request below arrives. In this case, should the scheduler just ignores the `cancel` request below?  

If such a race can indeed occur, we should perhaps document a proper semantics. 

BTW, does `schedutil` wrap `cancel` request below?

I don't recall if I added any support for this.

spec_21.rst
```diff
@@ -120,6 +120,21 @@ INACTIVE
    Job data in KVS is now read-only (captive state).
 
 
+Virtual States
+~~~~~~~~~~~~~~
+
+For convenience, the following virtual states are defined:
```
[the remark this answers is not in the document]
garlick: How about something like this?
> In the interest of encouraging consistent language, we define the following "virtual states" as shorthand for the union of two or more actual job states

spec_27.rst
```diff
@@ -0,0 +1,486 @@
...
+
+Background
+----------
+
+The Flux job manager's role is managing the queue of pending job requests
+and transitioning jobs through the job states defined in RFC 21, actively
+initiating actions in each state.  A Flux scheduler's role is passively
+fulfilling resource allocation requests that the job manager makes on
+behalf of jobs.  The job manager requests resources from the scheduler with
+an ``sched.alloc`` request when a job enters SCHED state.  It releases
```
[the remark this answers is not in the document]
garlick: Oh good point.  I'll add something about the scheduler needing to regsister the service named "sched".

spec_27.rst
garlick: I'm eliminating both FIFO and FCFS in favor of "simplest imaginable scheduler" to avoid any implications about specific implementations.

spec_27.rst
garlick: How about
> ...to prioritize or balance measures of success, such as resource utilization or fairness

spec_27.rst
garlick: Agreed.  I put it in there anticipating "what about grow or shrink" comments, but I actually agree with you about RFC's standing alone.

spec_27.rst
garlick: You can have any queue depth you want as long as it's one or infinity.

Seriously: anticipating fixing that.

spec_27.rst
garlick: Already fixed (I think you're reviewing a slightly outdated version)

spec_27.rst
garlick: How about 
> Allow the scheduler module to be reloaded with recovery of resource allocations of running jobs

spec_27.rst
garlick: To put a finer point on it, how about
> The protocol should not present any obstacles to scaling to O(1M) jobs

And the specific limitation in the protocol (the matchtag pool) that needed to be overcome is described later.

spec_27.rst
garlick: Deleted.

spec_27.rst
garlick: > Does it mean that if you want to annotate a job with key value pairs, you are required to send an additional response?

Yes, but if the annotations are for letting the user know why the job is still pending, then it seems like it's not useful to annotate in the allocation SUCCESS response which causes the job to leave PENDING.

spec_27.rst
garlick: The idea there was to create a namespace, leaving the door open for other forms of annotation.  For example we had talked about allowing jobs to annotate themselves, thinking of letting an application add application specific info `flux jobs` output.  However, that's for the future - just leaving the door open here.

spec_27.rst
garlick: It should be fine to ignore cancel if a terminating response has already been sent to  the alloc request.  Later on I stated
>  If the specified job does not have a pending ``sched.alloc`` request, the request SHALL be ignored by the scheduler

> does schedutil wrap cancel request below?

Yes, see `schedutil_alloc-respond_cancel()`

garlick:
Just pushed a couple of fixups I think addressing the comments so far - thanks!

garlick:
I'll go ahead and squash if there are no more comments for today

spec_27.rst
dongahn: Sounds good to me.

spec_27.rst
dongahn: > If the specified job does not have a pending sched.alloc request, the request SHALL be ignored by the scheduler

Great. I didn't catch that. Thanks.

spec_27.rst
dongahn: Hmmm...

I think it would be wise to use the annotation to pass scheduler-specific information beyond "why the job is still pending". For example, which queue the job has been used for scheduling -- given that Fluxion has multi-queue support. And in the future there would be more scheduler specific data that the scheduler wants to pass to the job listing tool via this mechanism. 

For this type of information, I would much prefer sending that in the SUCCESS response in one message. For other types (like why the job is pending), separate annotation messages make sense. 

Can we perhaps expand "note" so that it is key value pairs but has a human readable string that can go into the event log? (e.g., one mandatory key is "note"?)

spec_27.rst
garlick: Good question.  I purposefully didn't add any restrictions, thinking that our listing tool is flexible enough to handle any JSON type in a custom format, however we might want to exclude objects and arrays.  Thoughts from people more familiar with `flux jobs` internals?

spec_27.rst
garlick: Thank you @dongahn, that is is critical feedback!

It does seem like that sort of information should be persistent (part of job record, preserved across instance restart), while the start time estimate _updates_ are practically speaking, not scalable to preserve.

What about this:
1) let the ANNOTATE response continue to return a dict that is stored in memory but does NOT generate an eventlog entry, since doing so might blow up the storage required by each job
2) change the SUCCESS response to return a dict that also merged in memory with the ones from ANNOTATE
3) include a snapshot of all of the in-memory annotations in the context of the `alloc` eventlog entry

This would mean the annotations are not preserved if the alloc request does not end in a SUCCESS response.

(I think this is a variant @chu11 and I discussed but without a use case I wasn't sure we wanted to go there yet)

I think we could just do away with `note` in the SUCCESS response in this case.  Perhaps we could define a more descriptive annotation key `sched.resource-summary` to replace it.

chu11: My only comment is, if we're going to remove the human readable `note` from SUCCESS and ANNOTATE, should we just go ahead and do it for DENY as well?

chu11: > include a snapshot of all of the in-memory annotations in the context of the alloc eventlog entry

Just throwing out an idea, since saving a json-dict inside of a json-dict in the `alloc` eventlog entry can be ugly.  Perhaps only specific annotations are saved to the `alloc eventlog`.  Obviously saving `sched.t_estimate` is sort of useless.  We could say only `reason` and `summary ` get saved?

spec_27.rst
chu11: in the bowels of the python bindings, ultimately there's a `json.loads()` call, so it should be able to handle any JSON type.  
So I think we can leave this typeless.  If `json.loads()` can't handle it, I suspect we've got bigger problems.

spec_27.rst
garlick: > If we're going to remove the human readable note from SUCCESS and ANNOTATE, should we just go ahead and do it for DENY as well?

When we receive a DENY response, we generate a fatal exception for the job.  RFC 21 defines all exception events generally to have an (optional) `note` in the context dict.  I don't think we want to change the general exception event structure.  Therefore, I would argue that DENY needs a `note` to transfer directly to the exception.  It would not go in the annotations.

Changing the `alloc` event in RFC 21 seems OK for this PR though as that has a narrower scope.  As far as how to map the annotation dict into the event context, why is a dict within a dict a problem?  I think the main use case is the eventlog replay in the job manager at instance startup would need to populate `job->annotations`.

I'm a little worried about defining which keys are volatile in the RFC.  Do we need to make values special in that way?  I would rather just say that all annotations get saved if the allocation is successful, and all are tossed if not.  I suppose we could let the scheduler set keys to null to remove them from the dict?  Would that solve the problem you are thinking of?  (E.g. scheduler removes keys in the SUCCESS response that were only useful while the job was pending, to avoid saving them in the eventlog)

chu11: > why is a dict within a dict a problem?

Its no issue.  You had mentioned it in  https://github.com/flux-framework/flux-core/pull/2960

> Do we need to make values special in that way?

I was mostly worried about excessive annotations being put into the eventlog and eating up space.

>  I suppose we could let the scheduler set keys to null to remove them from the dict?  Would that solve the problem you are thinking of? 

Allowing the clearing of some keys seems like a good idea.  Within my testing, it was annoying to see some keys show up after a job was allocated.

garlick: Sounds like we are on the same page then.  I'll add a fixup commit that proposes this change.

Regarding dict within a dict, I think my problem in the PR was that the inner dict was encoded as a string, _then_ stored as a string value in the outer dict.

garlick:
Pushed a fixup for the proposed annotation changes (oops, introduced a spelling error - fixed it and forced a push).

Also added a commit to RFC 21 to redefine the alloc event context as discussed.

garlick:
@chu11 and @dongahn, if you have a moment to review the changes since last review I'd appreciate it.  Are we getting close?

chu11:
it LGTM

SteVwonder:
I noticed that the readthedocs builder didn't kick in and provide a test build/view of this PR.  I resync'd the webhook, since it looks like it got dropped on the github side.  @garlick, when you do the rebase to squash the fixup commits, I'd be interested to see if the readthedocs check works.

spec_27.rst
```diff
@@ -0,0 +1,524 @@
+.. github display
+   GitHub is NOT the preferred viewer for this file. Please visit
+   https://flux-framework.rtfd.io/projects/flux-rfc/en/latest/spec_21.html
```
SteVwonder, line 3:
```suggestion
   https://flux-framework.rtfd.io/projects/flux-rfc/en/latest/spec_27.html
```

spec_27.rst
dongahn: > change the SUCCESS response to return a dict that also merged in memory with the ones from ANNOTATE

This will work for me well. For high through put case with simple queuing policies, this will generally result in one SUCCESS response per `alloc` request so no performance impact and yet still allow the scheduler to pass all key information to be used in job listing.

spec_27.rst
garlick: Nice catch!

garlick:
Thanks I went ahead and squashed and rebased.  And one more time to rebase on current master.

@dongahn, if you have time please skim the "Alloc Success" and "Alloc Annotate" sections.

garlick:
That fixed the rendering @stevwonder - thanks!

https://flux-framework--251.org.readthedocs.build/projects/flux-rfc/en/251/spec_27.html

dongahn:
> @dongahn, if you have time please skim the "Alloc Success" and "Alloc Annotate" sections.

They read very well. Thanks!

spec_27.rst
```diff
@@ -0,0 +1,524 @@
...
+
+The key MAY be one of the following:
+
+sched.t_estimate
+  (double) estimated absolute start time in seconds since UNIX epoch
+
+sched.reason_pending
+  (string) human readable reason job is pending
+
+sched.resource_summary
```
chu11, line 356: everything LGTM, but wanted to bring up one question / possible requirement since it occurred to me while working on https://github.com/flux-framework/flux-core/pull/2960

Would we want some of these special fields to be automatically cleared after a success or denied response?

Some like `t_estimate` or `reason_pending` have no purpose for being written out after a `success` response.

garlick: My thinking was that special casing those keys might be strange if the scheduler decides to introduce other keys that have the exact same semantics (only useful during PENDING).  Let's let the scheduler clean up after itself.

chu11: Sounds good.  Another aside as I've been converting code.  Do we want to rename `resource_summary` to `R`?  As that's what we call it in the KVS..

grondo: `resource_summary` is not the same as `R` that might be confusing...

garlick: R is a different beast - this is just a human readable summary so I would vote no on that.

Just pushed a new draft, dare I say approaching final?  I did add "The scheduler MAY delete annotations such as sched.t_estimate that are not relevant now that the allocation request has been satisfied." in the _Alloc Success_ section after your comment above.

Plus a few minor edits.  I'm comfortable with this draft if someone wants to approve and set MWP.

p.s. the readthedocs preview in the travis result is _so very nice_!

spec_21.rst
```diff
@@ -32,6 +32,8 @@ Related Standards
 
 -  :doc:`22/Idset String Representation <spec_22>`
 
+-  :doc:`27/Flux Resource Allocation Protocol Cersion 1 <spec_27>`
```
garlick, line 35: DOH typo!

spec_27.rst
chu11: Yeah, I realized it just a second ago, R goes into the KVS.  `resource_summary` is different.

spec_21.rst
garlick: force pushed a fix.  Sheesh.  Last one I promise!

garlick:
I'll go ahead set MWP.

garlick:
Thanks everybody!

PR #253, opened 2020-06-27
rfc28: add new RFC for resource acquisition

garlick:
I thought it would make sense to complement RFC 27 with an RFC for the other major scheduler interface:  `resource.acquire`.  This will need an update once we settle on the "resource object" format (*R*).

### DIFF
--- a/spec_28.rst
+++ b/spec_28.rst
@@ -191,7 +191,7 @@ If subtracted resources are already allocated to a job, the scheduler should
 raise a fatal exception on the job.  The scheduler SHALL then be prepared
 to process ``sched.free`` request(s) involving these resources.
 
-Once the subtracted resources are free, the scheduler SHALL send a
+As the subtracted resources become free, the scheduler SHALL send a
 ``resource.release`` request to the flux-core resource module (see below).
 
 The scheduler SHOULD re-evaluate the satisfiability of all jobs in its
@@ -265,8 +265,9 @@ conform to this protocol.
 Release Request
 ^^^^^^^^^^^^^^^
 
-The scheduler SHALL send a ``resource.release`` request to the resource
-module to acknowledge *shrink* above, once the subtracted resources are free.
+The scheduler SHALL send one or more ``resource.release`` requests to the
+resource module to acknowledge *shrink* above, as the subtracted resources
+become free.
 
 The ``resource.release`` request is a JSON object with the following
 REQUIRED keys:
@@ -286,13 +287,11 @@ Example:
       "targets": "1-3,5"
    }
 
-There SHALL be exactly one ``resource.release`` request for a given *shrink*.
-
 The *group* SHALL match the group used in the ``resource.acquire`` request.
-The *targets* SHALL match the *shrink* targets exactly.
+The *targets* SHALL match, or be a proper subset of, the *shrink* targets.
 
-The scheduler SHALL NOT send a ``resource.release`` request that was not
-solicited by *shrink*.
+The scheduler SHALL NOT send a ``resource.release`` request for execution
+targets that were not previously included in *shrink*.
 
 The release response SHALL have an empty payload.
 
--- a/spec_28.rst
+++ b/spec_28.rst
@@ -1,0 +1,319 @@
+.. github display
+   GitHub is NOT the preferred viewer for this file. Please visit
+   https://flux-framework.rtfd.io/projects/flux-rfc/en/latest/spec_28.html
+
+28/Flux Resource Acquisition Protocol Version 1
+===============================================
+
+This specification describes the Flux service that schedulers use to
+acquire exclusive access to resources and monitor their ongoing
+availability.
+
+-  Name: github.com/flux-framework/rfc/spec_28.rst
+
+-  Editor: Jim Garlick <garlick@llnl.gov>
+
+-  State: raw
+
+
+Language
+--------
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to
+be interpreted as described in `RFC 2119 <http://tools.ietf.org/html/rfc2119>`__.
+
+
+Related Standards
+-----------------
+
+-  :doc:`20/Resource Set Specification Version 1 <spec_20>`
+
+-  :doc:`22/Idset String Representation <spec_22>`
+
+-  :doc:`27/Flux Resource Allocation Protocol Version 1 <spec_27>`
+
+
+Background
+----------
+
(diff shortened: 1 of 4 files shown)


dongahn:
@garlick: I have a few tasks that I need to finish up soon. But so that you now, I do plan to review this. You can assign me as one of your reviewers if you want.

garlick:
Any comments would be welcome, thanks.  I'm not in any rush to merge this - we may want to leave it open until we are are further along with Rv2.

spec_28.rst
```diff
@@ -0,0 +1,204 @@
...
+The resource acquisition protocol allows the scheduler to obtain the initial
+set of resources and monitor ongoing availability, without dealing directly
+with the details, which are managed by the *resource* module.
+
+Version 1 of this protocol maps chunks of resources to integer *execution
+targets*, and reports availability at the target level.  All resources are
+mapped to targets, and all the resources associated with a given target are
+either up or down as an atomic unit.
+
+Note: at this time, execution targets map directly to Flux broker ranks,
```
dongahn, line 53: Can we say: version 1 maps does this mapping at the Flux broker rank level instead of using a loose term: "at this time?" 

That can make this statement (and thus this v1 RFC) a bit tighter. 

If and when we change the exec target levels, we will probably update the version to something higher anyway?

spec_28.rst
```diff
@@ -0,0 +1,204 @@
...
+however this fact should not be relied upon in scheduler design.
+
+When a scheduler initializes, it obtains exclusive access to a set of
+resources with the streaming ``resource.acquire`` RPC.  The initial response
+includes a *resources* object and an *up* idset (RFC 22) defining which of the
```
dongahn, line 58: So the plan is to keep the current protocol as is? If it will not be too time consuming, I think we should consider introducing a few enhancements to V1 protocol to make the scheduler architecture written on top of V1 a bit cleaner.

1. In addition to `resources`, we can introduce `grow` key or similar so that the scheduler doesn't have to parse out `resources` to get the execution target set.

2. "unexclude" can also map to the same "grow" key, and this would be the special case of elasticity that V1 can support and allow schedulers to add some elastic features. 

3. "unexclude" currently doesn't generate a response and this probably needs to be addressed before V1 is cast in stone? Or document this deficiency? 

4. Currently an "exclude" event shows up as "down" to the scheduler, and this breaks the satisfiability semantics of schedulers. This would be better mapped to the "shrink" key or similar, and so that satisfiability is calculated based on this resulting resources. "down" resources shouldn't affect  satifiability.

garlick: Yes I think we can make a small change to the protocol to address this.  Sorry I was moving fast when resource was created and had thought the exclusion set change was not likely to be a high priority TOSS production issue so had pushed it off, but given that it's so easy to fix, that was short sighted.

The first response currently includes "resource" and "up" keys.  I would propose that we add an "include" key that includes the idset for execution targets that should be considered for satisfiability.

Then for subsequent response we would also send, as needed, "include" and "exclude" keys to allow this list to be changed.

(I was wondering if we should call those grow and shrink but I think maybe grow might eventually need to set another resource object and so we might want to reserve those as more advanced operations to be added later?)

grondo: Since we're changing the protocol, could `resource` be renamed `grow` (or something more appropriate), to reduce the specialness of the first response, since technically the first response is just a "grow from nothing"?

Sorry if that is a dumb idea.

garlick: I think the first response _Is_ special since you really cannot tell the job manager "ready" until you have some resources, otherwise all jobs would be rejected as unsatisfiable.

That said, I'm fine with renaming `resource` to `grow` if that makes more sense to people.  What say you @donghan?  I see the use of "grow" in your PR, does this change fit?

grondo: Sorry, I didn't phrase my question correctly. Of course the first response is special, but why does the internal protocol/keys that tell the scheduler to add some resources have to be special?

Actually, on my ride I was wondering if we could have `grow`/`shrink` or `add`/`remove` keys under a `resource` key for more organization (ie. no ad hoc keys). The first response, plus any future response that wants to "add" resources or "grow" would set `resource.add`. In the future a `resource.subtract` or `remove` could be used to "shrink".

> since you really cannot tell the job manager "ready" until you have some resources, otherwise all jobs would be rejected as unsatisfiable.

It also occurred to me that there will be a similar problem with "grow". If you are in the middle of a grow operation and some jobs are submitted that can take advantage of the new resources, they will rejected as unsatisfiable until the scheduler has received the resource response with the new resources.

garlick: That sounds reasonable.

On the last point, it seems like a "growable" instance will want to disable or relax the satisfiability check since it's no longer impossible jobs that exceed available resources to run.  Maybe if there are boundaries on its growth it could check those?

dongahn: Yes, grow/shrint and up/down are exactly how fluxion is mapping resource.acquire keys. So this should work very well currently as well as our future elastic scheduling support.

dongahn: https://github.com/flux-framework/flux-sched/blob/5f58fc00168583820e3c9ae210521ea81abbdc39/resource/modules/resource_match.cpp#L702

dongahn: > On the last point, it seems like a "growable" instance will want to disable or relax the satisfiability check since it's no longer impossible jobs that exceed available resources to run. Maybe if there are boundaries on its growth it could check those?

Yes something like this. But I think even without this, the code should support the workloads pretty well for awhile.

grondo: > On the last point, it seems like a "growable" instance will want to disable or relax the satisfiability check since it's no longer impossible jobs that exceed available resources to run. Maybe if there are boundaries on its growth it could check those?

It would be nice if *any* flux instance could turn into a growable instance. In my mind this would indicate a clean design, vs a set of conditional logic for growable vs non-growable instance.

It seems like there might be a use for a workload requesting an instance to grow, then submitting more jobs to be scheduled which would take advantage of the increase in resources. In that case, the initial `resource.acquire` response is just a little less special (is the only point I was trying to make initially).

Is this an impossible design?

dongahn:
@garlick: was there any change to this that I need to review?

dongahn:
Ah... probably, e13b40b and 668c205

dongahn (changes requested):
@garlick: This is a great improvement and step into the right direction! 

I made some mostly minor points.

But having a guaranteed way to receive the `.scheduling` key needs to be described. (see the inline comment above.).

@milroy and @SteVwonder : after you are done with IPDPS submission, it would be great for your to review this. I can see various ways to extend this to support general elasticity and multi-tired storage scheduling.

garlick:
IMHO we should just document the current protocol in RFC 28 as version 1 and revisit this discussion  for  version 2.  Closing this PR for now.

PR #255, opened 2020-07-08
rfc19: propose a standard (alternate) encoding for FLUIDs

grondo:
As discussed in flux-framework/flux-core#3038, this PR proposes a standard alternate encoding for FLUIDs, most useful in decreasing the space required for JOBID output in tools such as `flux jobs`. I figured we'd get the idea down in the RFC so we could refer to it during development (assuming there is concensus).

### DIFF
--- a/spec_19.rst
+++ b/spec_19.rst
@@ -90,12 +90,84 @@ equal to zero, so no collision detection is required.
 Representation
 ~~~~~~~~~~~~~~
 
-A FLUID is a 64-bit integer, e.g. ``4208082371413712``.
+A FLUID is a 64-bit integer, e.g. ``6731191091817518``.
 
-Encoding as a 16-character hexadecimal string is a fairly compact string
-representation, e.g. the above example encodes to ``"749de79000800"``.
+Representations other than decimal MAY be used where appropriate,
+for instance for compactness or ease of transcription over the phone.
 
-Schemes like `mnemonicode <https://github.com/singpolyma/mnemonicode>`__ can
-convert integers to strings of pronounceable words and back again, e.g. the
-above example encodes to ``"memo-felix-virus—​warning-france-academy"``.
-A larger dictionary decreases the number of words required.
+The following sections describe the set of supported alternate
+representations for FLUIDs.
+
+
+FLUID base58 (F58) Encoding
++++++++++++++++++++++++++++
+
+In order to create a compact, human readable representation
+of a FLUID, the main alternate encoding of a FLUID SHALL be `Base58
+<https://en.bitcoinwiki.org/wiki/Base58>`__, using the alphabet
+
+ ``123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz``
+
+To disambiguate this Base58 representation from decimal or other
+representations, FLUIDs with this encoding SHALL be prefixed with the
+Unicode character U+0192 ``ƒ``, and the final result SHALL be termed
+"FLUID base58 Encoding" or F58 for short. For example, the FLUID
+``6731191091817518`` from above SHALL be represented in F58 as
+``ƒuZZybuNNy``. As a fallback mechanism, a FLUID prefixed with an
+ASCII lowercase f ``f`` will also be decoded as F58.
+
+Examples: ``ƒZemgA8Bzf``, ``ƒ278oEf7zGf``
+
+FLUID Hexadecimal (hex) Encoding
+++++++++++++++++++++++++++++++++
+
+A hexadecimal encoding SHALL represent a FLUID in base16, including
+a ``0x`` prefix to unambiguously differentiate the representation from
+other FLUID standard encodings.
+
+Examples: ``0x17e9fb8df16c2e``, ``0xedaf97d000000``
+
+
+FLUID Dotted-Hexadecimal (dothex) Encoding
+++++++++++++++++++++++++++++++++++++++++++
+
+In order to support indexing of FLUIDs in a hierarchical KVS namespace,
+a dotted-hexadecimal encoding SHALL represent a FLUID in base16,
+with each 4 hexadecimal digits separated by dots (``.``).
+
+Examples: ``0017.e9fb.8df1.6c2e``, ``000e.daf9.7d00.0000``
+
+
+FLUID Mnemonic (words) Encoding
++++++++++++++++++++++++++++++++
+
+In order to ease transferring of FLUIDs via human interaction, a
+mnemonic representation of FLUIDS SHALL be supported by a conformant
+implementation.
+
+The `mnemonicode <https://github.com/singpolyma/mnemonicode>`__
+implementation converts integers to strings of pronounceable words and
+back again. This encoding MAY be used when a FLUID must be conveyed
+by speaking, e.g. over the phone.
+
+Examples: ``reform-remote-galileo--heart-package-academy``, ``random-idea-yoyo--sugar-printer-academy``
+
+
+Decoding Alternate FLUID Representations
+++++++++++++++++++++++++++++++++++++++++
+
+The standard FLUID representations described in this RFC are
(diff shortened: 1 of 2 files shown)


grondo:
Pushed a commit to update the example FLUID (to one with more human interest.)

Also thinking that FSE is too close to FSD, and maybe a better name for this "standard" FLUID encoding is simply "f58".

garlick:
Fuzzbunny - Love it.!

grondo:
Yeah, who is against fuzzy bunnies? I dare you to speak up!

spec_19.rst
```diff
@@ -90,12 +90,34 @@ equal to zero, so no collision detection is required.
...
+<https://en.wikipedia.org/wiki/Base58>`__, using the alphabet
+
+ ``123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz``
+
+To disambiguate this Base58 representation from decimal or other
+representations, FLUIDs with this encoding SHALL be prefixed with the
+Unicode character U+0192 ``ƒ``, and the final result SHALL be termed
+"FLUID Standard Encoding" or FSE. For example, the FLUID
+``6731191091817518`` from above SHALL be represented in FSE as
+``ƒuZZybuNNy``.
```
SteVwonder, line 123: tl;dr: uneducated unicode FUD, feel free to ignore

I like the look of the unicode ƒ, my main concern is that users might confuse it with the base58 f.

If a user is copying the ID by sight, are they going to be able to easily type the unicode f?  On a mac it is easy once you know the key-binding: option+f.  Not sure what it is elsewhere.

I'm also not sure what the general terminal unicode support story is. My  experience is that every once and a while I end up on a system where my unicode prompt doesn't display right.  Usually due to fonts that don't support unicode or a bad set of LC_* variables.

grondo: > I like the look of the unicode ƒ, my main concern is that users might confuse it with the base58 f.

> If a user is copying the ID by sight, are they going to be able to easily type the unicode f? On a mac it is easy once you know the key-binding: option+f. Not sure what it is elsewhere.

Both of these are reasonable objections. But how often will users by typing FLUIDs by hand? I don't think I've done that once with the current FLUIDs, they are too long. I would hope users would copy and paste in most cases.

To cover this case, though, could we have a heuristic that if a JOBID isn't an integer and starts 
with an `f` it is decoded as a base58 FLUID?

>  where my unicode prompt doesn't display right. 

Worst case you end up with JOBIDs that are prefixed by a little box. I tried various fonts with xterm on my system and I could not get it to fail to display though, so I'm not sure what will actually happen.

I can sympathize with the worry, though, especially since the chances of unicode problems at LC is probably higher than the outside world at large. Maybe we opt for `f` of `f.` as a prefix instead?

grondo: The unicode `ƒ`, proves we've reached a higher level of conciousness, but I don't mind `f.` either:

```
$ flux jobs -a -c 10
JOBID        USER     NAME       STATUS NTASKS NNODES  RUNTIME RANKS
f.X6GDYF9    grondo   hostname       CD      1      1   0.071s 0
f.J13fL15    grondo   hostname       CD      1      1   0.080s 0
f.3DAUuz3    grondo   hostname       CD      1      1   0.010s 0
f.3DAUuz2    grondo   hostname       CD      1      1   0.010s 3
```

SteVwonder: > The unicode ƒ proves we've reached a higher level of conciousness,

Yeah, I definitely don't want to block achieving enlightenment. 😆 

> Maybe we opt for f of f. as a prefix instead?

Maybe we move ahead with ƒ and then if it's an issue we circle back to optionally allow`f.` as a prefix?

garlick: I like F58 as the name for this encoding.

Maybe retyping could come up when someone opens a ticket about a job that's running on an air gapped system?

However if a job id begins with `f` instead of ƒ we do at least know it's not an integer, so if we say that F58 SHALL have a single character prefix of ƒ or `f`, are we good?

garlick: IOW never try to interpret a jobid as F58 if the prefix is missing, so the fact that `f` is in the alphabet does not create an ambiguity.

grondo: That is what I was thinking, just pushed updated language to allow either `f` or `ƒ` as prefix, and rename encoding to F58.

grondo:
Looks like the [Wikipedia reference](https://en.wikipedia.org/wiki/Base58) was literally [just deleted](https://en.wikipedia.org/wiki/Wikipedia:Articles_for_deletion/Base58). I have never seen that before (is _un_ serendipitous a word?)

Maybe drop the reference since the concept of baseN should be clear? There's also a [rfc draft](https://tools.ietf.org/id/draft-msporny-base58-01.html), but it is expired.

grondo:
Pushed a fixup commit to replace the Wikipedia reference with a reference to the Bitcoin wiki description of Base58. Honestly we could also probably drop the reference since we're just representing 64-bit integers in base58, not describing a generic binary encoding.

grondo:
Unfortunately, once I had an implementation *mostly* complete, I did run into a problem with Unicode support as @SteVwonder feared. If the current locale is set to a non-UTF-8 locale, and attempt to print a jobid encoded in `F58` fails with:
```
flux-jobs: ERROR: 'ascii' codec can't encode character '\u0192' in position 2: ordinal not in range(128)
```

`sharness` explicitly set the locale to `C`, so this problem hits the sharness tests.

It seems like this is a [well](https://click.palletsprojects.com/en/5.x/python3/#python-3-surrogate-handling) [known](https://vstinner.github.io/posix-locale.html) [issue](https://fedoraproject.org/wiki/Changes/python3_c.utf-8_locale) such that [PEP 538](https://www.python.org/dev/peps/pep-0538/) (coercion to UTF-8 locale) was proposed, but it seems like that PEP only made it into 3.7.

Not sure if we can do a similar thing and coerce locales to a valid UTF-8. This kludge does seem to work:
```
    # Re-open stdout with utf8 encoding
    sys.stdout = open(sys.stdout.fileno(), mode='w', encoding='utf8')
```
 
Only needed in `flux-*.py` commands that output JOBIDs encoded as F58 strings.

Also seems that `PYTHONIOENCODING=utf-8:surrogateescape` works.

I guess if these  minor changes work around the issue in sharness, we shouldn't have troubles in various locales that might be set on production systems, but also maybe it would be safer to use ascii `f` as the F85 default prefix.

(Sorry for doubting you earlier @SteVwonder, this unicode thing in Python is indeed a mess!)

garlick:
I'm fine with the coersion as it seems like that's where python is heading.  The unicode output looks cool and we do have a fallback plan if problems arise.

grondo:
In the implementation proposed in flux-framework/flux-core#3045, a new `fluid_parse()` function was introduced that allows decoding a FLUID/JOBID string from any supported encoding using the following rules:

 - if the string contains a `.`, decode as `dothex`
 - if the string contains a `-`, decode as `words` (mnemonic)
 - if the string starts with `ƒ` or `f`, decode as F58
 - if the string starts with `0x` decode as hexidecimal
 - O/w, decode as decimal

E.g.:
```console
$ flux job id ƒLUX  
65512
$ flux job id 0000.0000.0000.ffe8
65512
$ flux job id modular-archive-academy--academy-academy-academy
65512
```

Should these rules be made explicit in this RFC? Or is that overkill? Just throwing it out there.

garlick:
> Should these rules be made explicit in this RFC? Or is that overkill? Just throwing it out there.

Yeah, that IMHO would be useful to document.

grondo:
I've pushed an update here that reworks the "Representation" section to include descriptions of all supported FLUID encodings, as well as a set of "rules" for decoding FLUID strings in supported representations.

If this looks ok (might need some wordsmithing), then I can squash all the commits together.

garlick:
Nice!  That works for me.  Please squash and set MWP when you're ready.

grondo:
Cast in stone!

PR #259, opened 2020-07-22
rfc27: Update per discussion and add new user namespace

chu11:
Per some discussion in https://github.com/flux-framework/flux-core/issues/2608, instead of annotations being in a single flat namespace, have `sched` fields within a dictionary within the annotations.

Given choices in https://github.com/flux-framework/flux-core/pull/3065, add a `user` namespace/dictionary.

Not 100% sure if I used the right language (should I call it a `dictionary`?  an `object`? a `namespace`?)

### DIFF
--- a/spec_27.rst
+++ b/spec_27.rst
@@ -363,6 +363,9 @@ sched.reason_pending
 sched.resource_summary
   (string) human readable overview of assigned resources
 
+user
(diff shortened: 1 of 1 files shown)


chu11:
oh crap, i noticed a typo! :-)   I'll send a PR.

garlick:
So much for my awesome review skills :-(

PR #263, opened 2020-09-02
rfc27: Add queue name as special case annotation

chu11:
i think `queue` name is common enough we can go ahead and define it as a known special case


dongahn:
LGTM!

PR #265, opened 2020-09-23
rfc20: replace R_lite "node" key with execution.nodelist

grondo:
Currently we're working on adding support for carrying along hostnames instead of/in addition to ranks in flux-core. I started with adding the `"node"` key to `R_lite` object in `R`, but this will make every `R` on a normal system explode in size, since every rank entry in `R_lite` will now need to be separated.

This experimental addition to RFC 20 adds an alternative -- a `nodelist` key in the `execution` section of `R` version 1. If present, this key is a list of hostnames which directly correspond to the rank entries in the `R_lite` array. E.g. on a system where `fluke[0-16]` correspond to broker ranks `[0-16]`, and we have the `R`

```json
{"version":1,"execution":{"R_lite":[{"rank":"0,2-3","children":{"core":"0-3"}}]}},
```

Then `nodelist` would be `["fluke0","fluke2","fluke3"]`, i.e. the first rank in `R_lite` corresponds to `fluke0` the second to `fluke2`, and so on.

Besides allowing better natural compression, this scheme would allow the host list to be tacked on as a post-processing step (possibly simplifying changes in code now that will need to be replaced when Rv2 is implemented).

Another benefit is that this scheme will make it simpler for `job-info` and other users of `R` to pick off the nodelist from an `R` object. Instead of iterating all `R_lite` entries, the `nodelist` key can simply be read. (Perhaps that is not a strong argument since `job-info` already must iterate `R_lite` to get the rank list)

Finally, separating the hostnames from the rank information might make a better (i.e. more easily written) initial "config file" syntax (if we decide to go that route)

Anyway, just throwing this out there to see what people think. I have the per-R_lite-nodename implemented so no big deal if this is not acceptable.

### DIFF
--- a/spec_20.rst
+++ b/spec_20.rst
@@ -125,9 +125,9 @@ the format version.
 Execution
 ~~~~~~~~~
 
-The value of the ``execution`` key SHALL contain at least the single key
-``R_lite``, with optional keys ``starttime`` and ``expiration`` and other keys
-reserved for future extensions.
+The value of the ``execution`` key SHALL contain at least the keys
+``R_lite``, and ``nodelist``, with optional keys ``starttime`` and
+``expiration``. Other keys are reserved for future extensions.
 
 ``R_lite`` is a strict list of dictionaries each of which SHALL contain
 at least the following two keys:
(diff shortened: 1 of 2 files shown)


SteVwonder (approved):
LGTM!  Keeping the compressed rank format is quite nice.

On the parsing end of this, it seems totally reasonable to traverse the expanded idset at the same time as the nodelist. So that is quite nice as well. 👍

grondo:
Do you think Fluxion could easily be switched to use this scheme instead of the `node` key in each entry of `R_lite`. If so, this would really help on the tools side, since we won't have to parse both methods to get the hostname to rank mapping.

dongahn:
@grondo: I need to play with this. But at first glance, I don't see why this would be difficult to switch Fluxion to your proposed RV1 format at all.

dongahn:
@grondo or @SteVwonder: it will be great if one of you can create an issue ticket to add this support to flux-sched so that we can follow up.

garlick (approved):
It sounds like the language should be changed to make this mandatory rather than optional based on the discussion.  LGTM once that is changed.

grondo:
Yeah, let me change the nodelist to be mandatory, though then we might want to update the version number, since all previous Rv1 objects will now be non-conformant? (I'm fine though if we want to give this one a "pass" since we're rapidly breaking other stuff anyway)

I also updated the language to allow each entry in the `nodelist` array to optionally contain embedded `idset`, expandable by `idset_format_map(3)`. This allows even more space efficiency, and will be especially nice if someone is hand-crafting an R for testing.

grondo:
Ok, I've just update the branch with the following changes:

 * `nodelist` is now a required key in the `.execution` dictionary of `R` version 1
 * The `node` key in entries of `R_lite` is removed, since it is redundant with the mandatory `nodelist`

grondo:
Seems like there was consensus on the call so I've set MWP.

PR #268, opened 2020-09-30
rfc20: allow nodelist entries to use hostlist format

grondo:
Update description of the Rv1 `nodelist` to allow "hostlists" instead of host sets.
Reference rfc 29 for definition of the hostlist format.


garlick (approved):
Good!

SteVwonder (approved):
LGTM too!

PR #273, opened 2020-11-11
rfc27: add sched.prioritize RPC

garlick:
Add an RPC that allows the job manager to inform the scheduler of new priorities assigned to job(s) that have outstanding alloc requests.  It should be possible for the job manager to send _one_ of these requests to the scheduler after the fair share priority plugin finishes a batch of priority updates, so the scheduler receives the total order in one go.

This request would not be used in `single` mode.  In that case, if the single outstanding alloc request is no longer the highest priority job, the job manager must cancel its alloc request and let the top job take its place.

Also, the alloc request is modified to drop `t_submit` and redefine `priority` from the administrative priority to the queue priority.  (This assumes that the scheduler would no longer need the submit time because it will now be reflected in the queue priority)

### DIFF
--- a/spec_27.rst
+++ b/spec_27.rst
@@ -151,13 +151,10 @@ id
   (integer) job ID
 
 priority
-  (integer) job priority in the range of 0 through 31
+  (integer) queue priority in the range of 0 through 4294967295
 
 userid
   (integer) job owner
-
-t_submit
-  (double) job submit time in seconds since the UNIX epoch
 
 Example:
 
@@ -167,15 +164,13 @@ Example:
      "alloc": [
        {
          "id": 1552593348,
-         "priority": 20,
+         "priority": 43444,
          "userid": 5588,
-         "t_submit": 593011542.2359374
        },
        {
          "id": 1552599944,
-         "priority": 20,
+         "priority": 222,
          "userid": 5588,
-         "t_submit": 593069342.2334354
        }
      ]
    }
@@ -239,13 +234,10 @@ id
   (integer) job ID
 
 priority
-  (integer) job priority in the range of 0 through 31
+  (integer) queue priority in the range of 0 through 4294967295
 
 userid
   (integer) job owner
-
-t_submit
-  (double) job submit time in seconds since the UNIX epoch
 
 Example:
 
@@ -253,9 +245,8 @@ Example:
 
    {
      "id": 1552593348,
-     "priority": 20,
+     "priority": 53444,
      "userid": 5588,
-     "t_submit": 593011542.2359374
    }
 
 Upon receipt of the ``alloc`` request, the scheduler SHALL look up *jobspec*
--- a/spec_27.rst
+++ b/spec_27.rst
@@ -462,7 +462,52 @@ as described above.  If the specified job does not have a pending
 Note that receipt of a ``sched.cancel`` does not necessarily indicate
 that the *job* is canceled. For example, the job manager may cancel all
 outstanding ``sched.alloc`` requests in response to the queue being
-administratively disabled.
+administratively disabled, or to make room for higher priority jobs
+in ``single`` mode.
+
+
+Prioritize
+~~~~~~~~~~
+
+When jobs with outstanding ``sched.alloc`` requests are re-prioritized,
+the job manager notifies the scheduler by sending a ``sched.prioritize``
+request.  The request payload consists of a JSON object with the following
+REQUIRED key:
+
+jobs
+  (array) list of [id, priority] tuples
+
+Each tuple SHALL consist of a two element array, containing:
+
+[0]
+  (integer) job ID
+
+[1]
+  (integer) queue priority in the range of 0 through 4294967295
+
+Example:
+
+.. code:: json
+
+   {
+     "jobs":[
+       [49056579584, 444],
+       [57428410368, 298],
+       [63988301824, 343205],
+       [69675778048, 99]
+     ]
+   }
+
+
+Job IDs which cannot be correlated to a pending ``sched.alloc`` request
+may be safely ignored.
+
+No response is sent to the ``sched.prioritize`` request.
+
+.. note::
+    A job manager priority plugin MAY initiate a priority update of many
+    jobs at once.  The job manager captures these updates in a single
+    ``sched.prioritize`` request.
 
 
 Free


grondo (approved):
LGTM.

As an aside, the removal of `t_submit` made me wonder if `userid` could also be removed from the alloc request. I guess one reason userid is required might be that a scheduler may have some resources that are restricted to a set of users? But I suppose if so the scheduler could fetch the userid for a job as it must do for jobspec and R...

chu11:
LGTM too

dongahn:
I will be looking at this soon.

spec_27.rst
```diff
@@ -151,14 +151,11 @@ id
   (integer) job ID
 
 priority
-  (integer) job priority in the range of 0 through 31
+  (integer) queue priority in the range of 0 through 4294967295
```
dongahn, line 154: I wonder if we want to keep two keys here -- one for the administrative priority and multi-factor priority (or secondary priority). Some schedulers may want to respond differently on either priority type. Any scheduler would immediately act on an administrative priority change. But some schedulers may not act on the multi-factor priority immediately.

spec_27.rst
```diff
@@ -462,7 +462,52 @@ as described above.  If the specified job does not have a pending
...
+
+
+Prioritize
+~~~~~~~~~~
+
+When jobs with outstanding ``sched.alloc`` requests are re-prioritized,
+the job manager notifies the scheduler by sending a ``sched.prioritize``
+request.  The request payload consists of a JSON object with the following
+REQUIRED key:
+
```
dongahn, line 476: We can perhaps introduce a priority type as a key to distinguish administrative priority vs. secondary priority changes?

garlick:
Not sure on the userid question.  We could drop it if we think we won't need it.  However, if we think a scheduler might need it, it's probably reasonable to pass it in with the alloc request since it's just an integer, and fetching it from the KVS would require the eventlog to be parsed.

Maybe @dongahn has a thought here?

spec_27.rst
```diff
@@ -151,14 +151,11 @@ id
   (integer) job ID
 
 priority
-  (integer) job priority in the range of 0 through 31
+  (integer) queue priority in the range of 0 through 4294967295
 
 userid
   (integer) job owner
```
dongahn, line 158: I thought we still need `t_submit` to be able to reconstruct the scheduling queue for scheduler reload... Maybe I'm missing something.

grondo:
> However, if we think a scheduler might need it, it's probably reasonable to pass it in with the alloc request since it's just an integer, and fetching it from the KVS would require the eventlog to be parsed.

That's a good point. We should keep eyes open for any other keys or data that might possibly be needed by a scheduler and consider including those as well (I think you had even proposed including some part of the jobspec in the alloc request?)

dongahn:
> Also, the alloc request is modified to drop t_submit and redefine priority from the administrative priority to the queue priority. (This assumes that the scheduler would no longer need the submit time because it will now be reflected in the queue priority)

Ah... I see this better. 

But still my position would be to keep `t_submit` if it isn't too expensive. There may be a scheduler which ignores the secondary priority and schedule with FCFS or other queuing policies (e.g., for debugging etc). Funneling many of the scheduling parameters into one number from job-manager may be too confining for future scheduler innovation?

dongahn:
> That's a good point. We should keep eyes open for any other keys or data that might possibly be needed by a scheduler and consider including those as well (I think you had even proposed including some part of the jobspec in the alloc request?)

I agree. IMHO, passing the info that can be used by different schedulers (albeit judiciously) to the scheduler will lead to more scheduling advancements.

spec_27.rst
grondo: I think the idea is that the queue priority is a function of admin priority, and thus including the administrative priority is redundant. Therefore the logic to enact a priority change should be fully within the priority plugin, and the scheduler would then be able to act on any change thereof.

spec_27.rst
grondo: t_submit is used as a factor in the calculation of queue priority. So the queue can be reconstructed solely on queue priority value. Does that make sense? is there a reason a scheduler would need to separately know `t_submit`?

grondo:
> But still my position would be to keep t_submit if it isn't too expensive. There may be a scheduler which ignores the secondary priority and schedule with FCFS or other queuing policies (e.g., for debugging etc). Funneling many of the scheduling parameters into one number from job-manager may be too confining for future scheduler innovation?

In the case of FCFS the default priority would just be an encoding of t_submit to a priority integer, so they will essentially be the same value. The scheduler innovation could always occur on the priority plugin side, which will have access to all job data known by the job-manager, and the queue order would be kept consistent between job-manager/job-info and the scheduler (the latter is an important point)

spec_27.rst
dongahn: Hmmm, how would an administrative priority change come to the scheduler, though?

garlick: The default priority plugin would combine t_submit and administrative priority into this "queue priority".  When the administrative priority changes, the queue priority changes. Other priority plugins would consider these as weighted factors.

dongahn:
> In the case of FCFS the default priority would just be an encoding of t_submit to a priority integer, so they will essentially be the same value. 

Agreed.

> The scheduler innovation could always occur on the priority plugin side, which will have access to all job data known by the job-manager, 

This means for a scheduler to be able to enforce a different queuing policies, we have to reload the scheduler and the priority plugin. And this is to provide the following property. 

> and the queue order would be kept consistent between job-manager/job-info and the scheduler (the latter is an important point)

Glad that you mentioned this since I had a question on this. Is this really necessary?

spec_27.rst
garlick: The assumption was that the scheduler shouldn't need to separately track the administrative priority if it is represented in the value of the queue priority.  (We are now calling it "queue priority" to underscore that there is only one priority used to order the queue that includes t_submit and administrative priority, and possibly other factors).

grondo:
> This means for a scheduler to be able to enforce a different queuing policies, we have to reload the scheduler and the priority plugin. And this is to provide the following property.

Good question. Actually, I think it would require reloading just the priority plugin, which could re-prioritize all jobs if reloaded. We could also allow the job-manager to have multiple priority plugins loaded and switch between them fairly seamlessly.

> Glad that you mentioned this since I had a question on this. Is this really necessary?

I'm not the authority on that question, however my opinion is that when users list the job queue they would prefer seeing the highest priority job at the top of the queue. If the scheduler can ignore the priority then there would be no way to offer this functionality short of a scheduler-specific query.

spec_27.rst
dongahn: I guess this boils down to my comment far below:

> This means for a scheduler to be able to enforce a different queuing policies, we have to reload the scheduler and the priority plugin. And this is to provide the following property.

>> and the queue order would be kept consistent between job-manager/job-info and the scheduler (the latter is an important point)

grondo: No, I think you'd just be able to reload the priority plugin, or switch to a different implementation. The protocol described in this PR allows the priority plugin to reprioritize jobs and update the scheduler with the new queue order.

grondo:
To give a high level summary:  There are two main benefits of the proposed priority plugin design. The first is that the overall job priority is known to flux-core and can be used in core-provided commands and utilities (the reuse benefit). The second is that we can test prioritization without integrated testing with flux-sched or other "advanced schedulers" loaded. (the encapsulation benefit)

If we have to leak factors used in the priority plugin to the scheduler, then the abstraction is broken and we essentially get neither of these benefits. In that case perhaps we need to rethink the design.

garlick:
>> and the queue order would be kept consistent between job-manager/job-info and the scheduler (the latter is an important point)

>> Glad that you mentioned this since I had a question on this. Is this really necessary?

It might be good to distinguish between the queue (in priority order)  and "the plan" made bye the scheduler.

The queue in priority order would be what users see and list the queue.

The plan gets made and remade based on resource availability, relationships between resources, and higher order goals like utilization or throughput.

Keeping the input to the plan-making the same as what's listed by the tools may be beneficial.

garlick:
Would it be accurate to say that "queuing policy" could be provided by the priority plugin, and planning is provided by the scheduler?  I think we are positing that "FCFS" and "fair share" are queuing policies and they can be provided by the priority plugin.  Something like "FCFS with backfill" maybe is not really a queuing policiy - more like a combination of FCFS queuing policy and backfill planning?

dongahn:
I feel we are getting a really good discussion here. So thank you @garlick and @grondo. 

In terms of the high level architecture:

1. job-manager will keep one master queue
2. a priority plugin will be used to order the jobs in this master queue (but priority plugin may not have all the information about the scheduler loaded)
3. the job-manager will send alloc requests to the scheduler in this priority-sorted order
4. the loaded scheduler will receive and enqueue those jobs into one or more of its queues and annotate the jobs such as way that flux-core job listing utility can use them to present the jobs to the users in a way that makes sense
5. the scheduler applies scheduling policies (e.g. queuing and backfilling policies) for jobs in each of these queues, which can lead to out of order execution

I believe we had two main discussion points:
A. how much information should job-manager pass to the scheduler to do its scheduling (4 and 5)
B. whether the priority plug-in should know what the loaded scheduler will do and have priority calculation based on that (2)

dongahn:
We had a good discussion at our 11AM team meeting. Decision:

- We will land the current proposal as this addresses all the short term use cases
- If our scheduler requires more information (beyond a single priority number) to create a better "plan", it will be proposed and changes should be straightforward

dongahn (approved):

garlick:
Thanks! I'll set MWP.  This is good - we can keep some momentum going in this direction.

PR #274, opened 2020-11-14
rfc21: refine PRIORITY state transitions

garlick:
This PR changes the description of the `priority` event that transitions job state from PRIORITY to SCHED, describing it as assigning the initial priority (not implying it is used for updates), and dropping the note that says it is not written to the job eventlog.  It also adds the `flux-restart` event which will be posted to all active jobs on instance restart, and will trigger the transition from SCHED back to PRIORITY.  Finally, it removes the no-cycle rule, quoting some rationale from #3311 in the commit message.

This change was discussed in #3311 and referred to again in #3342


chu11:
i think 

```
The job is waiting for a queue priority to be assigned by the job manager priority plugin. Upon priority assignment, the job manager posts a priority event. The state transitions to SCHED.
```

should be altered to "logs the priority event", similar to other transition language.

chu11:
overall LGTM, but wanted to make sure I understand things correctly.  Lets say eventlog is

```
submit
priority
depend
```

if the `job-manager` reloads, we'd see job-manager add to the eventlog to:

```
submit
priority
depend
flux-restart
priority
sched
```

garlick:
Right except the sequence is `depend` -> `priority` -> `sched`, so

```
depend
priority
sched
flux-restart
priority
sched
```

chu11:
oops on the `submit` -> `depend` -> `priority`.  But after an `alloc`, the job is running, so presumably we don't need to go back to priority?  

so I think it should be

```
submit
depend
priority
flux-restart
priority
```

if job is pending and job-manager restarted.  But if the job is running, as in an eventlog


```
submit
depend
priority
alloc
```

We shouldn't log `flux-restart`?  and it shouldn't go back to priority?  the state diagram in your RFC21 suggests this

garlick:
(just pushed a fix to the wording suggestion above)

Yes you are right (my bad!) on everything in your last post!

chu11 (approved):
LGTM!

garlick:
Thanks!  I'll set MWP

PR #277, opened 2020-11-30
rename administrative priority to "urgency"

garlick:
As discussed in flux-framework/flux-core#3377, the terms "queue priority" and "administrative priority" are a bit cumbersome/ambiguous.  This PR proposing changing the terms to "priority" and "urgency".

We had not written down the idea of urgency=0 for hold and urgency=31 for expedite so I went ahead and added a mini RFC about urgency and the meaning of these special values.  Hope that's not too much overkill.

### DIFF
--- a/spec_30.rst
+++ b/spec_30.rst
@@ -1,0 +1,70 @@
+.. github display
+   GitHub is NOT the preferred viewer for this file. Please visit
+   https://flux-framework.rtfd.io/projects/flux-rfc/en/latest/spec_30.html
+
+30/Job Urgency
+==============
+
+This specification describes the Flux job urgency parameter.
+
+-  Name: github.com/flux-framework/rfc/spec_30.rst
+
+-  Editor: Jim Garlick <garlick@llnl.gov>
+
+-  State: raw
+
+
+Language
+--------
+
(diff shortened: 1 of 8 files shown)


spec_30.rst
```diff
@@ -0,0 +1,69 @@
...
+the lowest possible priority, bypassing the job manager priority plugin.
+Although it remains in the queue while pending, a held job SHALL NOT be
+assigned resources.
+
+A value of 31 indicates that the job should be *expedited*.  An expedited job
+is assigned the higest possible priority, bypassing the job manager priority
+plugin.
+
+If the urgency is *updated*, the job manager priority plugin SHALL be notified
+os it can adjust the job's priority.
```
chu11, line 67: typo `os` -> `as`

chu11:
reading through these changes, i like it.  the `priority` state / event makes more sense, as does the future `priority-update` event discussed in https://github.com/flux-framework/flux-core/issues/3342 (i haven't written up the rfc change yet).

spec_30.rst
garlick: Thanks I forced a push.

grondo:
I agree, nice suggestion! It turned out a lot nicer than a job nice parameter! ;-)

garlick:
Nicer than nice!  Wow that's high praise. :-)

grondo (approved):

PR #278, opened 2020-11-30
rfc21: Add priority-update event

chu11:
Problem: There is no way to know if a job's priority has changed
after the PRIORITY state / after the "priority" event.

Solution: Add a new priority-update event, which can advertise that
the priority of a job has changed.

### DIFF
--- a/spec_21.rst
+++ b/spec_21.rst
@@ -253,6 +253,21 @@ userid
    {"timestamp":1552593547.411336,"name":"urgency","context":{"urgency":0,"userid":5588}}
 
 
+Priority-Update Event
+^^^^^^^^^^^^^^^^^^^^^
+
+Job's priority has changed.
+
+The following keys are REQUIRED in the event context object:
+
+priority
+   (integer) Priority in the range of 0-4294967295.
+
+.. code:: json
+
+   {"timestamp":1552593547.411336,"name":"priority-update","context":{"priority":1001}}
+
+
 Alloc Event
 ^^^^^^^^^^^
 


garlick:
This should be a journal only event, which is why I shied away from putting it in the RFC before.  Thoughts?

chu11:
Per discussion in https://github.com/flux-framework/flux-core/issues/3342 we decided to put it in the eventlog?

garlick:
Ohhhhhhhh, forgot about that discussion.

Actually do we need separate `priorty` and `pririty-update` events?   Can we just go with `priority`?

chu11:
> Actually do we need separate priorty and pririty-update events? Can we just go with priority?

I thought about that while developing. I think it'd be doable, but it introduces a minor complication that the `priority` event says we move to the `SCHED` state.  The `priority-update` event doesn't.  So I think it'd be better to separate it out to keep things simpler.  Last thing we want to deal with while processing the eventlog during a restart is how to deal with this special case?

garlick:
Well, OK.  I didn't think that was ambiguous since we define all possible state transitions and the events that drive them in the RFC, but if you think we need it I'm OK with adding it.  I'll go ahead and approve.

Sorry for forgetting our past discussion!

garlick (approved):

chu11:
> I thought about that while developing. I think it'd be doable, but it introduces a minor complication that the priority event says we move to the SCHED state. The priority-update event doesn't. So I think it'd be better to separate it out to keep things simpler. Last thing we want to deal with while processing the eventlog during a restart is how to deal with this special case?

Oh yeah, I remember now.  The hiccup I hit was setting the `t_priority` timestamp.  Would I set the `t_priority` for the first `priority` event only / the first one after a `flux-restart`?  Can totally be dealt with, but figured it'd be easier to leave out the ambiguity.

garlick:
There are timestamps in the eventlog so a user can find when each event occurred if they need to.

Should we drop `t_priority` from job-info as a broken-out thing?  Seems like it wouldn't be a hot item for users.

grondo:
Good point. I think users (or maybe more system administrators) may more be interested in the time deltas between the different states, not the timestamps, and we could develop a different tool for that (or have flux-jobs read eventlogs directly)

chu11:
> Should we drop t_priority from job-info as a broken-out thing? Seems like it wouldn't be a hot item for users.

ok, sounds like a plan.  I'll leave this PR for the time being, until I prototype and determine there's no other gotchas.

chu11:
> Good point. I think users (or maybe more system administrators) may more be interested in the time deltas between the different states, not the timestamps, and we could develop a different tool for that (or have flux-jobs read eventlogs directly)

should we consider if `job-info` and `flux-jobs` should drop support of `t_submit`, `t_depend`, etc. as well?

grondo:
No those seem fine to keep as is since they are static (IMO)

chu11:
> No those seem fine to keep as is since they are static (IMO)

then perhaps `t_sched` should be dropped as well?

chu11:
moving issue of removing `t_priority` and such to https://github.com/flux-framework/flux-core/issues/3395

chu11:
prototyping i think re-using the `priority` event will work out, closing

PR #279, opened 2020-12-03
rfc27: add jobspec to sched.alloc request

garlick:
Change the definition of the sched.alloc request to include
jobspec (minus environment), since that can now be trivially
included by the job manager, and will save a KVS lookup per job.


grondo (approved):
Thanks! LGTM and a nice little improvement.

This change also removes the text that said

>  The `sched.free` request implicitly refers to *R* in the KVS by job ID

I assume this was intentional since it is obvious, however, I mention it here in case it was accidental.

garlick:
Thanks!

Yeah I figured that was no longer worth of mention in the intro, since it is now just one message that implicitly refers to a KVS object.

grondo:
Hm, Travis is stuck, maybe time to update this repo to GH Actions as well.

PR #281, opened 2020-12-16
rfc12: update overlay security with new PMI strategy

garlick:
Problem: RFC 12 description of overlay secrurity is out of date.

Explicitly mention CURVE and GSSAPI as reasonable ZMQ security methods,
instead of implying that anything but NONE is OK.

Describe the shared CURVE cert used by a system instance.

Describe the new CURVE public key exchange strategy for PMI-booted
Flux instances.

Drop the paragraph on security EPGM, as we don't support that anymore.

### DIFF
--- a/spec_12.rst
+++ b/spec_12.rst
@@ -230,7 +230,9 @@ within an instance.
 Users other than the instance owner SHALL NOT be permitted to connect
 to an instance’s overlay networks. Since overlay networks are implemented
 using the ZeroMQ messaging library, these connections SHALL be secured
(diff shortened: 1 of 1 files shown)


garlick:
This PR just updates the security RFC to track recent changes in the broker for CURVE key exchange.  It seems to have been lost in the cracks.  Would someone mind giving it a quick review and merge?

grondo (approved):
LGTM!

garlick:
Thanks!

PR #283, opened 2021-01-11
RFC27: Refactor hello RPC to be streaming

chu11:
Per discussion in https://github.com/flux-framework/flux-core/pull/3447, we will update the hello protocol to be a streaming RPC instead of a single request/response with an array of jobs.

I went back and forth on how to write the language for this section.  I'm still not 100% about it.

Also, I noticed that we never documented that `t_submit` is in the hello response.  So I added it.

### DIFF
--- a/spec_27.rst
+++ b/spec_27.rst
@@ -138,11 +138,11 @@ Hello
 ~~~~~
 
 Before any other RPCs are sent to the job manager, the scheduler SHALL
-send an empty request to ``job-manager.sched-hello`` and wait for a response.
-
-The response payload SHALL consist of a JSON object with one key: ``alloc``,
-whose value is an array of zero or more jobs with allocated resources.
-Each array entry SHALL be a JSON object with the following REQUIRED keys:
+send an empty request to ``job-manager.sched-hello`` with the
+FLUX_MSGFLAG_STREAMING flag set.  The job manager SHALL send one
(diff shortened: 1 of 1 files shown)


spec_27.rst
```diff
@@ -138,11 +138,11 @@ Hello
...
-send an empty request to ``job-manager.sched-hello`` and wait for a response.
-
-The response payload SHALL consist of a JSON object with one key: ``alloc``,
-whose value is an array of zero or more jobs with allocated resources.
-Each array entry SHALL be a JSON object with the following REQUIRED keys:
+send an empty request to ``job-manager.sched-hello`` with the
+FLUX_MSGFLAG_STREAMING flag set.  The job manager SHALL stream
+responses representing jobs with allocated resources.  Each response
+payload SHALL consist of a JSON object with the following REQUIRED
+keys:
```
garlick, line 145: instead of
> The job manager SHALL stream responses representing jobs with allocated resources.

Would it be slightly better to say:
> The job manager SHALL send one response message for each job with allocated resources.

garlick (approved):
LGTM, just one minor rewording suggestion that you can take or leave.

chu11:
@garlick thanks, re-pushed with tweak per your suggestion.

garlick:
Great! I'll set MWP.

PR #284, opened 2021-01-19
rfc27: support mode=limited in sched.ready

chu11:
Per issue https://github.com/flux-framework/flux-core/issues/2070, support mode=N in the `job-manager`.  Update RFC27 to allow a `mode=N` in the `sched.ready` protocol.

Not 100% if this is the best way to go about things, but thought I'd setup this PR for discussion / consideration of this change:

- should "mode" be an integer instead of a "string"?  We treat "unlimited" to mean 0 and obviously "single" is just numeral 1.  Seems silly to keep it a string.

  - But we may want to keep as a string for future flexibility.  IMO, this is a good reason to keep it.
  - Changing it to a numeral we'd have to break ABI in `libschedutil` in `flux-core`

- Another possibility, do we want to set `N` via an option?  So instead of `mode=2`, we could do `mode=N=2`.  Perhaps in the future we might want to do, `mode=N=1000,someflag` or `mode=foo=bar,N=1000`.  We could keep the "flags", `single` and `unlimited` as special case flags meaning `N=1` and `N=0`.

- If we keep `N` a string, should we define a cap on N in RFC27?  I'm assuming `int` range is reasonable?  Or `unsigned int` range?  Or something much smaller to be reasonable?

### DIFF
--- a/spec_27.rst
+++ b/spec_27.rst
@@ -214,7 +214,10 @@ Example:
 
    {"mode":"limited","limit":42}
 
-The response payload SHALL be empty.
+The response payload is a JSON object with the following REQUIRED keys:
+
+count
+  (integer) current queue depth
 
 After responding to the ``job-manager.sched-ready`` request, the job manager
 MAY immediately begin sending ``sched.alloc`` and ``sched.free`` requests.
--- a/spec_27.rst
+++ b/spec_27.rst
@@ -194,20 +194,25 @@ mode
 
 The mode string SHALL be one of the following:
 
-single
-  The job manager SHALL send a ``sched.alloc`` request only when there are
-  no outstanding ``sched.alloc`` requests.  This mode is only useful for simple
-  schedulers that run jobs strictly in the job manager queue order.
-
 unlimited
   The job manager SHALL send a ``sched.alloc`` request for all jobs in SCHED
   state, with no limit on concurrency.
+
+limited
+  The job manager SHALL limit the number of concurrent ``sched.alloc``
+  requests to value specified by the ``limit`` key (described below).
+
+The following key is REQUIRED for ``limited`` mode only:
+
+limit
+  (integer) The number of concurrent ``sched.alloc`` requests that can
+  be sent.  ``limit`` can be in the range of 1 to 2147483647.
 
 Example:
 
 .. code:: json
 
-   {"mode":"unlimited"}
+   {"mode":"limited","limit":42}
 
 The response payload SHALL be empty.
 


grondo:
Hm, I don't remember all the details of the original design, but I don't think the mode should be `=N`, but should be something like `mode=limited`, and `N` should be an option or perhaps even negotiated with an additional RPC.

I found part of the original credit-based proposal in flux-framework/flux-core#1697, though I'm not sure if @garlick thinks this scheme is still necessary. Copying from there:

> A job manager (or other sched client) may only send one allocate request per credit, then must request more credits. The scheduler may throttle the flow of allocate requests and limit concurrency by delaying the response to a request for credits. It must implement a per-client count of credits issued, and return an error to allocate requests when there are no credits.
>
> To allow high priority jobs to be accepted when the scheduler has already issued all its credits, a request for credits may contain a priority. If credits are not available, the scheduler may cancel lower priority requests to allow the higher priority requests to be accepted. Requests for credits are handled first in priority order, then in FCFS order.
>
>To avoid deadlock, free requests must not be subject to credits


If going with the simpler approach proposed here, though, the value for `N` would need to be tunable at runtime. (Therefore I still propose `mode=limited`. A future credit-based scheme could be `mode=credit`)

garlick:
Thanks for dredging that old proposal up.  I was trying to avoid an issue where a deluge of jobs would cause the system to become unresponsive, but now that we have the real system I'm not sure that's where we should focus, at least not until we have a clear indication that such work would be beneficial.

On this proposal, I think @grondo's suggestion of `mode=limited` with a limit argument makes sense.   Could we avoid breaking the libschedutil ABI by allowing `schedutil_ready()` to be called with with `mode=limited=42` or similar, and translate that to the protocol:
```json
{"mode":"limited","limit":42}
```

grondo:
Since this would practically be the scheduler's "queue depth", I wonder if we should allow it to be changed at runtime? Though I suppose, for now at least, re-loading the scheduler with a new alloc limit is possible and would allow the parameter to be modified at runtime.

chu11:
Well right now `libschedutil` is `int schedutil_ready (schedutil_t *util, const char *mode, int *queue_depth);`, just a single string "mode".

Perhaps if the user passed in `limited,limit=42`

grondo:
> Perhaps if the user passed in "limited,limit=42"

I'm sure I'm missing the point, but isn't `limited,limit=42` a little redundant?

garlick:
> Since this would practically be the scheduler's "queue depth", I wonder if we should allow it to be changed at runtime?

Should we separately add a way to query and set the limit at the job manager level?  Or do we think that such changes would always need to be initiated by, or filtered through, the scheduler?  (just thinking it might be possible to minimize the surface area of scheduler/job manager protocol), and code duplicated in each scheduler.

grondo:
I'm unsure. Seems like the queue-depth of e.g. a backfill scheduler would need to be synchronized with this limit?

Maybe more design is required here.

chu11:
> I'm sure I'm missing the point, but isn't limited,limit=42 a little redundant?

yeah, i guess it is.  I was mostly following @garlick's `{"mode":"limited","limit":42}`.

So perhaps we allow

`mode=limited=N` (values <= 0 are an error, values > some cap an error??)
`mode=unlimited`
`mode=single`

`single` is obviously there only if we care about maintaining it for legacy reasons.

grondo:
Ah, good point, single just becomes `limit=1`. :+1:

chu11:
re-pushed, going with

```
mode=limited=N
mode=unlimited
```

deciding to eliminate `mode=single`.  We could add back if we wanted to maintain legacy support.

I also said that `N` is in the range of 0 to 2^31-1 (signed 32 bit int) range, defining `0` to be the same as `unlimited`.  I guess the upper bound is less important, I just picked one.

spec_27.rst
```diff
@@ -194,10 +194,11 @@ mode
 
 The mode string SHALL be one of the following:
 
-single
-  The job manager SHALL send a ``sched.alloc`` request only when there are
-  no outstanding ``sched.alloc`` requests.  This mode is only useful for simple
-  schedulers that run jobs strictly in the job manager queue order.
+limited=N
+
```
garlick, line 198: Should we just parse `unlimited=N` in `schedutil_ready()` and break out the mode and N in the protocol?

Also, delete extra line.

spec_27.rst
```diff
@@ -207,7 +208,7 @@ Example:
 
 .. code:: json
 
-   {"mode":"unlimited"}
+   {"mode":"limited=42"}
 
 The response payload SHALL be empty.
```
garlick, line 213: The response actually isn't empty. It contains the queue depth.  It doesn't seem all that useful but we'd have to break the ABI to make the code match the RFC so maybe we should just fix the RFC to match reality? 

(Not added by this PR, but might as well fix it)

chu11:
re-pushed trying to address @garlick comments above.  We're going with `mode=unlimited | limited` and if mode is limited, `limit=N` is a required additional payload key.  Not sure if how i worded/organized the RFC changes is best.

Maybe i should instead have a new header that says "if mode is "limited", then the following keys are REQUIRED"?

spec_27.rst
```diff
@@ -194,20 +194,28 @@ mode
...
+
+In addition to the above REQUIRED keys, the request may include
+the OPTIONAL key:
+
+limit
+  (integer) Specifies the number of concurrent ``sched.alloc``
+  requests can be sent for mode ``limited`` specified above.  The
+  ``limit`` key is REQUIRED when mode is ``limited``.  ``limit`` can
+  be in the range of 1 to 2147483647.
+
```
garlick, line 213: Hmm, it read like `limit` caps total number of alloc requests.  Should the description be something more like
> The job manager SHALL limit the number of concurrent ``sched.alloc`` requests to value specified by the `limit` key.

And I agree that `limit` should be REQUIRED for `limited` mode.  Maybe just change the "in addtiion" lead-in to something like
> The following key is REQUIRED for ``limited`` mode only:

chu11:
re-pushed with tweaks per comments above

garlick (approved):
LGTM!  Thanks!

chu11:
thanks, setting MWP

PR #287, opened 2021-02-16
remove outdated terminology, update keepalive description in RFC5

grondo:
This PR does two things, I'm fine breaking it up into separate PRs if desired.

 1. Update outdated terminology similar to flux-framework/flux-core#3456
 2. reword keepalive passages in RFC5 to de-emphasize running/sleeping transition keepalive messages


garlick (approved):
Good improvements - thanks for doing this!

Apologies for the wordiness of RFC 5.  I think that needs to be cut way down, but that's probably for another day.

PR #290, opened 2021-03-17
rfc21: add dependency-add/remove events

grondo:
This simple addition to RFC21 allows management of dependencies via `dependency-add` and `dependency-remove` events. This enables some simple dependencies to be handled internally in the job manager, possibly via plugins. 

In the common case of no dependencies, there would be no change to existing behavior. If a job manager plugin wants to add a dependency to a job, it can issue the `dependency-add` event while in the depend state, e.g. via the `job.state.depend` callback. This will add a dependency reference to the job. When the job dependency is satisfied (e.g. at a future time, or when a job on which the job depends becomes inactive) the plugin can issue the `dependency-remove` event, the reference will be removed and the job manager will issue the `depend` event.

If a dependency service is being used instead, then a single reference can be issued for the dependency service and subsequently removed when RPC response from the service is received.

An additional benefit of issuing dependency-add/remove events to the eventlog is that outstanding dependencies can be gathered and listed by the `job-list` module. Also, it may be interesting to have dependencies logged for debugging and post-mortem analysis.

There is still some question as to how to handle these events on a job manager restart.

### DIFF
--- a/spec_21.rst
+++ b/spec_21.rst
@@ -89,10 +89,9 @@ NEW
    transitions the state to DEPEND.
 
 DEPEND
-   The job is blocked waiting for dependencies to be satisfied. The job manager
-   makes a request to the dependency service and receives a response once
-   the job’s dependencies are satisfied, then logs the ``depend`` event.
-   The state transitions to PRIORITY.
+   The job is blocked waiting for dependencies to be satisfied. Once all
+   dependencies have been satisfied the ``depend`` event is logged and
+   the state transitions to PRIORITY.
 
 PRIORITY
    The job is blocked waiting for a priority to be assigned by the job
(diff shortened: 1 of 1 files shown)


grondo:
Current thought on restart, though not covered in this PR:

On restart, the job manager adds dependency references as they are replayed from the eventlog. If the job had previously passed through the DEPEND state, then all dependency references that were added will have been removed, and therefore after replay there will no longer be any remaining dependency references. After replay the job will skip calling the `job.state.depend` plugin callback, and therefore plugins will not re-issue dependency-add events.

If the job previously had outstanding dependency references, then the job will be in the DEPEND state after replay, and the `job.state.depend` callback may be called for the job. Any dependency plugin could issue a new `dependency.add` event. If the event matches a previously issued dependency, then the event will be ignored (i.e. `dependency.add` is not a _counting_ reference). This allows any loaded dependency plugins to become aware of the outstanding dependency and resolve them after a restart.

garlick (approved):
LGTM!

SteVwonder (approved):
LGTM!  I really like this design overall.  Makes it easy for the vendor-specific plugin I'd like to write to add a dependency on the fly. Also handles the flux-core-only versus flux-core + arbitrary dependency module/plugin split very nicely.

This isn't specified in the RFC so this question shouldn't hold up merging of the PR, but just to make sure I understand the proposal, if a jobspec has dependencies that follow the RFC26 format, the idea would be for a single `dependency-add` to be emitted for the dependency module/plugin that handles RFC26 dependencies?  Or would one `dependency-add` be emitted for each RFC26 dependency declared in the jobspec?

grondo:
Thanks @SteVwonder!

> the idea would be for a single dependency-add to be emitted for the dependency module/plugin that handles RFC26 dependencies? Or would one dependency-add be emitted for each RFC26 dependency declared in the jobspec?

I think this could go either way. I would lean toward a single `dependency-add` when using a remote service so that the service can send one response when all dependencies are satisfied. O/w, you either have the service sending multiple responses to resolve each dependency (and job manager plugin and service have to agree on names), or if you are waiting for a single response, then there are some outstanding dependencies in the eventlog that may have actually already been satisfied.

grondo:
If this is implemented, the first thing I was going to try was adding a plugin to handle begin-time as a dependency. You could also support simple jobid dependencies with a fairly simple job manager plugin. When advanced dependencies are required, this plugin would be replaced with a plugin that uses the dependency service.

We have to determine if the simple jobid dependencies and/or begin-time dependency should use the same  jobspec schema as RFC26 (which I need to look back at)

grondo:
Setting MWP. Thanks!

PR #292, opened 2021-06-03
rfc26: add background, divide into sections, add simple dependencies

garlick:
Since RFC 26 was originally written, it was decided to relax the dependency object spec so that only "scheme" and "value" keys were required.

I felt the document could benefit by adding a bit of background, especially calling out the new `dependency-add` and `dependency-remove` job events, and introducing the recent development of job manager plugins providing the hook for adding a new dependency scheme

Also, I added descriptions of the new simple dependency schemes proposed in flux-framework/flux-core#3696

I'm marking this as a WIP, indicating that it's sort of a straw man proposal and I'm happy to help rework it in another form or turn this PR over to others who have been working on the code and probably have deeper insights to offer.

### DIFF
--- a/spec_26.rst
+++ b/spec_26.rst
@@ -77,7 +77,7 @@ scheme
 value
   (string) semantics determined by the scheme.
 
-A dependency object MAY contain additional OPTIONAL keys-value pairs,
+A dependency object MAY contain additional OPTIONAL key-value pairs,
 whose semantics are determined by the scheme.
 
 in jobspec
@@ -110,6 +110,52 @@ Examples:
 
 This form SHOULD be translated by the command line tool to the object
 form above before being shared with other parts of the system.
+
+
+Simple Dependencies
+-------------------
+
+The following dependency schemes are built-in.
+
+after
+~~~~~
+
+``value`` SHALL be interpreted as the antecedent jobid, in any valid
+FLUID encoding from RFC 19.
+
+The dependency SHALL be satisfied once the antecedent job enters RUN state.
+If the antecedent job reaches INACTIVE state without entering RUN state,
+a fatal exception SHOULD be raised on the dependent job.
+
+
+afterany
+~~~~~~~~
+
+``value`` SHALL be interpreted as the antecedent jobid, in any valid
+FLUID encoding from RFC 19.
+
+The dependency SHALL be satisfied once the antecedent job enters INACTIVE
+state, regardless of result.
+
+afterok
+~~~~~~~
+
+``value`` SHALL be interpreted as the antecedent jobid, in any valid
+FLUID encoding from RFC 19.
+
+The dependency SHALL be satisfied once the antecedent job enters INACTIVE
+state, with a successful result.  If the antecedent job does not conclude
+successfully, a fatal exception SHOULD be raised on the dependent job.
+
+afternotok
+~~~~~~~~~~
+
+``value`` SHALL be interpreted as the antecedent jobid, in any valid
+FLUID encoding from RFC 19.
+
+The dependency SHALL be satisfied once the antecedent job enters INACTIVE
+state, with an unsuccessful result.  If the antecedent job concludes
+successfully, a fatal exception SHOULD be raised on the dependent job.
 
 
 OpenMP-style Dependencies
--- a/spec_26.rst
+++ b/spec_26.rst
@@ -24,6 +24,7 @@ Related Standards
 
 -  :doc:`14/Canonical Job Specification <spec_14>`
 -  :doc:`19/Flux Locally Unique ID <spec_19>`
+-  :doc:`21/Job States and Events <spec_21>`
 -  `OpenMP Specification <https://www.openmp.org/wp-content/uploads/OpenMP-API-Specification-5.0.pdf>`__
 -  `IETF RFC3986: Uniform Resource Identifier (URI) <https://tools.ietf.org/html/rfc3986>`__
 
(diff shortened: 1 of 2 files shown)


spec_26.rst
```diff
@@ -24,34 +24,165 @@ Related Standards
...
+--------------
+
+A job dependency SHALL be represented as a JSON object with the following
+REQUIRED keys:
+
+scheme
+  (string) name of the dependency scheme
+
+value
+  (string) semantics determined by the scheme.
```
grondo, line 78: The current implementation allows `value` to be a valid JSON value that is not an array or object. It might actually simplify things in some cases to only allow a string,(e.g. jobid parsing), but in other cases this could be a pain (e.g. timestamp)

spec_26.rst
```diff
@@ -24,34 +24,165 @@ Related Standards
...
+whose semantics are determined by the scheme.
+
+in jobspec
+~~~~~~~~~~
+
+Each dependency requested by the user SHALL be represented as an element in
+the jobspec ``attributes.system.dependencies`` array.  Each element SHALL
+conform to the object definition above.
+
+If job requests no dependencies, the jobspec ``attributes.system.dependencies``
```
grondo, line 90:
```suggestion
If job requests no dependencies, the key ``attributes.system.dependencies``
```

spec_26.rst
```diff
@@ -24,34 +24,165 @@ Related Standards
...
+
+on command line
+~~~~~~~~~~~~~~~
+
+On the command line, a job dependency MAY be expressed in a compact, URI-like
+form:
+
+::
+
+   scheme:value[&key=val...]
```
grondo, line 101: The implementation uses Python `urllib.parse()` which uses `?` as the separator between the "value" and the "query string", e.g. `"string:foo?type=out"`. Then either `&` or `;` can be used to separate key=value pairs in the query string, e.g.:

```
ƒ(s=4,d=0,builddir) grondo@asp:~/git/f.git$ flux mini run --dry-run --dependency="string:foo?type=out;scope=global" hostname | jq .attributes.system.dependencies
[
  {
    "scheme": "string",
    "value": "foo",
    "type": "out",
    "scope": "global"
  }
]
```

spec_26.rst
```diff
@@ -24,34 +24,165 @@ Related Standards
...
+
+Examples:
+
+-  ``afterany:ƒ2oLkTLb``
+-  ``string:foo&type=out``
+-  ``fluid:hungry-hippos-white-elephant``
+
+This from SHOULD be translated by the command line tool to the object
```
grondo, line 109:
```suggestion
This form SHOULD be translated by the command line tool to the object
```

spec_26.rst
```diff
@@ -24,34 +24,165 @@ Related Standards
...
+The following dependency schemes are built-in.
+
+afterany
+~~~~~~~~
+
+``value`` SHALL be interpreted as the antecedent jobid, in any valid
+FLUID encoding from RFC 19.
+
+The dependency SHALL be satisfied once the antecedent job enters CLEANUP
+state, regardless of result.
```
grondo, line 125: Small point: in the current WIP implementation  the dependency is satisfied when the job enters the INACTIVE state. The reason chosen is twofold:

 - Removes the necessity to check for a job already in the CLEANUP state when a dependent job is submitted
 -  Ensures stage-out or other activities performed in CLEANUP state of the antecedent job (love that term btw) have completed before starting the dependent job.

Both of these are encumbered by assumptions that may not be valid. It would be easy to change the WIP implementation if the description here should be following instead.

spec_26.rst
```diff
@@ -24,34 +24,165 @@ Related Standards
...
+~~~~~~~~~~
+
+``value`` SHALL be interpreted as the antecedent jobid, in any valid
+FLUID encoding from RFC 19.
+
+The dependency SHALL be satisfied once the antecedent job enters CLEANUP
+state, with an unsuccessful result.  If the antecedent job concludes
+successfully, a fatal exception SHOULD be raised on the dependent job.
+
+
```
grondo, line 147: A description of the `after` scheme is missing here.

grondo (commented):
This is great, thank you! This adds a lot of clarity.

I apologize this should have been done along with the work to add support for the `--dependency` option to flux-mini.

Just a few inline comments noticed on first read.

spec_26.rst
grondo: Should have noted I'm willing to change the implementation. A guaranteed string here will really help in error reporting, the WIP implementation of jobid dependencies has to coerce non-string values to a string for error messages

garlick:
I'll go ahead and force push the easy changes.  The questions about value type and URI optional key parsing seem like they might require more thought/discussion so first things first.

spec_26.rst
garlick: Is this description correct WRT the exception?

> The dependency SHALL be satisfied once the antecedent job enters RUN state.
If the antecedent job reaches INACTIVE state without entering RUN state,
a fatal exception SHOULD be raised on the dependent job.

spec_26.rst
garlick: Got it, will change to describe that behavior.

garlick:
I went ahead and  force pushed a change describing the URI options.

spec_26.rst
garlick: Based on the fact that we've had difficulties in the past with interpreting JSON on the command line (is it a string or a JSON string?)  I probably lean towards string here, however if you want to go the other way I won't put up a fight :-)

spec_26.rst
grondo: Yes, that is correct.

spec_26.rst
garlick: Oh well that was kind of a dumb comment since we're using the URI format on the command line.  OK, I have no opinion whatsoever.  You decide.

grondo: I say we keep this as a string and let the implementation parse any non-string values.

Assuming a string here will let the plugins detect non-conformant value in `unpack()`, and will ensure errors reported back to the user contain the user's actual input, and not some json converted value (e.g. I put in `1.25` and the error says `1.250000000000: invalid value`)

garlick: Makes sense to me.

grondo:
Should it also be mentioned here that dependencies as managed by `dependency-add` and `dependency-remove` events are not "reference counted" (not sure the correct term)? That is, multiple dependency-add events with the same description are ignored, and it is an error to add a duplicate named dependency that was previously removed.

garlick:
Just writing something up for rfc 26 - will post here  as a cinnebt and we can continue discussing...

grondo (approved):
This looks good to me as is. We could continue discussing the finer points here, though I think this could be merged as is with revisions posted as separate PRs if necessary.

garlick:
I'm fine if we want to take this to another pr, but here's a little sub-section I was going to propose to add to the overview section.  Not sure I entirely captured what you were mentioning?

```rst
Dependency Event Semantics
~~~~~~~~~~~~~~~~~~~~~~~~~~

Dependency events SHALL only be posted to the job eventlog by job manager
plugins.

Dependency events SHALL be treated as matching when their ``description``
fields have the same value.

A dependency SHALL be considered satisfied when matching ``dependency-add``
and ``dependency-remove`` events have been posted.

Some special semantics for these events are needed to allow plugins
to reacquire their internal state when the job manager is restarted:

Attempts to post duplicate ``dependency-add`` events for unsatisfied
dependencies SHALL NOT raise a plugin error and SHALL NOT be posted.

Attempts to post duplicate ``dependency-add`` events for satisfied
dependencies SHALL raise a plugin error.
```

grondo:
That sounds good to me!

garlick:
Want to give it another look or shall I set MWP?

grondo:
Go ahead and set MWP :+1:

SteVwonder:
Thanks @garlick!

garlick:
> Thanks @garlick!

No problem!  If we merged that before you had a chance to review and comment, please do bring up any issues.  We can always do another quick pr.

PR #295, opened 2021-07-02
rfc3: require route IDs to be 16 bit UUID

garlick:
Per discussion with @chu11, tighten up the definition of a message route ID from "one or more octets" to a 16 octet UUID.  This might simplify the message route handling code.

It should be noted that currently all message route IDs are UUIDs, but they are encoded as 37 byte strings.  This change would require the IDs to not be encoded.

Don't merge yet - we need to see how @chu11 gets on with the message code.


chu11:
gonna post some comments in https://github.com/flux-framework/flux-core/issues/3617

TL;DR - I think the change would be a lot of work and unclear if there's a real gain.

garlick:
Withdrawing this for now.

PR #297, opened 2021-09-15
rfc26: after:JOBID jobs should start after start event

grondo:
Problem: RFC 26 specifies that after:JOBID jobs start after the
antecedent enters the RUN state. However, it is more practical and useful
if dependent jobs are not started until the dependency also posts a
'start' event.

Update the language to indicate that dependent jobs with the 'after'
scheme are not started until the antecedent enters the RUN state *and*
a "start" event is posted.


SteVwonder (approved):
LGTM!

grondo:
Thanks!

PR #302, opened 2021-10-30
rfc25: add moldable specification into Jobspec V1

dongahn:
Copy the moldable specification from our canonical
job spec (RFC14) to job spec version 1 (RFC25).

Need to support a moldable jobspec including
resources.count.min both within flux-core and
fluxion in order to enable node-exclusive scheduling
(e.g., typical system instance scheduling policy).

### DIFF
--- a/spec_25.rst
+++ b/spec_25.rst
@@ -104,9 +104,23 @@ following:
    below.
 
 **count**
-   The ``count`` key SHALL indicate the desired number of
-   resources matching the current vertex. The ``count`` SHALL be a single integer
-   value representing a fixed count
+   The ``count`` key SHALL indicate the desired number or range of
+   resources matching the current vertex. The ``count`` SHALL have one
+   of two possible values: either a single integer value representing
+   a fixed count, or a dictionary which SHALL contain the following key:
+
+   **min**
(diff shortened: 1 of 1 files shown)


dongahn:
This should not be merged until Issue https://github.com/flux-framework/flux-core/issues/3928 is resolved.

grondo:
I wonder if we should make things easier on ourselves (at least for flux-core) only allow `count` or a `min` keyword for V1 jobspec, instead of also allowing the optional `max`, `operator` and `operand` keys?

I don't feel too strongly about this, but allowing those keys will be slightly more work in flux-core I assume, when we really only require the `min` keyword to support first-cut modlable jobs.

dongahn:
> I don't feel too strongly about this, but allowing those keys will be slightly more work in flux-core I assume, when we really only require the min keyword to support first-cut modlable jobs.

Yes, I support this idea since this will be minimum viable and I will revise the PR. @grondo: we still don't want to change the version number for our Jobspec (V1) for this minor change, correct?

dongahn:
OK. I made the suggested change and forced a push.

dongahn:
Closing this PR per our discussion at today's meeting.

PR #303, opened 2021-11-12
[WIP] rfc21: document job reattach events

chu11:
This documents what I did in https://github.com/flux-framework/flux-core/pull/3947

I don't know if this should really be considered a mergeable candidate, but mostly wanted to just document what is in that PR and/or potentially lead to discussion.  The terminology "reattach" is of course debatable.

### DIFF
--- a/spec_21.rst
+++ b/spec_21.rst
@@ -391,6 +391,37 @@ status
    {"timestamp":1552594348.0,"name":"epilog-finish","context":{"description":"/usr/sbin/job-epilog.sh", "status":0}}
 
 
+Flux-Reattach Event
+^^^^^^^^^^^^^^^^^^^
+
+The job manager has restarted and will attempt to reattach to a running job.
+
+The following keys are OPTIONAL in the event context object:
+
+id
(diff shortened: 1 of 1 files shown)


garlick:
What do the events signify, and do they need to be in the main eventlog or should the shells log disconnect/reconnect events in their eventlog?

chu11:
> What do the events signify, and do they need to be in the main eventlog or should the shells log disconnect/reconnect events in their eventlog?

At the moment they are informational.  So perhaps they don't even need to be documented?  Or don't even need to be in the eventlog?  The only reason I put them in the eventlog is that we want to know when a job has been "reattached", because there info that could be lost.

For example, we normally calculate runtime by calculating `t_cleanup` - `t_run` timestamps.  But in a "reattach" case, this runtime calculation may not be accurate b/c a job may have finished before we reattached and output the `cleanup` state to the eventlog.

chu11:
closing, for the purposes of https://github.com/flux-framework/flux-core/pull/3947 decided to prefix eventlog events with `debug` as we design the longer term solution

PR #304, opened 2021-11-24
rfc21: add job memo event

grondo:
This PR introduces a `memo` event for posting user data to the job eventlog as an augmentation of the current annotations system. The intent is to have a way to associate data with a job and have it persist across job-manager restart.

The specific use case is a mechanism to allow a child job of a Flux instance to publish its remote-uri in a place with easier access than the job's kvs, without losing the uri if the instance is restarted.

### DIFF
--- a/spec_21.rst
+++ b/spec_21.rst
@@ -520,6 +520,14 @@ start
 free
    A problem occurred while releasing resources to the scheduler.
 
+Memo Event
+^^^^^^^^^^
+
+A brief data record has been associated with the job.
+
+The context object SHALL be considered the data to associate with
+the job.  Any data attached to the job via a previous "memo"
+event SHALL be updated, i.e. existing keys SHALL be replaced.
 
 Debug Event
 ^^^^^^^^^^^


spec_21.rst
```diff
@@ -520,6 +520,14 @@ start
...
    A problem occurred while releasing resources to the scheduler.
 
+Memo Event
+^^^^^^^^^^
+
+A brief data record has been associated with the job.
+
+The context object SHALL be considered the data to associate with
+the job.  Any data attached to the job via a previous "memo"
+event SHALL be updated, i.e. existing keys SHALL be replaced.
```
chu11, line 530: perhaps should put an example like other events in the doc?

garlick: Should the context object contain a key-value pair?

grondo: > perhaps should put an example like other events in the doc?

Good point! Thanks!

> Should the context object contain a key-value pair?

Since it is a JSON object, that is by definition a "a collection of name–value pairs where the names are strings." We could make that explicit here, though.

grondo:
I'm going to close this PR and reopen a new one due to a typo in the branch name :facepalm: which could be confusing.

spec_21.rst
garlick: The first sentence by itself seemed like it could be read as one memo per job, with the context object as the value.

grondo: In essence that is what it is -- one memo object per job, updated by each `memo` event (existing keys are replaced, new keys updated). However, I'm fine with changing the description if you have a suggestion.

grondo: How about:

> The context object SHALL be a set of key-value pairs to associate with the job. Existing identical keys from a previous `memo` event SHALL be replaced.

PR #305, opened 2021-11-24
rfc21: add memo event

grondo:
Problem: No defined job event allows user data to be associated with
a job and captured in the eventlog for replay.

Add a "memo" event which associates generic data with a job.

New version of #304 with corrected branch name.

### DIFF
--- a/spec_21.rst
+++ b/spec_21.rst
@@ -520,6 +520,20 @@ start
 free
    A problem occurred while releasing resources to the scheduler.
 
+Memo Event
+^^^^^^^^^^
+
+A brief data record has been associated with the job.
+
+The context object SHALL contain a set of key-value pairs to associate
+with the job. Existing identical keys from a previous memo event SHALL
+be replaced. A value of JSON ``null`` SHALL remove the associated key.
+
+Example:
+
+.. code:: json
+
+  {"timestamp":1637723184.3725791,"name":"memo","context":{"key":"value"})
 
 Debug Event
 ^^^^^^^^^^^


grondo:
I've pushed an example as suggested by @chu11, still an open comment that needs to be resolved by @garlick:

> Should the context object contain a key-value pair?

Maybe we could add "The context object SHALL contain at least one key-value pair"?

grondo:
Ok, I've force pushed an update to the description of the `memo` event:

> The context object SHALL contain a set of key-value pairs to associate
> with the job. Existing identical keys from a previous memo event SHALL
> be replaced.

Any better @garlick?

grondo:
Pushed an update to indicate that a value of JSON `null` indicates the associated key is to be removed. (In order to be compatible with current annotations behavior)

chu11:
Just curious, what is the eventual use case?  Like possibly a `flux job memo get <jobid>` will get the current memo for the job so the user can use it?

garlick (approved):
LGTM!

@chu11, if I understand @grondo correctly, this solves an immediate problem that could not be resolved with the current preservation semantics for annotations, but is groundwork for a possible future replacement for annotations, the design of which got a little wonky when we decided to commit to the eventlog in the alloc event context.  More discussion required on that I think.

grondo:
> Just curious, what is the eventual use case? 

The specific use case is allowing a child job to register its URI with the parent in a form that can be easily obtained with `flux jobs`. Annotations would work, but they are not preserved across a restart, so we need something that can be recorded in the eventlog.

The current implementation just uses the `memo` to update `annotations.user` (though I'm wonder if that's a bad idea. Maybe "user annotations" should be dropped now and we could extend `job-list` and `JobInfo` to know about a `{memo}` attribute, so the URI would be available as `{memo.uri}` instead of `{user.uri}`. I'm on the fence though because I was looking for a least effort solution :slightly_smiling_face: )

PR #306, opened 2021-12-02
rfc21: add reattach job state

chu11:
a first cut at the addition of a "reattach" state to the job state.  "reattach" word choice can be debated of course.

My knowledge of dot is not immense, but I appear to have got something that looks pretty :P.


garlick:
Quick comment:  maybe the state should be DETACHED and the events should be `detach` and `attach`?

My only misgiving about that name is it might be confused with `flux job attach`.

What about OFFLINE with events `disconnect` and `reconnect`?  The term "offline mode" is used for mail user agents that can continue to function while disconnected from their IMAP client.  I guess I always had that comparison in the back of my mind.

chu11:
re-pushed, i went with the suggested "offline" nomenclature above.

PR #307, opened 2021-12-22
rfc2: relax licensing

garlick:
Per offline discussion about KubeFlux licensing needs, this relax the LGPL-3.0 REQUIRED language to RECOMMENDED for Flux framework projects.

Also, brought over a fix for rendering bullet lists on readthedocs from flux-docs, as I noticed they weren't rendering right here.


grondo (approved):
Thanks!

PR #310, opened 2022-01-30
mergify: replace strict merge with queue+rebase

grondo:
Problem: mergify.io has deprecated the strict merge mode action:

https://blog.mergify.io/strict-mode-deprecation/

Replace the strict mode configuration with a default queue with a
rebase action which should be equivalent.


grondo:
As usual, this will need to be manually merged.

garlick (approved):
Thanks!

PR #312, opened 2022-02-07
rename keepalives to control messages

garlick:
Problem: it's going to be confusing having both application level "keepalives" (not even really named properly for how they are now used) and TCP keepalives in configuration, etc..

Rename keepalives to "control messages".

Update module rfc to describe regular RPCs instead of keepalive use for this purpose, as proposed in flux-framework/flux-core#4110.

If accepted, another flux-core PR would switch the broker over to using "control messages".

### DIFF
--- a/spec_3.rst
+++ b/spec_3.rst
@@ -30,7 +30,7 @@ Goals
 -----
 
 The Flux message protocol v1 provides a way for Flux utilities and services to
-communicate with one another within the context of a job. It has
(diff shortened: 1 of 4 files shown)


grondo (approved):
Somehow missed this PR, sorry!

LGTM

garlick:
I had thought it was already in!  I think we already changed flux-core. 

Thanks!  Setting MWP.

PR #313, opened 2022-02-25
rfc20: add optional execution.properties key to Rv1

grondo:
Problem: Rv1 does not support a method for assigning properties to
resources, but simple property support is a near term requirement for
Flux.

Add an optional `properties` key to the Rv1 `execution` dictionary which
allows for the simple assignment of named properties to sets of ranks.

### DIFF
--- a/spec_20.rst
+++ b/spec_20.rst
@@ -128,8 +128,9 @@ Execution
 ~~~~~~~~~
 
 The value of the ``execution`` key SHALL contain at least the keys
-``R_lite``, and ``nodelist``, with optional keys ``starttime`` and
-``expiration``. Other keys are reserved for future extensions.
+``R_lite``, and ``nodelist``, with optional keys ``properties``,
+``starttime`` and ``expiration``. Other keys are reserved for future
+extensions.
 
 ``R_lite`` is a strict list of dictionaries each of which SHALL contain
 at least the following two keys:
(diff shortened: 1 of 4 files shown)


dongahn:
This looks good to me. Since other people may ask the same question that I asked on the other issue ticket, we can say to the effect that each execution target can be tagged with  an arbitrary set of properties using this method? (i.e., overlapping rank sets can be used to specify different properties)?

grondo:
Thanks @dongahn, I added "A rank SHALL be assigned all properties for which its ID appears in the designated idset"

grondo:
I thought it might be good to reserve some characters that can not be used in property names, since in the future we may want to support boolean queries on the command line like `--requires='(foo || bar) && baz'`. Unsure if I used the correct RFC syntax for that, but here's a starting point anyway.

spec_20.rst
```diff
@@ -162,6 +163,18 @@ contain a string in RFC 29 *Hostlist Format*, e.g. ``host[0-16]``.
...
+   names, the values of which SHALL be a set of ranks in RFC 22 **idset
+   format** to which to assign the named property. A rank SHALL be assigned
+   all properties for which its ID appears in the designated idset.
+   Property names SHALL be valid UTF-8, and MUST NOT contain the
+   following reserved characters:
+
+   ::
+
+      ! & ' " ^ ` | ( )
+
```
garlick, line 177: This is fine, but I find myself rereading it multiple times to "get it".  Not sure if this is any better.  It just breaks things up a bit so there is less english to parse (IMHO). :shrug: 

**properties**
The optional `properties` key SHALL be a dictionary where each key maps a single property name to a RFC 22 idset string.  The idset string SHALL represent a set of execution target ranks.  A given execution target rank MAY  appear in multiple property mappings.  Property names SHALL be valid UTF-8, and MUST NOT contain the
   following reserved characters:

garlick (approved):
This seems fine to me.  It sounds like @dongahn is on board?

I tried rewording the paragraph for clarity but I'm not sure I succeeded.  Your call.  It's fine as is IMHO.

spec_20.rst
dongahn: @grondo: Can we add `@` as a reserved character as well? Then, we state to the effect that the suffix string that appears after `@` SHALL be a scheduler-specific string. Given `amd-mi50@gpu`, `amd-mi50` SHALL be the property string, but the scheduler MAY use the `gpu` suffix to perform a scheduling optimization for the gpu resources of the corresponding ranks?

grondo: @garlick: Thanks, your rewrite sounds better IMO.

@dongahn: I will add something to that effect for the `@` character, though I won't lump it with the other reserved characters that are not allowed to appear in a property name.

dongahn: Yes, the description def. should be its own thing. Are you thinking `@` would still be listed as one of the reserved characters, though?

grondo: I changed the "reserved" characters to "illegal" characters since that seems more apt, then separately "reserved" the `@` character for scheduler specific use. Happy to iterate until we're all happy with the final wording.

dongahn: Works for me.

dongahn (approved):

grondo:
Thanks!

PR #314, opened 2022-03-01
rfc31: add format for the specification of job constraints

### LINKED ISSUE #243: RCF14: add property support into V1 jobspec.
I just created https://github.com/flux-framework/flux-sched/issues/652 on the need for our graph scheduler to do resource property resource selection. The other side of this problem is how to specify one or more properties on certain resources in a jobspec. Is there already a way to do this with flux mini interface?

### LINKED ISSUE #246: RFC25: Need bare minimum support for properties spec
As part of LLNL's short-term Corona use case, we found that we will need properties spec support within our jobspec. After we converge on the canonical spec (https://github.com/flux-framework/rfc/issues/243), we should reduce that down to a bare minimum for RFC25.

grondo:
This is a first cut proposal for encoding generic job constraints in a JSON object for inclusion in jobspec.

The proposed format is general enough to allow fairly complex matching (in case this is ever required), but in its simplest form, looks identical to the initial proposal in flux-framework/flux-core#4143, i.e. `attributes.system.constraints.properties=[ array of required properties ]`.

The format is largely inspired by [JsonLogic](https://jsonlogic.com).

I decided it would be cleaner to propose a separate RFC for job constraints rather than adding separately to each of the jobspec RFCs. We can then refer to this RFC when describing where constraints go in the jobspec RFCs, as we have done with dependencies.

Fixes #243
Fixes #246


garlick (approved):
> decided it would be cleaner to propose a separate RFC for job constraints rather than adding separately to each of the jobspec RFCs. We can then refer to this RFC when describing where constraints go in the jobspec RFCs, as we have done with dependencies.

I think that was the right move.  This looks good as is!

grondo:
@garlick noticed the commit here was accidentally included in the merged PR #313.

I'll rework this PR to accumulate any amendments (somehow), or maybe close it if the end result was acceptable.

Sorry about that!

grondo:
Ok, what I've done is revert the accidental commit and then committed it again immediately on top so that it can properly be merged in this PR.

If the original was acceptable we could optionally close this PR and make changes in future PRs.

dongahn:
> attributes.system.constraints.properties

We have used attributes.system.scheduler.queue to capture queue name constraints. I have a slight preference to use `scheduler` key instead of `constraints`. That way, all of the scheduler's constraints can be found under one parent key. 

Having said that, do you see a user case of properties beyond scheduler constraints? In case you have a specific reason to use `constraints` as your parent key choice.

grondo:
Here is my main argument against `attributes.system.scheduler.constraints.properties`:

Since the main purpose of jobspec is to express program requirements to the scheduler, I find a separate `scheduler` key to be a slight redundancy -- though for scheduler specific, non-RFC 14 data, it does make sense to adopt a `scheduler` key where the contents are _opaque_ to flux-core.

Since `constraints` will be standardized via an RFC, it thus makes sense to promote the key to the top level of jobspec system attributes. This will make it easier (i.e. less verbose) for flux-core utilities to generate and query jobspec constraints, and easier for users crafting jobspec with constraints.

dongahn:
@grondo: that's a good argument. Keeping generic properties at a higher level and scheduler-specific parameters under `scheduler` key. I am sold on that then.

grondo:
closing this one since RFC 31 was merged in a separate PR

PR #315, opened 2022-03-02
rfc14: update definition of duration=0

grondo:
Problem: RFC 14 defines a jobspec duration of 0 as being unset or
unlimited, but does not specify further what that means. This could
allow different scheduler implementations to drift in how they apply
time limits when no duration is specified.

Add wording that explicitly indicates how a scheduler should treat an
unset duration.

### DIFF
--- a/spec_14.rst
+++ b/spec_14.rst
@@ -316,11 +316,16 @@ which, if present, must have values. Values MAY have any valid YAML type.
 Some common system attributes are:
 
 **duration**
-   The value of the ``duration`` attribute is a floating-point number greater than
-   or equal to zero representing time span in seconds. A ``duration`` of 0 SHALL be
-   interpreted as "unlimited" or "not set" by the system. The scheduler will make
-   an effort to allocate the requested resources for the number of seconds
-   specified in ``duration``.
+   The value of the ``duration`` attribute is a floating-point number
+   greater than or equal to zero representing time span in seconds.
+   If ``duration`` is greater than zero, then the scheduler SHALL
+   allocate the requested resources for the number of seconds specified
+   in ``duration``. A ``duration`` value of ``0.`` SHALL be interpreted
+   as unlimited or unset, and the scheduler SHALL use a ``duration``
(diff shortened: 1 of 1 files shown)


spec_14.rst
```diff
@@ -316,11 +316,15 @@ which, if present, must have values. Values MAY have any valid YAML type.
...
 
 **duration**
-   The value of the ``duration`` attribute is a floating-point number greater than
-   or equal to zero representing time span in seconds. A ``duration`` of 0 SHALL be
-   interpreted as "unlimited" or "not set" by the system. The scheduler will make
-   an effort to allocate the requested resources for the number of seconds
-   specified in ``duration``.
+   The value of the ``duration`` attribute is a floating-point number greater
+   than or equal to zero representing time span in seconds.  The scheduler
+   will make an effort to allocate the requested resources for the number
```
dongahn, line 321: I know this is not your change. But is there any case where the scheduler will not allocate for duration? If not, perhaps this can be a stronger requirement? 

```suggestion
SHALL allocate the requested resources for the number
```
This seems more aligned with the spirit of this change which makes the scheduler's behavior more explicit -- as a  requirement. 

I understand this may try to cover a case where the scheduler may not be able to allocate the duration at all. IMHO, we can assume that the scheduler can allocate in this section?

grondo: Yes, that seems reasonable. I will make that change.

grondo: Ok, I pushed the requested change, and also added the qualifier `If the duration is greater than zero` before that statement.

dongahn (commented):
LGTM! Thanks @grondo

spec_14.rst
dongahn: Yup better!

grondo:
Great, I'll set MWP once you officially set approval.

dongahn (approved):

dongahn:
Done

grondo:
Thanks!

PR #318, opened 2022-03-23
rfc14: add constraints attribute to jobspec

grondo:
Problem: RFC 31 defines a format for specification of generic job
constraints, but RFC 14 does not define the location for a constraints
in jobspec.

Define attributes.system.constraints as the location in jobspec for
RFC 31 defined constraints.


grondo:
Thanks! I forgot to update the jsonschema though (and we even just talked about that :facepalm:)

garlick (approved):
Looks good.

PR #320, opened 2022-03-24
rfc20: expand description of scheduler specific properties

grondo:
Problem: The purpose of the scheduler-specific `@` character in
property strings is vague in RFC 20, and could cause confusion.

Expand the description of the purpose and possible use case for `@`
in execution target properties. Specify that the literal property
string shall still apply to the execution target, but the scheduler
implementation may use all or part of the prefix and suffix before the
`@` for scheduler specific optimizations.


dongahn (approved):
LGTM. Since the use case for @suffix is for the future, the @suffix use case is *rightly* vague IMHO. Thanks. Feel free to put MWP.

grondo:
Thanks @dongahn!

PR #321, opened 2022-03-24
rfc14,25: add example and references for constraints, dependencies

### LINKED ISSUE #319: constraints key in RFC14 and RFC24 need some more work
Per https://github.com/flux-framework/flux-core/pull/4236#issuecomment-1077049826, RFC14 should list RFC34 in the Related Standard. Also as a consideration, mentioning "properties" as an example constraint can make RFC14 a bit more self contained.

In addition RFC24 need to specify "constraints" key.

grondo:
As described in #319, the lack of an example and references for the `constraints` key makes searching the RFC for this support challenging. Also references for `constraints` and `dependencies` is missing from RFC 25 (Jobspec v1 definition).

This PR addresses these concerns.

Fixes #319

### DIFF
--- a/data/spec_14/use_case_2.9.yaml
+++ b/data/spec_14/use_case_2.9.yaml
@@ -1,0 +1,20 @@
+version: 999
+resources:
+  - type: slot
+    count: 1
+    label: default
+    with:
+      - type: node
+        count: 1
+tasks:
+  - command: [ "flux", "start" ]
+    slot: default
+    count:
+      per_slot: 1
+attributes:
+  system:
+    duration: 3600.
+    cwd: "/home/flux"
+    constraints:
+      properties:
+        - amd-mi50
--- a/spec_14.rst
+++ b/spec_14.rst
@@ -37,6 +37,8 @@ Related Standards
 
 -  :doc:`26/Job Dependency Specification <spec_26>`
 
+-  :doc:`31/Job Constraints Specification <spec_31>`
+
 
 Goals
 -----
@@ -653,6 +655,17 @@ Specific Example
 Jobspec YAML
    .. literalinclude:: data/spec_14/use_case_2.8.yaml
       :language: yaml
+
+Use Case 2.9
+   Specify constraints based on properties
+
+Specific Example
+   Require that allocated resources have the ``amd-mi50`` property
+
+Jobspec YAML
+   .. literalinclude:: data/spec_14/use_case_2.9.yaml
+      :language: yaml
+
 
 .. [#f1] `YAML Ain’t Markup Language (YAML) Version 1.1 <http://yaml.org/spec/1.1/current.html>`__, O. Ben-Kiki, C. Evans, B. Ingerson, 2004.
 
--- a/spec_25.rst
+++ b/spec_25.rst
@@ -38,6 +38,9 @@ Related Standards
 
 -  :doc:`20/Resource Set Specification Version 1 <spec_20>`
 
+-  :doc:`26/Job Dependency Specification <spec_26>`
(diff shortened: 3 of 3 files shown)


data/spec_14/use_case_2.9.yaml
```diff
@@ -0,0 +1,23 @@
...
+  - command: [ "flux", "start" ]
+    slot: default
+    count:
+      per_slot: 1
+attributes:
+  system:
+    duration: 3600.
+    cwd: "/home/flux"
+    constraints:
+      or:
```
dongahn, line 19: This operator is future extension, correct? Perhaps we should just use a simple use case we support today? My burden is to follow canonical jobspec in fluxion and at this point I do hope to minimize that effort.

spec_14.rst
```diff
@@ -654,6 +656,18 @@ Jobspec YAML
    .. literalinclude:: data/spec_14/use_case_2.8.yaml
       :language: yaml
 
+Use Case 2.9
+   Specify constraints based on properties
+
+Specific Example
+   Require that allocated resources have the ``amd-mi50`` or ``nvidia-t100``
```
dongahn, line 663: `nvidia-v100` or `nvidia-a100` is a bit better?

dongahn (commented):
I left a couple of minor comments

data/spec_14/use_case_2.9.yaml
grondo: I can remove that, but support for it was so easy it is supported in sched-simple. (I was going to promote the `and` `or` and `not` operators to required in RFC 31, but I can hold off if you want. It is just that I figured someone may want to support `or` in the near term..)

spec_14.rst
grondo: Yeah, I'll choose one of those (it was a typo and meant to be `t1000` anyway). I wasn't really sure what would be a likely nvidia device name anyway.

data/spec_14/use_case_2.9.yaml
dongahn: That'd be appreciated. I am working on a PR for fluxion side and I'd like to do a minimum viable for that one. (Not even including fine-grained property attachment which I was planning -- based on our coordination.) 

For the next PR on this topic can have all of the remaining items. 

Having this in the code in sched-simple and jobspec library in flux-core is just fine. 

It is just that, once these appear in the RFCs, we will all feel obligated to support them in our codes. But IMHO my time should go to a more high priority item than perfecting this support.

grondo: Ok, I will leave RFC 31 as is and I will simplify the example to be conforming to the _minimum_ of the spec. :+1:

grondo: Done.

spec_14.rst
dongahn: :-). 

Yeah I love those consumer GPUs -- I have a 1080ti that my friend gave me and it took 5 mins to train a very small transfer learned model. 

My ubuntu  box was initially misconfigured so tensorflow was accidentally using the AMD EPYC cores and this took over 1 hour to train!! 

v100, a100 and h100 (newly announced) are server classes with more vendor verification, reliability... at the cost of much much high prices. Feels like preaching to the choir though

grondo:
Ok, @dongahn, I think I've addressed all comments (and finally got rtd build working).

garlick (approved):
This looks fine, and reviewing the discussion, I don't think there's any reason to hold off merging this?

grondo:
Correct. I'll set mwp

PR #323, opened 2022-04-05
rfc28: add property-add, property-remove keys

grondo:
This PR is a first stab at an extension to the resource acquire protocol which would allow dynamic addition and removal of execution target properties.

The end result of such an extension might be a `flux resource property add/remove` command, which could end up being useful in even single-user instances of Flux (e.g. if a workflow wants to partition resources in some way, detects slow vs fast nodes, mark nodes on which a job ran for data locality, etc.)

This is just a proposal at this point. Feel free to discuss the implementation here.

One drawback to adding new keys to the `resource.acquire` response stream payload is that a scheduler could just ignore these new keys and `flux resource property add/remove` would silently do nothing. Perhaps the scheduler should additionally have some way to announce supported capabilities when it issues the `resource.acquire` RPC? Or, maybe since we only really have 2 schedulers, this is not a practical concern at this point..

### DIFF
--- a/spec_28.rst
+++ b/spec_28.rst
@@ -150,6 +150,20 @@ down
   for scheduling.  The idset only contains targets that are transitioning,
   not the full set of unavailable targets.
 
+property-add
+  (object) RFC 20 conforming properties object containing properties that
+  should be added to the specified execution targets. When present, this
+  key reflects an update to the instance resource inventory which MAY
+  affect job satisfiability, the determination of which is left to the
+  scheduler implementation.
+
+property-remove
+  (object) RFC 20 conforming properties object containing properties that
+  should be removed from the specified execution targets. When present,
+  this key reflects an update to the instance resource inventory which
+  MAY affect job satisfiability, the determination of which is left to the
+  scheduler implementation.
+
 
 Example:
 
@@ -158,6 +172,8 @@ Example:
    {
       "up": "3-6",
       "down": "2"
+      "property-add": { "foo": "0-1" },
+      "property-remove" { "bar": "3" }
    }
 
 If down resources are assigned to a job, the scheduler SHALL NOT raise an


spec_28.rst
```diff
@@ -150,6 +150,14 @@ down
   for scheduling.  The idset only contains targets that are transitioning,
   not the full set of unavailable targets.
 
+property-add
+  (object) RFC 20 conforming properties object containing properties that
+  should be added to the specified execution targets.
```
garlick, line 155: Right now we reject jobs as unsatisfiable if they require a property that is unknown, or if there are not enough nodes with the property to fulfill the resource request.  That seems right to me.

Do we want that to continue to hold if properties can be defined or increase counts dyanmically?  (I would argue yes).  Does that need to be stated here somehow?

spec_28.rst
```diff
@@ -150,6 +150,14 @@ down
...
+
+property-remove
+  (object) RFC 20 conforming properties object containing properties that
+  should be removed from the specified execution targets.
+
```
garlick, line 160: What happens to pending jobs if a property they require is removed or its counts decrease so that the job is no longer satisfiable?

Does it require the job manager to re-run the satisfiability checks when there is a property change?

garlick (commented):
Couple of questions inline.

Also, does resource.R and the resource inventory change as property assignments change?

spec_28.rst
grondo: Good point. I agree also that feasibility check would only apply to the current set of resource properties. That is, all this proposal adds is a way to amend the scheduler's R without having to rewrite `resource.R` and reload modules.

spec_28.rst
grondo: Good question:

The job manager does not re-run any satisfiability checks, though now it occurs to me you possibly meant the scheduler here? Since feasibility checks are only made at ingest time, whether or not the scheduler runs a feasibility check on all known jobs is implementation dependent and outside the scope of this RFC. Worst case, the scheduler could raise a job exception on a job that is not feasible when it is next considered for scheduling (as is done now when the feasibility job-validator plugin is not active)

grondo:
> Also, does resource.R and the resource inventory change as property assignments change?

That was my assumption, though `resource.R` and the internal implementation of the flux-core `resource` module seemed to be outside the scope of this RFC, which seems to detail the resource module and scheduler protocol only? (I could be wrong here because I did not do a through re-reading of the entire RFC)

spec_28.rst
garlick: Good point, sorry I forgot that it's `job-ingest` making an RPC to `sched` to determine feasiblilty at ingest time.

It seems reasonable to make it scheduler dependent.

grondo:
In terms of the rest of the implementation details, I was thinking something simple would be to have `resource` expose a new instance-only service that allowed `property-add/remove` or similar keys that would be translated into events in the resource eventlog, so that they properties could be re-established on a restart. Though perhaps allowing dynamic properties in a system instance is a bad idea, and we should have some way to disable the feature?

I agree this change is probably low priority given other work that needs to be done, so I'd be fine tabling this for now.

garlick:
This all sounds reasonable and I am fine moving forward with this.

We might want to add something about the fact that these operations reflect changes to R, which impacts job request satisfiability, the determination of which is left to the scheduler implementation.  They should not be thought of as comparable to offline/online which don't affect satisfiability.

When we support grow and shrink for execution targets hopefully we can handle them the same way.

grondo:
Ok, I basically plagiarized your wording above and added to the document:
> When present, this
> key reflects an update to the instance resource inventory which MAY
>  affect job satisfiability, the determination of which is left to the
>  scheduler implementation.

garlick (approved):
Sounds good!

garlick:
did you want to set MWP on this @grondo?  Seems like it's ready for merging?

grondo:
Yeah, I suppose we might as well so we can reference these keys if the code ever gets implemented...

PR #325, opened 2022-04-14
rfc23: add "ms" suffix to Flux Standard Duration

grondo:
Problem: The Flux Standard Duration specification only allows a unit
suffix of "s" (seconds) or greater, but there are times when it would
convenient to represent a duration in milliseconds (ms), especially
when converting a small duration into FSD.

Add an "ms" suffix to the FSD spec. Reword the specification to allow
for string suffixes instead of only a single character suffix.

### DIFF
--- a/spec_23.rst
+++ b/spec_23.rst
@@ -39,11 +39,14 @@ said to support "Flux Standard Duration."
 Implementation
 --------------
 
-A duration in Flux Standard Duration SHALL be of the form ``N[SUFFIX]`` where
-``SUFFIX`` SHALL be optional and, if provided, MUST be a single character from the
-set ``s,m,h,d``. The value ``N`` MUST be a non-negative, non-infinite,
-floating-point number excluding ``NaN``. The value ``N`` SHALL be in one of the
-forms allowed by C99  [#f1]_ ``strtof`` or ``strtod`` and SHALL be interpreted as:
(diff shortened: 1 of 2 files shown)


spec_23.rst
```diff
@@ -39,14 +39,17 @@ said to support "Flux Standard Duration."
...
+where ``SUFFIX`` SHALL be optional and, if provided, MUST be a string from
+the set { ``ms``, ``s``, ``m``, ``h``, ``d`` }. The value ``N`` MUST be a
+non-negative, non-infinite, floating-point number excluding ``NaN``. The
+value ``N`` SHALL be in one of the forms allowed by C99  [#f1]_ ``strtof``
+or ``strtod`` and SHALL be interpreted as:
 
 -  *seconds* if ``SUFFIX`` is not provided, or is ``s``.
 
+-  *milliseconds* if ``SUFFIX`` is ``ms``.
+
```
chu11, line 52: shouldn't milliseconds go before seconds so its in order from smallest to biggest?

grondo: No, because seconds in the default if there is no SUFFIX. At least it worked better in my mind that way.

grondo: Hm, I guess re-reading it, it would be fine if milliseconds was first if that is preferable.

grondo: Ok, fixed. That is better now, I agree!

chu11 (approved):

PR #327, opened 2022-04-19
add new RFC for job execution protocol

garlick:
This documents the current protocol used between the job manager and the job execution system, as a precursor to reworking the job execution service.  @grondo already had RFC 32 in progress so I skipped to RFC 33.

PR #328, opened 2022-04-20
rfc32: minor fixes

chu11:
Per discussion in #327, fix some minor gotchas that were missed.


garlick (approved):
:cat: says "purrfect"

PR #329, opened 2022-04-20
WIP: experimental: add new RFC for job execution module protocol

grondo:
This PR is a WIP definition of a distributed protocol for the Flux job execution module, and defines an _initial_ design for distributing the job execution module across a Flux instance.

A prototype will be developed using this RFC as a starting point, with the idea that any or all of the specification detailed here may change along the way. However, having something proposed in writing will help capture the design constraints, along with discussion and changes along the way.


spec_33.rst
```diff
@@ -0,0 +1,330 @@
...
+Job execution modules SHALL be loaded on all ranks in an instance, and are
+organized in a hierarchy with rank 0 at the root. Each module SHALL track
+the set of all running jobs for itself and all of its children. This state
+SHALL include at a minimum the jobid, userid, job state, and the idset of
+execution targets on which the job has an allocation.
+
+All job execution modules register a ``job-exec.hello`` service endpoint.
+Downstream execution modules send a ``hello`` request to their upstream
+peer to initiate the execution module protocol. An execution module SHALL
+wait to send a ``hello`` response to its downstream peers until n an initial
```
chu11, line 92: typo here `n an`

spec_33.rst
```diff
@@ -0,0 +1,330 @@
...
+the following:
+
+state-update
+ A ``state-update`` response is used to update the distributed state of
+ jobs. The ``data`` object SHALL have a single key, ``jobs``, which SHALL
+ be an array of (id, userid, type, idset) tuples. The ``type`` entry of the
+ tuple SHALL indicate how the state is to be resolved on ranks in ``idset``.
+ Possible values for ``type`` MAY be *add*, *remove*, or *check*, to *add*
+ a new job, *remove* an inactive job, or *check* that an existing job is
+ active as expected.
```
chu11, line 112: can replace the last sentence with "described below" (or equivalent)

chu11 (commented):
awesome to see how this is coming together.  just saw two nits when reading through it.  no biggie, I know this is WIP

grondo:
Thanks @chu11! I fixed those two issues you noted and force-pushed the result.

README.md
```diff
@@ -43,6 +43,7 @@ Table of Contents
 - [30/Job Urgency](spec_30.rst)
 - [31/Job Constraints Specification](spec_31.rst)
 - [32/Flux Job Execution Protocol Version 1](spec_32.rst)
+- [33/Flux Job Execution Module Protocol Version 1](spec_33.rst)
```
garlick, line 46: Suggestion: title the document something like "Distributed Job Control Protocol" or similar.

Main thought is to call out "distributed" to distinguish it from what we currently have, and to choose words that don't make it look too much like it is a replacement for RFC 32.

spec_33.rst
```diff
@@ -0,0 +1,328 @@
...
+
+Background
+----------
+
+RFC 32 describes the protocol between the execution service and job manager
+used to initiate and control jobs during the execution phase.  Upon receipt
+of a ``start`` request, the execution service is responsible for the launch,
+monitoring, and control of job shells on all execution targets involved
+in the job. Therefore, the execution service is necessarily distributed
+among all ranks of a Flux instance.
```
garlick, line 44: Since we have an execution service that isn't currently distributed (at least, not like this one), maybe this last sentence should be dropped or replaced with something like
> The initial execution service  was a minimum viable implementation concentrated on rank 0, launching remote processes using the simple `broker.rexec` protocol.  In contrast, the distributed job protocol takes advantage of the tree based overlay network structure to optimize performance, and is structured to allow jobs to be recovered upon Flux instance restart.  It also leverages some design insights from the early `wreck` execution system.

(This is background after all so it doesn't hurt to add some back story here)

spec_33.rst
```diff
@@ -0,0 +1,328 @@
...
+
+The job execution module protocol must adhere to the following criteria:
+
+ - Avoid global distributed operations which would require all ranks to
+   be online before the service is ready to execute work.
+
+ - Avoid presenting obstacles to the scaling of job size, the number of jobs
+   running concurrently, or job throughput.
+
+ - Support execution module reload.
```
garlick, line 61: Would it be too much detail to add "assuming modules are loaded in root to leaves order"?

spec_33.rst
```diff
@@ -0,0 +1,328 @@
...
+
+ - Support delivery of signals to jobs.
+
+Implementation
+--------------
+
+Job execution modules SHALL be loaded on all ranks in an instance, and are
+organized in a hierarchy with rank 0 at the root. Each module SHALL track
+the set of all running jobs for itself and all of its children. This state
+SHALL include at a minimum the jobid, userid, job state, and the idset of
```
garlick, line 86: One sentence refers to "the set" while next refers to "this state".   Maybe use "state" for the first one too?

spec_33.rst
```diff
@@ -0,0 +1,328 @@
...
+execution targets on which the job has an allocation.
+
+All job execution modules register a ``job-exec.hello`` service endpoint.
```
garlick, line 89: In other RFCs I think we've referred to the first word in the topic string as the "service" and the second as the "method".  So maybe s/service endpoint/service method/ ?

spec_33.rst
```diff
@@ -0,0 +1,328 @@
...
+Downstream execution modules send a ``hello`` request to their upstream
```
garlick, line 90: I'm not sure  how it would break down exactly, but subheadings might be useful to make the document easier to scan.  There's a transition from hello types to state update types that is a little run together here.

garlick (approved):
I think actually this is great as is.  I had a couple of nit picky comments at the beginning but as I went on I realized that this is probably going to change as we get experience prototyping and so getting too focused on those details is likely a distraction.  I left my initial comments in anyway, in case they are useful.

spec_33.rst
garlick: (other subheading comments deleted as "too soon")

spec_33.rst
grondo: Oh, I meant to remove this bullet, but yeah if we keep it then your suggestion should be added.

grondo:
>I think actually this is great as is. I had a couple of nit picky comments at the beginning but as I went on I realized that this is probably going to change as we get experience prototyping and so getting too focused on those details is likely a distraction. I left my initial comments in anyway, in case they are useful.

Feel free to push up any comments directly, I agree with all your comments. Whether you push individual commits or squash is up to you. I also imagine the details of the message types and contents will change as I try to implement a prototype.

garlick:
No worries, yeah I'll go ahead and push some changes if you're not working on it right now.

garlick:
> Feel free to push up any comments directly, I agree with them all. Whether you push individual commits or squash is up to you.

I made the changes I had suggested including the title change and other changes mostly confined to the introductory sections.   I didn't try to do any subheading reorg.  I pushed squashed my changes with yours, and then reread your comment and I realize you probably wanted me to  just squash my changes together.   Sorry?

grondo:
>  and then reread your comment and I realize you probably wanted me to just squash my changes together. Sorry?

No, I meant do whatever you preferred.

PR #330, opened 2022-04-23
rfc10: use hash digests in raw message payloads not blobrefs

garlick:
This changes the content load and store message definitions to carry raw hash digests rather than blobrefs, which reduces message size and processing overhead slightly.  This simplification is possible because we no longer expect to need to support multiple hash algorithms within a given instance, nor sharing of raw content between instances.  Also, we have restricted access by guests and recast the service as a KVS layer rather than general purpose.

It also feels like protocol cleanup, given that the blobref strings are being sent as raw payload.  If we're not sending them as strings within a JSON object, why not just send the more space efficient raw representation?

I also added some info that fills out this rather sparse RFC a bit.

I do have a prototype branch which implements these changes and uses the hash digest to as a zhashx key in the broker content cache, rather than running blobrefs through the default Bernstein hash that is the zhashx default which is...redundant.   I didn't notice any difference in the job throughput numbers in this branch though FWIW.  This still seems like good cleanup however.

### DIFF
--- a/spec_10.rst
+++ b/spec_10.rst
@@ -51,7 +51,13 @@ Store (KVS).  The goals of the content storage service are:
 -  The cryptographic hash algorithm is configurable per instance.
 
 This kind of store has interesting and well-understood properties, as
-explored in Venti, Git, and Camlistore (see References below).
+explored in Venti, Git, and Camlistore (see References below), for example:
+
+-  Writes are idempotent
+
+-  De-duplication is automatic
+
+-  The address may be used to check the integrity of the addressed content
 
 
 Implementation
--- a/spec_10.rst
+++ b/spec_10.rst
@@ -51,7 +51,7 @@ Store (KVS).  The goals of the content storage service are:
 -  The cryptographic hash algorithm is configurable per instance.
 
 This kind of store has interesting and well-understood properties, as
-explored in Venti, Git, and Camlistore (see References below), for example:
+explored in Venti, Git, and Perkeep (see References below), for example:
 
 -  Writes are idempotent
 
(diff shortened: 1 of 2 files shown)


spec_10.rst
```diff
@@ -51,7 +51,13 @@ Store (KVS).  The goals of the content storage service are:
 -  The cryptographic hash algorithm is configurable per instance.
 
 This kind of store has interesting and well-understood properties, as
-explored in Venti, Git, and Camlistore (see References below).
+explored in Venti, Git, and Camlistore (see References below).  for example,
```
grondo, line 54: This reminded me that Camlistore was renamed to Perkeep awhile ago. I guess the old references are still around, so perhaps this comment is just fyi

https://github.com/perkeep/perkeep

grondo (approved):
LGTM! Thanks for conscientiously keeping these RFCs up to date with the implementation!

spec_10.rst
garlick: Oh cool.  I'll update the doc.

> Thanks for conscientiously keeping these RFCs up to date with the implementation!

You're welcome: -)  But in case it wasn't clear, note that this also proposes a new change to the protocol.

garlick:
Ooops just noticed a punctuation problem and fixed it, forcing a push.

grondo:
Oops accidentally "resolved" the conversation to which I was replying. I meant to say that yes, I did see that this PR proposes a change in the protocol, but it is nice to see the RFC updated before or along with the implementation!

garlick:
I'll set MWP - thanks.

PR #331, opened 2022-05-05
[WIP] Add RFC33 draft

dongahn:
This is a WIP PR to add  RFC33: Flux multi-level queue scheduling architecture

Problem: Finalizing our multi-level queue scheduling subsystem design and implementation requires a more formal document that describes a distributed software architecture.

- Add the first draft of the multi-level queue scheduling architecture and subsystem to get comments. This RFC draft will continue to be refined as more comments and reference implementations are incorporated.

- Incorporate Sphinx mermaid plugin to support the rendering of a mermaid-specified figure in RFC 33.

In case the figure isn't rendered in the PR, I add the rendered figure here as well.

Screen Shot 2022-05-04 at 10 53 05 PM

### DIFF
--- a/spec_33.rst
+++ b/spec_33.rst
@@ -1,0 +1,269 @@
+.. github display
+   GitHub is NOT the preferred viewer for this file. Please visit
+   https://flux-framework.rtfd.io/projects/flux-rfc/en/latest/spec_33.html
+
+33/Multi-Level Queue Scheduling Architecture
+=================================================
+
+This document describes a software architecture for multi-level queue job
+scheduling within Flux.
+
+-  Name: github.com/flux-framework/rfc/spec_33.rst
+
+-  Editor: Dong H. Ahn <ahn1@llnl.gov>
+
+-  State: raw
+
+
+Language
+--------
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in `RFC 2119 <https://tools.ietf.org/html/rfc2119>`__.
+
+
+Related Standards
+-----------------
+
+-  :doc:`14/Canonical Job Specification <spec_14>`
+
+-  :doc:`25/Job Specification Version 1 <spec_25>`
+
+-  :doc:`27/Flux Resource Allocation Protocol Version 1 <spec_27>`
+
+
+Background
+----------
+
+High performance computing (HPC) workloads are becoming
+increasingly more diverse. The convergence of traditional HPC and new
+simulation, analysis, and data-science approaches including artificial
+intelligence (AI) provides unprecedented opportunities for discovery, but also
+creates diverse workloads more than ever before. Workloads from different users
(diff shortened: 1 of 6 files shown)


spec_33.rst
```diff
@@ -0,0 +1,269 @@
...
+-------------------------
+
+The above separation of concerns can be provided only when all of the distributed
+components that participate in this subsystem share the consistent view
+on the multi-level job queues. The best design to ensure or enforce configuration
+consistency for this architecture is actively being discussed, and the design will
+be specified in this document once finalized. This will include but not be
+limited to how the MLQS-specific per-queue controller can be configured,
+in particular when the configuration must come from external sources
+such as use accounting database.
```
cmoussa1, line 238: Did you mean to say **user accounting database** here? 

```suggestion
such as user accounting database.
```

spec_33.rst
```diff
@@ -0,0 +1,269 @@
...
+CONSERVATIVE backfilling (``K = unlimited``) and HYBRID backfilling (``K > 1 and < unlimited``).
+
+The limit controller within MLQS controls MLQS-specific limits that
+cannot otherwise accurately be enforced. The prominent limits of this
+class are various resource limits. Because a same jobspec can result
+in a different amount of resources assigned to the job (e.g., a jobspec
+requesting a single CPU core can be assigned to the entire compute node resources when
+node exclusive scheduling is used), only the limits controller within MLQS
+can enforce resource limits such as a maximum number of compute nodes per job
+or across all running jobs per user on the requesting queue.
```
cmoussa1, line 196: Not a suggestion here but just a thought: this `max_nodes` limit that needs to be enforced across all of a user's running jobs _probably_ makes the most sense to define in the user accounting database. So I might need to think about how to send this information either to the MLQS, or write it to this TOML file, where the MLQS can read it there. I'm not sure if this warrants defining such capability or interface in this RFC, but I figured I would throw this out there!

garlick (commented):
@dongahn, I apologize if we did not communicate well up front about the goals for this RFC, but what I was hoping for was a crisp definition of Flux queues, their essential properties, how this relates to specific motivating use cases, and then a proposal for core changes needed to elevate the queue design to the appropriate level.  I didn't expect a lot of that to materialize up front but I did hope for a structure that supports further discussion.

IMHO the expository verbiage here gets in the way of discussing the core concepts, and doesn't help me understand the motivation or the design.  The introduction of "controllers" and "MLQS" component seem like distracting fluxion details, not cross cutting concepts.  There are some important points in here, but they are hard to dig for.

What if instead of asking you to address my review comments, I were to try to propose something structured in more of an outline form to show you what I mean, and then we could use that as a basis for discussing the design, and fill it out as we work things out?  Then if I badly miss the mark we will discover quickly that our shared understanding isn't quite what we thought it was :-)

dongahn:
@garlick and @cmoussa1: Sound all good! If you propose an outline, we can start to co-author to get the gold nudget we need! Then the current draft can turn into an internal WIKI page somewhere for others who try to understand how the whole thing interoperate currently.

dongahn:
I am closing this and either send a rendered doc to some team members or post it in a WIKI somewhere.

PR #332, opened 2022-05-06
rfc33: add queues RFC

garlick:
Per discussion in yesterday's meeting and in #331, here's a proposed skeletal RFC on queues and some initial content designed to provoke discussion.  The goal was to strip down the prose and tie the design to the use cases proposed by @ryanday36.  I'm not sure I got there, especially on that last point, although it was making sense in my head.

This proposes that we make queues (plural) a first class abstraction in Flux, and that we let each queue have its own priority domain, scheduling algorithm, limits, access policies, etc..

It proposes that there is one default queue named **main** that is a stand-in for the current single anonymous queue, which you get for free without configuration.

Just remembered a couple of things I forgot, but I'll leave them off for now and give others a chance to weigh in:
- need a way to configure the name of the default queue, if **main** is not desired
- need a section on "resource affinity" (edit: nevermind, there is a placeholder)
- edit: discuss enforcement of limits/policy by scheduler or by job manager after validation

### DIFF
--- a/spec_33.rst
+++ b/spec_33.rst
@@ -1,0 +1,269 @@
+.. github display
+   GitHub is NOT the preferred viewer for this file. Please visit
+   https://flux-framework.rtfd.io/projects/flux-rfc/en/latest/spec_33.html
+
+33/Flux Job Queues
+==================
+
+This specification describes Flux Job Queues.  A Flux Job Queue is a named,
+user-visible container for job requests sorted by priority.
+
+
+-  Name: github.com/flux-framework/rfc/spec_33.rst
+
+-  Editor: Jim Garlick <garlick@llnl.gov>
+
+-  State: raw
+
+Language
+--------
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to
+be interpreted as described in `RFC 2119 <https://tools.ietf.org/html/rfc2119>`__.
+
+Related Standards
+-----------------
+
+-  :doc:`14/Canonical Job Specification <spec_14>`
+
+-  :doc:`20/Resource Set Specification Version 1 <spec_20>`
+
+-  :doc:`21/Job States and Events <spec_21>`
+
+-  :doc:`27/Flux Resource Allocation Protocol Version 1 <spec_27>`
+
+Background
+----------
+
+Support for multiple queues is motivated by the following use cases:
+
+- Schedule "debug" and "batch" jobs independently, on disjoint or overlapping
+  resource sets.
+
+- Allow jobs to be submitted in advance of dedicated application time,
+  to run on a resource superset.
+
+- Impose different limits on a given resource set.
+
+- Implement a different priority scheme and/or scheduling algorithm on
+  a given resource set.
+
+- Allow only certain users or banks to access a given resource set.
+
+- Allow for exceptions to limits and access controls.
+
+.. note::
+   Use cases are spelled out in more detail in
+   https://github.com/flux-framework/flux-core/issues/4306
+
+Design Criteria
+---------------
+
+- Handle the motivating use cases in the background section.
+
+- Minimize overhead for single-queue Flux instances, which will remain
+  the common case for non-system instances.
+
+- Minimize disruption to the original design of Flux with a single anonymous
+  queue.
+
+- Elevate named queues to a first class abstraction in Flux.
+
+- Promote a separation of concerns among Flux components such as scheduler,
+  job manager, accounting plugins, and Flux command line utilities.
+
+- Avoid overloading the queue abstraction.  For example, a new queue should not
+  be required merely to affect job priority or bypass limits (e.g. "standby",
+  "expedite", or "exempt" queues).
+
+Implementation
+--------------
+
+A Flux Job Queue is a user-visible container for job requests stored in
+priority order.
+
+Queue configuration is OPTIONAL.  If queues are not configured, the Flux
+instance SHALL have one anonymous queue.  If queues are configured, then
+all queues SHALL be named.  If more than one named queue is configured,
+one queue MUST be designated as the default queue by configuring
+``policy.jobspec.defaults.system.queue`` (see below).
+
+Queues MAY be independently configured with:
+
+- jobspec defaults policy
+
+- limits policy
+
+- scheduler policy
+
+- priority policy
+
+- access policy
+
+- resource subset
+
+Policy Configuration
+~~~~~~~~~~~~~~~~~~~~
+
+A ``policy`` TOML table MAY specify queue policy that applies to all queues,
+including the anonymous one, unless overridden on a per-queue basis.  This
+table contains OPTIONAL sub-tables for ``jobspec.defaults``, ``limits``,
+``access``, and ``scheduler``.
+
+Each queue MAY contain a ``policy`` table.  If present, queue-specific policy
+values SHALL override global policy values.
+
+Jobspec Defaults Policy
+^^^^^^^^^^^^^^^^^^^^^^^
+
+The ``jobspec.defaults`` policy table contains default values for jobspec
+attributes that were not explicitly set by the user and MAY contain following
+OPTIONAL keys.  The key names are identical to the corresponding jobspec
+attribute names (RFC 14, 25).
+
+system.duration
+  (string) default duration, in Flux Standard Duration format (RFC 23).
+
+system.queue
+  (string) default queue name.
+
+.. note::
+   Jobspec updates are applied to the in-memory copy of the jobspec held
+   by the job manager.  The in-memory copy is also sent to the scheduler with
+   the job's ``alloc`` request.  The original jobspec in the KVS (both signed
+   and unsigned) is unaltered.  Since jobspec modifications are posted to the
+   job eventlog as ``jobspec-update`` events (RFC 21), the altered jobspec
+   may by reconstructed by fetching the original jobspec, then replaying any
+   ``jobspec-update`` events.
+
+Limits Policy
+^^^^^^^^^^^^^
+
+The `limits` policy table configures job limits and MAY contain the following
+OPTIONAL keys.
+
+job-size.max.nnodes
+  (integer) maximum number of nodes.
+
+job-size.max.ncores
+  (integer) maximum number of cores.
+
+job-size.max.ngpus
+  (integer) maximum number of gpus.
+
+job-size.min.nnodes
+  (integer) minimum number of nodes.
+
+duration
+  (string) maximum job duration, in Flux Standard Duration format (RFC 23).
+
+.. note::
+   A general mechanism for configuring and applying limits in a distributed
+   fashion is proposed in rough form in
+   https://github.com/flux-framework/flux-core/issues/4309.
+   Consider creating an RFC for limits that can be referenced from this one
+   and skip those details here.
+
+
+Scheduler Policy
+^^^^^^^^^^^^^^^^
+
+The ``scheduler`` policy table is read by the scheduler implementation
+and is opaque to the rest of Flux.
+
+Priority Policy
+^^^^^^^^^^^^^^^
+
+TBD
+
+Access Policy
+^^^^^^^^^^^^^
+
+The ``access`` policy table MAY restrict queue access by UNIX user and group.
+It MAY contain following OPTIONAL keys:
+
+allow-user
+  (list of strings) Specify a list of UNIX user names that are to be granted
+  access.
+
+allow-group
+  (list of strings) Specify a list of UNIX group names that are to be granted
+  access.
+
+The absence of ``allow-user`` and ``allow-group`` keys indicates that no queue
+access restrictions are in place.  However, the access policy MAY be extended
+by a jobtap plugin that enforces additional access conditions.  For example,
+the flux-accounting multi-factor priority plugin controls access to queues
+based on the user and bank information from the accounting database.
+
+Queue Configuration
+~~~~~~~~~~~~~~~~~~~
+
+A ``queues`` TOML table MAY define one or more named queues.  Each queue
+SHALL be represented as a sub-table that MAY contain the following OPTIONAL
+keys:
+
+requires
+  (table) Specify queue-specific resource constraint(s) in RFC 31 format,
+  that SHALL be added to the jobspec ``system.constraints`` attribute of all
+  jobs submitted to this queue.  If the jobspec already specifies constraints,
+  then the queue-specific constraints SHALL be added with an ``and`` operator.
+
+policy
+  (table) Specify policy fragments that apply only to this queue, using the
+  form described in the previous section.  If the same policy appears in the
+  top-level ``policy`` table  and a queue-specific ``policy`` table, the
+  queue-specific value takes precedence for jobs submitted to that queue.
+
+
+Initial Assignment of Job to Queue
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Job requests MAY specify a queue name at submission time by setting the
+``queue`` jobspec system attribute (RFC 14).  If a queue was not explicitly
+named in the jobspec, and a default queue is defined, the queue SHALL be
+assigned by the job manager during the initial request validation,
+before the jobtap ``job.validate`` callbacks are run.
+
+Request Validation
+~~~~~~~~~~~~~~~~~~
+
+A job request SHALL be rejected on submission if it names an unknown queue,
+or if it is possible to determine that the job would exceed limits or violate
+access policy of the assigned queue.
+
+Administrative Tools
+~~~~~~~~~~~~~~~~~~~~
+
+A Flux command line tool SHALL provide the ability to enable/disable job
+submission on each queue individually, or on all queues.
+
+A Flux command line tool SHALL provide the ability to start/stop scheduling
+on each queue individually, or on all queues.  When scheduling is stopped,
+any pending ``alloc`` requests to the scheduler SHALL be canceled.
+
+A Flux command line tool SHALL provide the ability to wait for a queue to
+become empty, or for all queues to become empty.
+
+A Flux command line tool SHALL provide the ability to wait for a queue to
+become idle, or for all queues to become idle, where idle is defined as
+containing no jobs in RUN or CLEANUP state.
+
+Job Submission and Listing Tools
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Job submission and listing tools SHOULD NOT need to parse the ``queues``
+TOML table.
+
+The service providing data to the job listing tool SHOULD list pending and
+running jobs in the default queue by default.  An option SHALL be provided
+to request jobs in other queues by name, or all queues.
+
+The job submission tools SHOULD leave the queue unset (thereby selecting
+the default.  An option SHALL be provided to direct jobs to other
+queues by name.
+
+.. note::
+   The ``sched.queue`` annotation defined in RFC 27 is no longer necessary
+   since the queue name is represented in the job jobspec as described above.
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ Table of Contents
 - [30/Job Urgency](spec_30.rst)
 - [31/Job Constraints Specification](spec_31.rst)
 - [32/Flux Job Execution Protocol Version 1](spec_32.rst)
+- [33/Flux Job Queues](spec_33.rst)
 
 Build Instructions
 ------------------
--- a/index.rst
+++ b/index.rst
@@ -228,6 +228,12 @@ job constraints.
 This specification describes Version 1 of the Flux Execution Protocol
 implemented by the job manager and job execution system.
 
+:doc:`33/Flux Job Queues <spec_33>`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This specification describes Flux Job Queues. A Flux Job queue is a named,
+user-visible container for job requests sorted by priority.
+
 .. Each file must appear in a toctree
 .. toctree::
    :hidden:
@@ -263,3 +269,4 @@ implemented by the job manager and job execution system.
    spec_30
    spec_31
    spec_32
+   spec_33
--- a/spell.en.pws
+++ b/spell.en.pws
@@ -451,3 +451,9 @@ afternotok
 login
 satisfiability
 satisfiable
+validator
+superset
+ncores
+ngpus
+ncpus
+jobtap


spec_33.rst
```diff
@@ -0,0 +1,189 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Job requests MAY specify a queue name at submission time by setting the
+``queue`` jobspec system attribute (RFC 14).  If a queue is not specified
+by the jobspec, the *main* queue SHALL be assumed.
+
+.. note::
+   The queue name can be determined by examining jobspec, but a job cannot
+   be added to a queue until it is assigned a priority, on state transition
+   from PRIORITY to SCHED (RFC 21).
```
grondo, line 117: I'm probably not understanding the motivation of this note, but if a job can't be added to a queue until it has a priority, but each queue might use a different priority calculation, does this lead to a catch-22? I.e. we can't assign priority because a job has not been added to a queue, but we can't add it to a queue until a priority has been assigned?

spec_33.rst
```diff
@@ -0,0 +1,189 @@
...
+  validator plugin)
+
+Queue Limits
+~~~~~~~~~~~~
+
+The instance owner MAY configure special limits for queues such as:
+
+- maximum or minimum number of nodes
+
+- maximum (or default?) time limit
```
grondo, line 143: I would say maximum _and_ default time limits should be able to be assigned.

A job SHALL be rejected at validation if it exceeds the configured maximum time limit (duration in jobspec parlance) for the selected queue (or `main` if no queue is set in jobspec)

Conformant schedulers SHALL use the default queue time limit when setting the expiration of resource sets generated for jobs which have duration set to unlimited in their jobspec.

spec_33.rst
```diff
@@ -0,0 +1,189 @@
...
+queue and multi-factor flux-accounting priority in another.
+
+Scheduling Algorithm
+~~~~~~~~~~~~~~~~~~~~
+
+TBD - each queue should be able to independently configure a scheduling
+algorithm.  For example, FIFO in one queue, fair-share in another.
+
+TODO - how to have separation of concerns and not configure scheduler/algorithm
+details in multiple places?
```
grondo, line 165: Would it be reasonable to state that scheduler-specific queue configuration still resides under the `queues` TOML table? We could either state the list of reserved keys and allow schedulers to extend that list, or we could try to come up with an encompassing list that would work for all known cases here.

An issue with the first proposal is that it would be difficult to detect unused keys, which could be the result of typos. In the second case, we may be limiting the extensibility of the queue configuration, but we could at least detect and emit errors on unsupported configuration keys.

spec_33.rst
```diff
@@ -0,0 +1,189 @@
...
+
+Resource Affinity
+~~~~~~~~~~~~~~~~~
+
+TBD - it should be possible to designate a queue for a specific resource subset,
+e.g. a partition.  OTOH, we want to avoid *requiring* resource fragmentation
+just to get a simple batch/debug queue separation.
```
grondo, line 172: One idea here would be to use resource properties as tags for resources assigned to queues. If a queue does not have a set of required properties, then that queue would apply to the entire resource set. Since resources can have multiple properties assigned, this easily allows overlapping resource sets.

For the purposes of the RFC, we could describe the properties configuration of the queue, and require that conformant schedulers append the required queue properties to all allocation requests for the given queue.

grondo (commented):
Thank you! This is a great start and I think starting from the small, specific ideas here  will really help us narrow down the minimum requirements here.

I placed some discussion comments inline that came to me as I was reading through the document.

Again, thanks!

spec_33.rst
garlick: This was just a reminder that although a job _has_ a queue assignment at submit time, it isn't given a priority until later and thus cannot have a valid position in the queue.  It's nothing new, that's how it works now - I just threw it in there because it seemed like a confusing tidbit (is the job _in_ the queue?).

There is no catch-22.  A queue specific priority calculation can happen, and then the priority can be used to determine the job's position in the queue.

I should remove the note.

spec_33.rst
garlick: Very good idea!

spec_33.rst
```diff
@@ -0,0 +1,189 @@
+.. github display
+   GitHub is NOT the preferred viewer for this file. Please visit
+   https://flux-framework.rtfd.io/projects/flux-rfc/en/latest/spec_33.html
+
+33/Flux Queues
```
dongahn, line 5: I like the brevity of this! 

And I understand you don't need to answer all questions in the title. 

But I would consider using Flux Job Queues or something  to make it a bit more descriptive. Queues are used in many difference places like "Message Queues."

spec_33.rst
garlick: I'll push some changes based on these suggestions.  If nobody is reviewing the original draft now, I'll squash them (please speak up if you in the middle of reviewing)

spec_33.rst
```diff
@@ -0,0 +1,189 @@
...
+   Use cases are spelled out in more detail in
+   https://github.com/flux-framework/flux-core/issues/4306
+
+
+Design Criteria
+---------------
+
+- Handle the motivating use cases in the background section.
+
+- Minimize disruption to current single-queue Flux design, which will remain
```
dongahn, line 64: I believe flux developers will understand the different between "single-queue" Flux design and Flux queues. But I think this can be a bit confusing for others who want to make contributions later. Perhaps you have this "single queue" implementation captured somewhere in another RFC, which you can refer back? Or maybe you are considering this single queue as "a queue" of Flux Queues you are specifying?

spec_33.rst
```diff
@@ -0,0 +1,189 @@
...
+  the common case for non-system instances.
+
+- Elevate named queues to a first class abstraction in Flux.
+
+- Promote a separation of concerns among Flux components such as scheduler,
```
dongahn, line 69: Nice!

spec_33.rst
```diff
@@ -0,0 +1,189 @@
...
+.. note::
+   :doc:`RFC 27 <spec_27>` will require changes to support multiple queues,
+   especially in combination with ``mode=limited``, since the highest priority
+   jobs *in each queue* must reach the scheduler in the form of alloc requests
+   before lower priority jobs.
+
+Queue Configuration
+~~~~~~~~~~~~~~~~~~~
+
+All Flux instances SHALL have a default queue named *main* which requires
```
dongahn, line 98: Ah so the "main" queue is the current single queue used in job manager?

spec_33.rst
```diff
@@ -0,0 +1,189 @@
...
+The service providing data to the job listing tool SHOULD list pending and
+running jobs in the *main* queue by default.  An option SHALL be provided
+to request jobs in other queues by name, or all queues.
+
+The job submission tools SHOULD leave the queue unset (thereby selecting
+*main*) by default.  An option SHALL be provided to direct jobs to other
+queues by name.
+
+.. note::
+   The ``sched.queue`` annotation defined in RFC 27 is no longer necessary
```
dongahn, line 188: Nice

dongahn:
@garlick: this is really nice as it will raise the concept of multiple queues as the first class citizen. The only thing that can still have some more discussion would be, like you said, regarding "cross cutting". How would a multi-level queue scheduler like fluxion can work in this context -- who is responsible for what. 

This is a great start!

spec_33.rst
garlick: I guess it's just a note to remind us that the current single-queue design is still a common case because of Flux's hierarchical nature.  Also that we this design is influenced by what we have now and the tight schedule that we are on - we don't have time to get too fancy :-)

Maybe I should just say that.

spec_33.rst
garlick: That's what I was thinking.   Then we can say "every job is assigned to a named queue".

spec_33.rst
dongahn: Yes I like it. Just whatever to clarify this. Yup big fan of not being too fancy here ;-)

spec_33.rst
dongahn: Yup.

spec_33.rst
garlick: Great point, will change.

spec_33.rst
grondo: Got it, it makes sense now that you've explained it :-)

garlick:
Just pushed some changes based on @dongahn and @grondo's comments.  I didn't squash.  LMK if you'd like me to do that.

I think I addressed everything above except one comment so I'll _resolve_ all the comments I addressed and be ready for another round :-)

grondo:
Feel free to squash as far as I'm concerned.

dongahn:
I agree -- just one reminder for myself.  I currently have "default" as the name of the default queue so that needs to change to "main" when this RFC is landed.

garlick:
Ok, I squashed and force pushed.

>  currently have "default" as the name of the default queue so that needs to change to "main" when this RFC is landed.

People might want a different name for the default queue on a system instance, so I changed the wording slightly to allow that.  So I guess if the queue is not defined in the jobspec, you'd first check for a `queues.default`(or whatever we decide) in the config, and then use main if that is not defined.  Perhaps that should be spelled out.

dongahn:
>People might want a different name for the default queue on a system instance, so I changed the wording slightly to allow that. So I guess if the queue is not defined in the jobspec, you'd first check for a queues.default(or whatever we decide) in the config, and then use main if that is not defined. Perhaps that should be spelled out.

Oh nice! Thank you for adding that @garlick!

garlick:
This needs to be addressed in the RFC:

@ryanday36 said in flux-framework/flux-core#4306
> give users a big priority boost.
E.g. users / jobs may be given access to an ‘expedite’ qos (Slurm) or ‘expedite’ queue (LSF) which does the same thing as ‘exempt’, but also has a higher baseline priority factor than the ‘normal’ qos (Slurm) or ‘pbatch’ queue (LSF).

Do we need to associate a priority level with each queue?

This seems like it leads to difficult questions, like how do queues running different scheduling algorithms prioritize resource assignment when drawing from the same pool?  Does the queue with the highest priority always win and starve the other one out, and round-robin if priorities are equal?

Round robin hardly seems right if jobs in one queue have a higher _job_ priority than jobs in the other.  OTOH always running the highest priority job first is FIFO behavior, not necessarily behavior of other scheduling algorithms more likely to be used.  How can a scheduler make a sane plan to service N queues with M algorithms from the same resource set?

Could we constrain the problem a bit to get on a better path?  For example, only permit queues to draw from the same resource set if running the same scheduling algorithm?  Then the scheduler can sanely plan one schedule for a given resource set, and merely use the additional queue to selectively boost or reduce priority?

grondo:
Keen observations. For the use case mentioned above (an expedite queue) it does seem like the one queue starving out the other queue is the desired effect. However, in general your other points all apply when you allow queues to use overlapping resource sets.

There are also more practical issues, like if a queue is configured with more permissive limits but is otherwise is configured identical to the `main` queue, users might expect job listing tools to list their jobs in the `main` queue listing. Maybe this is why Slurm chose to use QoS instead of queues for this kind of thing.

I don't think I have any answers, but I wonder if trying to use overlapping queues for some of these use cases is making things more difficult than it needs to be. It probably isn't too late to brainstorm other ideas.

One idea, and I'm not quite sure if it is a good idea or completely idiotic, would be to allow a single queue to have multiple submission names, each could have a different priority and limits configuration. However, jobs would still effectively be in the same "parent" queue. The TOML configuration could be hierarchical, and the name of these subqueues could be `<parent>.<name>`, e.g. `main.expedite`.

We'd still have the issue with overlapping queues, but maybe this feature would remove the requirement for overlapping queues in many cases. In cases where top-level queues have overlapping resources, maybe we can leave it up to the scheduler and just note the possible complications of overlapping resource sets assigned to top level queues?

garlick:
@dongahn, how does fluxion fulfill requests from multiple queues?  Does it fulfill rall equests in a global priority order without regard for which queue they are coming from?

Is backfill employed independently in each queue, or holistically, e.g. only once no queues have runnable jobs in them?

Is there a way to "partition" resources, e.g. designate a particular set of nodes or a node count for the debug queue?

cmoussa1:
>> give users a big priority boost.
E.g. users / jobs may be given access to an ‘expedite’ qos (Slurm) or ‘expedite’ queue (LSF) which does the same thing as ‘exempt’, but also has a higher baseline priority factor than the ‘normal’ qos (Slurm) or ‘pbatch’ queue (LSF).
>
> Do we need to associate a priority level with each queue?

I can say that from the flux-accounting side right now, associating a priority level with each queue is something that's defined in the flux-accounting DB and is part of the information that is sent to the priority plugin. For example, an `expedite` queue could have a much higher integer priority vs a `default` queue. Then, a user who has permission to submit jobs uinder this `expedite` queue could submit said job with `--setattr=system.queue=expedite` and increase the overall priority of their job.

garlick:
Thanks - to be sure I understand, the queues are processed one after the other, as though the requests were concatenated in one queue?

dongahn:
>Thanks - to be sure I understand, the queues are processed one after the other, as though the requests were concatenated in one queue?

Yup: https://github.com/flux-framework/flux-sched/blob/master/qmanager/modules/qmanager_callbacks.cpp#L393

garlick:
Based on coffee discussion today, I changed the intro a bit to focus the role of queues.    Specifically, 
> Avoid overloading the queue abstraction.  For example, a new queue should not
be required merely to affect job priority or bypass limits (e.g. "standby", "expedite", or "exempt" queues).

This reflects the idea that those things would be handled some other way.  So in this sense Flux queues would be more like slurm queues and less like LSF queues.

dongahn:
>> Do we need to associate a priority level with each queue?

> This seems like it leads to difficult questions, like how do queues running different scheduling algorithms prioritize resource assignment when drawing from the same pool? Does the queue with the highest priority always win and starve the other one out, and round-robin if priorities are equal?

@garlick: somehow I didn't read your earlier posting. These are really keen observations and just one thought.

I think one of the reasons we want to have multiple queues in the first place is so that the system can achieve both fairness and responsiveness at the same time. If you have a backfilling policy on a queue, this can help running small/short running jobs but generally speaking there will be cases where a 30 min 1 node job waiting 12 hours because of large long running batch jobs. 

A general solution is to have some amount of resources more readily used by (or preferred) by smaller/shorter running jobs.

For overlapping resources, perhaps we can have a fixed priority order among queues (e.g., pdebug jobs are scheduled first before considering pbatch jobs at every scheduling cycle) but allow sysadmins to impose a limit (e.g., the total amount of resources used by pdebug jobs will not exceed 10% lest we won't schedule)?  

This still wouldn't provide better guarantees for responsiveness or starvation but still worthwhile to support for overlapping queues? (Having distinct partitions solve the most of these issues but it comes at the cost of lower resource utilization so a tradeoff).

garlick:
> This still wouldn't provide better guarantees for responsiveness or starvation but still worthwhile to support for overlapping queues? (Having distinct partitions solve the most of these issues but it comes at the cost of lower resource utilization so a tradeoff).

We will need to allow overlapping queues to support the "pall" use case, where jobs for a dat are submitted before the dat time slot arrives.  If we extend our tools to allow submission to be enabled/disabled per queue, and scheduling enabled/disabled per queue then I think we can fulfill that use case without hitting flux-framework/flux-sched#939.  We already have a tool for this but it presumes Flux has only the one queue:
```
$ flux queue
Usage: flux-queue [OPTIONS] COMMAND ARGS
  -h, --help             Display this message.

Common commands from flux-queue:
   enable          Enable job submission
   disable         Disable job submission
   start           Start scheduling
   stop            Stop scheduling
   status          Get queue status
   drain           Wait for queue to become empty.
   idle            Wait for queue to become idle.
```

I agree that it's desirable to allow pdebug/pbatch to share a resource pool, but I imagine we can't do it without guaranteeing progress in both queues, so I would suggest this is a future effort and let's collect thoughts in flux-framework/flux-sched#939.

garlick:
I was thinking that those thing would go in a separate RFC, since queues are already a rather big topic.   So maybe we need two RFCs in addition to this one, for:
- limits e.g. maybe expanding on flux-framework/flux-core#4309
- priority/limit  administrative exceptions

Does that seem reasonable?  My intent was to pare down the use cases in this RFC to a subset of flux-framework/flux-core#4306.

ryanday36:
I think that a separation of concerns between queues and limits / policy is reasonable. At some point, and maybe this is yet another RFC or maybe it goes into the limits / policy RFC, we need to be able to associate queues and policy though. So, if the queue abstraction isn't going to know about policy, then the policy abstraction needs to know about queues.

It might also help to modify the "Background" section of this RFC to make it more clear what aspects of the use cases will be covered. E.g. "Impose different limits on a given resource set." doesn't sound like it's covered by this RFC anymore.

garlick:
Well I was thinking it would be OK to define queue-specific limiits / policy here.  My main points were that
- queues should not be the only tool for manipulating priority/limits/policy (something else is needed, in slurm that is QoS)
- this RFC should focus on queues that are similar to slurm partitions; hence this RFC won't be a complete solution to all the uses cases you articulated in that issue

I'm not sure what the entire picture looks like.  More discussion needed!

Everything discussed here is still open for change, if for example you'd like to argue for LSF style queues.  I really do like the "standby queue" concept since it's so easy to understand.  But the justification seems weak for the scheduler to be presented with an actual separate queue of jobs when those jobs could be added to the main queue at a low priority (with a preemptible flag or whatever).

ryanday36:
That makes sense. I'm also mostly just trying to get a better idea of what the overall picture looks like. I don't think I particularly want to argue for one approach or another, other than that you should do whatever approach works best with how Flux is architected. What that picture looks like to me right now is that flux-core will know about multiple queues, which are priority ordered lists of jobs that it will assign to resources as they become available, and flux-sched will put jobs into those queues according to policies, which will be sets of limits / other TBD access policies that will define whether or not a job is eligible to be added to a given queue and will be described in another RFC. Does that fit with how you all are seeing this?

I haven't really put the user in this picture, and what they need to put in their jobspec for all this. That's where the ease of understanding for LSF style queues is nice. I.e. only needing to specify a queue rather than Slurm's partition+qos. I'm not sure the users actually need to know about the queue though. The important part for the user is what policy / limits their job is bound by, so maybe they would only need to specify the policy and that plus their requested resources would tell flux-sched enough so that it would assign the job to the appropriate queue.

The ideal, from a simplifying things for the user standpoint, would probably actually be to not have to specify either. E.g. if the user submits a small, short job that fits within the limits / policy defined for pdebug, that job gets put in the pdebug queue without the user needing to tell us that it's a pdebug job. The user just tells the scheduler what they want and the scheduler figures out where to run it based on the policies that the scheduler is configured with. I don't recall the details, but I believe LBL has a perhaps extreme example of this hacked onto Slurm with something like hundreds of partitions/QOS and a complex job_submit.lua that automatically assigns jobs to the appropriate place based on user/bank/requested resources/etc.

garlick:
> flux-sched will put jobs into those queues according to policies, which will be sets of limits / other TBD access policies that will define whether or not a job is eligible to be added to a given queue and will be described in another RFC. Does that fit with how you all are seeing this?

Yes although I'm proposing that the queue is selected up front, at job submission time, not by flux sched.  It would immediately fail if the user doesn't have access to the specified queue or the request exceeds configured limits.  The queues would exist in the job manager and could be administratively controlled with `flux-queue` (e.g. scheduling can be started/stopped, submission can be started/stopped, they can be drained, etc).

So for example, jobs could be submitted to a "stopped" _pall_ queue before a DAT and the scheduler would not see those jobs until the queue is started on DAT day (and presumably the _pbatch_ queue would be stopped).

Regarding what the user sees, I was picturing that the user would specify both queue name and any special privileges they want (whatever form we decide that takes) on the command line.   It does sound very old school now that you mention it....

garlick:
Just pushed a couple more changes
- queue name is added to job _submit_ event
- add admin controls for start/stop, enable/disable, drain, and idle on individual queues

grondo:
The idea of routing a job to a queue which fits the request is interesting. I hadn't considered that before. It does require that users set enough information in the job to select a valid queue (e.g. a short duration or an interactive request for the debug queue), so I'm not sure it really saves them too much typing, but still it seems like a pretty clever idea.

I think @garlick's idea to issue the queue name in the `submit` event would allow a jobtap plugin, or stack of plugins, to select a queue for the job, if we expose that ability to the plugins during the validate callback.

garlick:
Well technically the `submit` event is posted by the ingest service.  We could always do it a bit later as a separate event.

garlick:
Wait a sec, could we get batch and interactive jobs running out of the same queue if we gave interactive jobs a priority boost and limited the maximum job size to somewhat less than the number of nodes in the partition?  Do we even need these damn queues?  :smiley_cat:

garlick:
Of course the usual problem though, interactive jobs could take over the system.  Sorry, things are never that simple.

grondo:
> Well technically the submit event is posted by the ingest service. We could always do it a bit later as a separate event.

Oh yeah, thus we have the special case code for the `submit` event in the job manager.

It might be good to have the queue (and possibly other related job properties) logged in a subsequent event. Then the job manager and its plugins can be the one entity that is knowledgeable about the related configuration (which in the case of job-accounting, may come from an external project)

grondo:
> Of course the usual problem though, interactive jobs could take over the system. Sorry, things are never that simple.

If "interactive" was a property of a job though, then you could do something like @dongahn was suggesting and ensure that jobs with +interactive never utilize more than 10% of the system.

I guess it is all just different nomenclature for the same thing.

grondo:
> It might be good to have the queue (and possibly other related job properties) logged in a subsequent event.

Alternately, we could leave the door open to adding an event that moves a job to a new queue. Then a plugin could move jobs from the default queue only when configured to do so (this would also allow admins to move jobs between queues if supported). The submit event might have to indicate whether the chosen queue was explicitly specified or not for this to work properly. (And, we could always add support for this later, it doesn't have to be done up front I don't think)

garlick:
It seems like we'll want to assign a queue as early as possible so that job listing tools that might list jobs "by queue" won't have to deal with unassigned jobs.  We can certainly leave the door open to move jobs to a new queue later on, and add a flag indicating whether the queue was selected explicitly or implicitly to make that easier.  The ingest module needs to parse the config to find the default queue, but I don't think that's a big deal.

grondo:
> Edit: oops, limits. It doesn't make sense to apply limits for the default queue if that's not where it's going to stay. Hmm, perhaps my concern about unassigned jobs is moot as long as the queue gets assigned quickly after submission.

If we accept that jobs can potentially move between queues, then we have to support adjustment of limits upon queue reassignment anyway. Since that is something we'd incorporate down the road, perhaps we don't need to worry about it for now.

I kind of like the idea of the first job eventlog event containing the queue assigned strictly by jobspec or configuration. Note that the submit event is posted, but jobs are not "announced" nor is a response sent to the job.submit request until the job manager has validated the submitted job (i.e. all validate plugins have run) (That is really neither here nor there, but if in the future a plugin decides to move a job to a new queue during the validate callback, that event will be emitted nearly at the same time the job is fully ingested)

garlick:
Ah OK, so the ingest module could assign the queue and put it in the submit event, and then during validation, a jobtap plugin could optionally make a reassignment before we check any limits.  Nice!

ryanday36:
> It seems like we'll want to assign a queue as early as possible so that job listing tools that might list jobs "by queue" won't have to deal with unassigned jobs. We can certainly leave the door open to move jobs to a new queue later on, and add a flag indicating whether the queue was selected explicitly or implicitly to make that easier. The ingest module needs to parse the config to find the default queue, but I don't think that's a big deal.
> 
> Edit: oops, limits. It doesn't make sense to apply limits for the default queue if that's not where it's going to stay. Hmm, perhaps my concern about unassigned jobs is moot as long as the queue gets assigned quickly after submission.

I'll just note that LSF has a 'TBD' state and Slurm has a 'WaitingForScheduling' job reason code for jobs that are awaiting full validation, priority assignment, etc, so I think it's not unreasonable to have jobs spend a brief amount of time in some sort of unassigned state. Of course, there's still a lot of value in keeping that unassigned time brief. If you ask me to choose between the LSF 'TBD' state that I've seen pretty regularly and the Slurm 'WaitingForScheduling' state that I never see and had to go digging into a man page to confirm, I'd choose Slurm.

dongahn:
> If "interactive" was a property of a job though, then you could do something like @dongahn was suggesting and ensure that jobs with +interactive never utilize more than 10% of the system.

> I guess it is all just different nomenclature for the same thing.

Agreed. 

Multiple queues are just one way to achieve both fairness and responsiveness along with other scheduling effectiveness metrics. If you use a single queue and try to deal with a wide range of different class of jobs through that (short running, long running, small jobs and larger jobs), you may still be able to do that but the policy and limit logic complexity on the queue will need to become pretty high. 

With multiple queues, this can be achieved by using one additional queue --  without increasing software complexity too much. But then having multiple queues are good as far as we produce enough knobs to admins/users to achieve key performance metrics. And this can open up a door for admins to overdo it and for users to game the system. (@ryanday36's example above).

In general, using an easy-to-understand software architecture is highly desirable and perhaps should be a significant design goal. Understanding the combined effect of policies and other control is already difficult. The SW architecture doesn't have to add more complexity to this complex picture.

I'd also think that once we have a few knobs, working with researchers to do some cheap exploration about the combined effects of scheduling does make sense to me as well, which can inform the centers for their flux scheduling administration.

garlick:
flux-framework/flux-core#3910 proposed a way for jobtap pluggins to modify jobspec, posting an event so that the modifications are reapplied after a flux restart (and there is a record).

That might be a way to fill in defaults for
* `system.queue`
* `system.duration`
* `system.constraints.properties`

I think the idea is that an event would be posted containing a replacement jobspec key, and then the "action" for the event would be to modify the redacted jobspec that is held in memory in the job manager, and is also passed to the scheduler.

garlick:
Here's a straw man toml configuration for consideration.
```toml
[job-manager]

default-queue = "pbatch"

[queues]

  [queues.pdebug]
  requires = [ "pdebug" ]
  default.duration = "15m"
  limit.duration = "30m"
  limit.max-node-count = 4

  # scheduler-specific, per-queue configuration (opaque to other subsytems)
  scheduler-options = {
    queue-policy = "hybrid"
  }

  [queues.pbatch]
  requires = [ "pbatch" ]
  default.duration = "4h"
  limit.duration = "24h"
  limit.max-node-count = 64
  scheduler-options = {
    queue-policy = "hybrid"
  }
```

grondo:
Great! I was having the same thought yesterday that the job-manager could take care of adding queue-based constraints to jobspec before sending to the scheduler.

Some other notes:

@trws had mentioned that we will hard and soft values for each limit. My assumption is that a job submitted above a soft limit but under the hard limit would be held (e.g. perhaps allowing time-based limits) and above the hard limit, jobs would be rejected. I'm not sure if this is a near term requirement thought. Perhaps @trws and @ryanday36 could chime in.

Note that the job manager will likely not be able to enforce a job size limit based on number of nodes, since it can't know how many nodes might be allocated to a request that only indicates slots and slot size.

cmoussa1:
Thanks @garlick! I'll also note that I believe one of the other per-queue limits @ryanday36 mentioned was a `min-nodes-per-job` limit, so we perhaps need to also add that limit to the TOML configuration:

```toml
[queues]

  [queues.pdebug]
  requires = [ "pdebug" ]
  default.duration = "15m"
  limit.duration = "30m"
  limit.max-node-count = 4
  limit.min-node-count = 1
```

garlick:
Ah good points about node count enforcement and soft limits.  I liked your ideas in  flux-framework/flux-core#4309 since potentially job manager could add limits to a job and then other subsystems could resolve them before allowing the job to proceed.

I'm a little concerned about the requirement that all jobs have a named queue that I put in the rfc 33 draft, when  most instances will just have the one, especially if additional events have to be generated.    Maybe we could just have a jobtap plugin that is loaded in the system instance that issues the jobspec change events, and otherwise have the job manager remain ignorant of queues.  I'm not sure what the ramifications are for user facing utilities though.

grondo:
> Maybe we could just have a jobtap plugin that is loaded in the system instance that issues the jobspec change events, and otherwise have the job manager remain ignorant of queues. I'm not sure what the ramifications are for user facing utilities though.

Yeah, that sounds like a good idea.

I don't think it will be a problem for user facing utilities. The job-list module or equivalent can store the queue as a first-class attribute, but the queue attribute can be empty just like several other attributes. Or, perhaps I'm missing what other issues you envision face user utilities?

grondo:
>  job manager could add limits to a job and then other subsystems could resolve them before allowing the job to proceed.

Random thought:

Perhaps if we can sketch out what a job limit event looks like in the job eventlog, we can use the same format for limits in the queue TOML config. I'm not sure what form a limit JSON object might take,  but if we want to allow hard/soft limits and an extensible format, maybe something where each limit has a `type`, or `name` which acts as the scheme or class of the limit. SImilar to other scheme-based solutions each scheme could have a different handler for enforcing or checking the limit. The type-dependent values within the limit object could optionally have hard/soft components, but each limit type _may_ have a different representation.


```json
{
    "type": "job-size",
    "hard-limit": {
        "nnodes": 24,
        "ncores": 384,
        "ngpus": 48,
    },
    "soft-limit": {
        "nnodes": 8,
        "ncores": 128,
        "ngpus": 16,
    }
}
```

Not sure if this is a good idea or not, but throwing it out there anyway...

ryanday36:
> @trws had mentioned that we will hard and soft values for each limit. My assumption is that a job submitted above a soft limit but under the hard limit would be held (e.g. perhaps allowing time-based limits) and above the hard limit, jobs would be rejected. I'm not sure if this is a near term requirement thought. Perhaps @trws and @ryanday36 could chime in.

I think that what @trws mentioned in the meeting chat is that we currently specify a default time that is different (much lower, generally) than the max time. The theory is more or less that having a default time that is too short for users to do anything useful encourages them to actually think about how long their job will run for and choose an appropriate time. I'm not sure how much I buy that theory, but it's what we currently do. You could also force users to think about time by just rejecting jobs that don't specify a time limit. If we do that, or even if we don't, it might be nice to allow users to just specify something like `-t MAX` to indicate that they just want whatever the max time allowed by the queue/partition/whatever.

grondo:
Thanks @ryanday36. Just a quick follow up question -- is there any near term use case for limits that change during the course of a day/week? If not, then maybe we can drop the requirement for hard vs soft limits for job size for now.

ryanday36:
> Thanks @ryanday36. Just a quick follow up question -- is there any near term use case for limits that change during the course of a day/week? If not, then maybe we can drop the requirement for hard vs soft limits for job size for now.

That's definitely a nice-to-have, not a need-to-have. It's something that we'd like to do with things like pdebug, but not something that we currently do.

dongahn:
@garlick: I like the straw man proposal:

```
  # scheduler-specific, per-queue configuration (opaque to other subsytems)
  scheduler-options = {
    queue-policy = "hybrid"
  }
```

Fluxion currently has two TOML configuration files: 

- https://flux-framework.readthedocs.io/projects/flux-sched/en/latest/man5/flux-config-sched-fluxion-qmanager.html
- https://flux-framework.readthedocs.io/projects/flux-sched/en/latest/man5/flux-config-sched-fluxion-resource.html

It would be great if all of the option sets for `qmanager` can be configured this way as well! This will also solve https://github.com/flux-framework/flux-sched/issues/650.

As the `sched-fluxion-resource` configuration, Fluxion currently don't have a great way to change match policies (and other options) per different partitions  just yet. So you won't be able to specify things through per-queue scheduler options much . @grondo and I have a ticket to discuss this, but per-queue partition-aware match policy still need to be worked out.  Just FYI

grondo:
Would it make sense to allow some top level keys in the `[queues]` table to allow configuration of defaults, inherited by all configured queues, but applied even when there is a single anonymous queue (the common case for anything but system instance?)

e.g.  for Fluxion use case

```toml
[queues]

scheduler-options = {
    queue-policy = "hybrid"
}
```

dongahn:
>Would it make sense to allow some top level keys in the [queues] table to allow configuration of defaults, inherited by all configured queues, but applied even when there is a single anonymous queue (the common case for anything but system instance?)

> e.g. for Fluxion use case

Yes that sounds perfect! That way we don't have repeats defaults for each queue. Fluxion qmanager config already knows how to handle this -- great thought.

trws:
Yeah, I was mainly referring to having a default time.  IIRC we frequently have a default of 30-60m on queues of max limits of even 8 hours.  That way smaller one-off jobs don't get penalized by defaulting to the max time.

grondo:
Thanks @trws, @ryanday36. I apologize then for getting off in the weeds with my comments about job size hard/soft limits :facepalm: (I was thinking jobs could be held above the soft limit, allowing a configuration change or administrator to release them at a later time.) Sounds like this isn't needed for the near term so let's ignore it for now!

garlick:
I like the idea of a general queue config that applies to the anon queue and is inherited (and overridden) by any explicitly defined queues. As @ryanday36 mentioned in our meeting, some of our production systems will just have one queue that could be the anon one, so we might want to be able to specify limits and defaults that apply to the anonymous queue too, not just scheduler options.

Should we maybe have top level tables for `[limits]`,  `[jobspec-defaults]`, and `[scheduler-options]`  that define the anonymous/inherited values and say that all keys in  `[queues]` are queue names, and let the top level stuff be overridden in each queue table?  Like

```toml
[jobspec-defaults]
system.duration = "1m"
system.queue = "pbatch"

[limits]
job-size = { nnodes = 24, ncores = 384, ngpus = 48 }

[scheduler-options]
queue-policy = "hybrid"

[queues]
  [queues.pbatch]
  requires = [ "pbatch" ]
  limits.job-size = { nnodes = 24, ncores = 384, ngpus = 48 }

  [queues.pdebug]
  requires = [ "pdebug" ]
  limits.job-size = { nnodes = 8, ncores = 128, ngpus = 16 }
```
Or maybe those three top level tables could be grouped in a `[policy]` or similar table?

grondo:
Yeah, I like the direction that is headed!

I do slightly prefer grouping them under another table like `[policy]` as you suggest. Just thinking of naming convention, I like the idea of `[policy.jobspec]` `[policy.scheduler]` and `[policy.limits]` tables. Where there could be a `[policy.jobspec.defaults]` table (just as easy to type but perhaps more flexible than just one `[jobspec-defaults]` table, and everything under `[policy.scheduler]` is opaque scheduler policy. (i.e. we can drop the `-options` as unnecessary). I realize those are trivial comments, but considering the possibility of setting/overriding these configuration values on the command line one day, saving characters might be nice.

trws:
I think I agree with @grondo, though if they're specifically queue limits possibly `policy.queues.limits` or `queues.policy`/`queues.default` or similar, keep the scopes consistent.

garlick:
So something like this?

If the structure of `policy` is mirrored within a queue definition, then we can just parse the top level object, then the queue specific object on top of it.
```toml
[policy]
  [policy.jobspec.defaults]
  system.duration = "1m"
  system.queue = "pbatch"

  [policy.limits]
  job-size = { nnodes = 24, ncores = 384, ngpus = 48 }
  duration = "24h"

  [policy.scheduler]
  queue-policy = "hybrid"

[queues]
  [queues.pbatch]
  requires = [ "pbatch" ]
  policy.limits.job-size = { nnodes = 24, min-nnodes = 8, ncores = 384, ngpus = 48 }
  policy.limits.duration = "24h"
  policy.jobspec.defaults.duration = "30m"

  [queues.pdebug]
  requires = [ "pdebug" ]
  policy.limits.job-size = { nnodes = 8, ncores = 128, ngpus = 16 }
  policy.limits.duration = "30m"
  policy.jobspec.defaults.duration = "5m"
```

grondo:
That seems pretty nice to me!

Seems like you could also support `policy.jobspec.default.system.constraints.properties = [ "pbatch" ]` but, my, that is long so I like the `requires` shorthand. :-)

garlick:
Yeah, anyway, we might need slightly different semantics for `requires` vs `jobspec.default`, since presumably we would add to, rather than replace, any constraint the user puts in their jobspec at submission.

garlick:
I made another pass on this draft, trying to capture
* one anon queue remains the default, unless named queues are configured
* global and per queue policy for limits, jobspec defaults, access control, and scheduler
* jobspec update sematics (posting a `jobspec-update` event, which needs to be added to RFC 21)

spec_33.rst
```diff
@@ -0,0 +1,265 @@
...
+``queue`` jobspec system attribute (RFC 14).  If a queue was not explicitly
+named in the jobspec, and a default queue is defined, the queue SHALL be
+assigned by the job manager during the initial request validation.
+
+Request Validation
+~~~~~~~~~~~~~~~~~~
+
+A job request SHALL be rejected on submission if it names an unknown queue,
+or if it is possible to determine that the job would exceed limits or violate
+access policy of the assigned queue.
```
cmoussa1, line 230: Should a job request also be rejected on submission if it names a queue that the user/bank association does not have access to? For example, I have a user/bank association that has access to queues `A,B` but not `C`, and I try to submit a job in queue `C`.

I'm sorry if this is out of the scope of this RFC, since this will probably involve information to be shared from the accounting system.

cmoussa1 (commented):
I like the direction that this RFC is headed in @garlick. I gave this a read and from my POV from the flux-accounting side, I just had one minor question/clarification. Thank you for continuing to lead this effort!

spec_33.rst
```diff
@@ -0,0 +1,265 @@
...
+^^^^^^^^^^^^^
+
+The instance owner MAY configure queue access controls in the form of
+allow/deny statements that refer to:
+
+- UNIX users
+
+- UNIX groups
+
+- bank names (as defined by accounting plugin)
```
grondo, line 193: Perhaps if instead of explicitly listing bank names here, we can state something like:

 - The access policy MAY also be deferred to an external entity through the use of one or more jobtap plugins. For example, the flux-accounting multi-factor priority plugin may control access to queues based on the user and bank information from the accounting database.

This could perhaps address @cmoussa1's earlier comment.

spec_33.rst
grondo: My reading is that this would fall under "violates access policy of the assigned queue".

Something to note: since jobtap plugins may be validating a job based on queue name, we have to ensure that any queue name is assigned before `job.validate` callbacks are run.

grondo (approved):
This LGTM! Just a couple notes inline. But I like that this turned into a very easy to follow RFC!

spec_33.rst
garlick: Oh good idea - thanks.

cmoussa1: I like this, and I think it greatly addresses my comment above. Thank you!

garlick:
I pushed some more changes based on @cmoussa1 and @grondo's feedback.  Thanks!

grondo:
Great! This seems like it is probably at a good point to merge then?

garlick:
It could be merged.  Shall we?

One point is that I left in place `Administrative Tools` section which implies that queues must be implemented in the job manager.  We have a path forward that stops short of that, but I felt it was good to have the full plan written down, especially since it queue control will be required to meet one of the use cases - submitting to a DAT early.  The `Priority Policy` section is also still just TBD.

I think it's probably OK to merge as is but wanted to point those things out in case they were missed.

garlick:
Thanks, setting MWP

PR #336, opened 2022-06-09
rfc21: allow SCHED->PRIORITY on urgency event

garlick:
Problem: the priority must be recalculated if the job urgency
chages, but since the job is already past PRIORITY state, this
requires special case handling.

In addition, the priority plugin may take a while to calculate
the new priority and this will not be visible to users.  It would
be better to transition back to the PRIORITY state.

Update state diagram to allow for SCHED->PRIORITY transition
on the urgency event.

See also flux-framework/flux-core#4364


data/spec_21/states.dot
```diff
@@ -24,6 +24,7 @@ digraph states {
     CLEANUP -> INACTIVE;
 
     SCHED -> PRIORITY [label="flux-restart"]
+    SCHED -> PRIORITY [label="urgency"]
```
grondo, line 27: Commit typo: "chages"

grondo (approved):
LGTM! Just one typo noted inline.

data/spec_21/states.dot
garlick: Doh!

garlick:
Fixed and setting mwp.  Thanks!

PR #337, opened 2022-06-13
rfc21: transition NEW->DEPEND on "validate" event

garlick:
Problem: the current state machine transitions jobs from NEW to DEPEND on the submit (first) event, which leaves no opportunity for events to be posted before the job is validated, as may be needed for assigning default values to empty jobspec attributes such as queue or duration.
    
Change the state machine to transition from NEW to DEPEND on a new "valid" event.  Or if the job is invalid, an "invalid" event signifies that the job should immediately be removed.

There are also a couple of commits here to repair small problems in the state diagram.

See also flux-framework/flux-core#4366.


garlick:
Renamed events to "validate" and "invalidate" per @grondo's suggestion in flux-framework/flux-core#4366.

grondo (approved):
This LGTM!

garlick:
Thanks!  Setting MWP.

PR #338, opened 2022-07-05
RFC21: add jobspec-update event

garlick:
Problem: RFC 33 refers to the `jobspec-update` event in RFC 21, but it does not appear there.

Add the event to RFC 21.


garlick:
Thanks!

PR #339, opened 2022-07-11
rfc21: add version to submit event

garlick:
Problem: the job eventlog format recently changed with the addition
of the validate event, and this made it difficult to detect and
handle eventlogs from earlier versions when restarting after a
Flux upgrade.

Add a version number to the submit event that can be used to
distinguish this and future changes to the eventlog format.

See also flux-framework/flux-core#4399

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -156,8 +156,8 @@ used to generate Flux job IDs.
 This specification defines the version 1 format of the resource-set
 representation or *R* in short.
 
-:doc:`21/Job States and Events <spec_21>`
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


chu11:
do we need to document what is version 0?  Or shall it be "anything goes"?  Or possible references an earlier version of this RFC?

grondo:
The way I read it, the version is optional, so there is no version 0, just formats that "predate versioning".

garlick:
It may be clearer if I put "version 1" in the RFC title and _require_ the version key.

Then the workaround for supporting pre-versioned eventlogs can be documented as such rather than calling it "version 0" which has no specification.

grondo (approved):
Forgot to approve!

PR #340, opened 2022-08-12
rfc19: fix English error / typo

chu11:
Problem: Lower case f was listed twice:

"lowercase f ``f``"

Solution: list the lower case f just once!


grondo (approved):
Good catch!

PR #341, opened 2022-09-10
rfc33: update after implementing limits and jobspec defaults

garlick:
Problem:  The design for applying jobspec defaults and implementing limits has evolved somewhat since the RFC was written.  In addition, some notes were placed there during discussions that were thought would be ongoing, but which did not go anywhere.

Make a pass through RFC 33 to get it in sync with current plans.

### DIFF
--- a/spec_10.rst
+++ b/spec_10.rst
@@ -200,7 +200,7 @@ garbage collection.
 References
 ----------
 
--  `Perkeep lets you permanently keep your stuff, for life. <https://perkeep.org/>`__.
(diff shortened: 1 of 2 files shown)


garlick:
Linkcheck is failing on the RFC 10 https://perkeep.org/ link that works for me from home, but apparently not in CI right now.   Maybe it'll start working if we wait a bit.

grondo (approved):
Sorry about missing this one. LGTM!

PR #342, opened 2022-09-19
rfc33: simplify queue.NAME.requires

garlick:
Problem: RFC 33 specifies that queue.NAME.requires may refer to advanced RFC 31 constraint specifications, but these are not yet implemented and may be confusing for admins to configure.

Re-define queue.NAME.requires as an array of property constraint names.


grondo (approved):
Oops, missed this one! LGTM.

garlick:
Thanks!

Issue #344, opened 2022-10-23
finish documentation of Flux wire protocol


garlick:
A slack discussion reminded me that we have gaps in our documentation of Flux's "wire protocol".  By that I mean how Flux messages are sent and received directly on a socket like the "local connector" (a UNIX domain socket)

What we have:

* [RFC 3](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_3.html) defines the Flux message format in terms of multi-frame messages (using zeromq terminology)
* [RFC 6](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_6.html) defines Flux RPCs in terms of Flux request and response messages

Doc gaps:

* [message encoding](https://github.com/flux-framework/flux-core/blob/master/src/common/libflux/message.c#L223), which used to be outsourced to zeromq
* [message framing for the local connector](https://github.com/flux-framework/flux-core/blob/master/src/common/librouter/sendfd.c), which has been stable for a long time but was never documented.

Ideally it should be possible to implement code for sending Flux RPCs wtih arbitrary (JSON or raw) payloads wtihout using libflux-core.so, in case someone wants to do a native implementation in some language instead of libflux-core bindings.  That might be useful, for example, when libflux-core's reactive programming model is not a good fit for a language's concurrency model.

vsoch:
Ping @alecbcs!

PR #345, opened 2022-10-23
RFC3: document message part framing, and message framing on UNIX stream socket

garlick:
As noted in #344, Flux has taken over the framing of message parts from ZeroMQ, but we neglected to document that.  Also, we never documented how full messages are framed on a UNIX stream socket by the `local` connector.

This fills in those missing pieces.

RFC 3 was written a while ago and when I read it with my "write a new message encoder/transport" hat on, it feels it doesn't have the right structure.  I'd like to improve that, but for now just wanted to put in the missing stuff.


vsoch:
I was going to try playing around with this today (and got totally consumed with a personal project!) I will definitely play around with this soon.

garlick:
Cool! I was trying to remember if there is anything I missed.   Like it seems like we had a hack for the local connector where a message was sent from server to client on the initial connection to provide the instance owner userid or something, but I'm not seeing that in a message trace at least, e.g.
```
garlick@picl0:~$ FLUX_HANDLE_TRACE=1 flux getattr rank
--------------------------------------
> 0.00000
> attr.get
> flags=topic,payload,route userid=unknown rolemask=none nodeid=any matchtag=1
> {"name":"rank"}
--------------------------------------
< 0.00130
< attr.get
< flags=topic,payload,route userid=500 rolemask=owner errnum=0 matchtag=1
< {"value":"0","flags":1}
0
```
Deeper spelunking required than i had time for today...

garlick:
> it seems like we had a hack for the local connector where a message was sent from server to client on the initial connection to provide the instance owner userid or something

For the record, I confirmed that is ancient history, no longer a thing.

spec_3.rst
```diff
@@ -215,29 +215,54 @@ established for common payload types:
 5. JSON payloads SHALL include a terminating NULL character.
 
 
-General Message Format
-~~~~~~~~~~~~~~~~~~~~~~
+Message Structure
+~~~~~~~~~~~~~~~~~
```
chu11, line 219: perhaps an example or two might be a good idea here?

spec_3.rst
```diff
@@ -341,6 +341,19 @@ Flux messages are specified by the following modified ABNF grammar [#f2]_
    ; unused 4-byte field
    unused          = %x00.00.00.00
 
+
+Message Framing
+~~~~~~~~~~~~~~~
+
+When Flux uses ZeroMQ as transport (``tcp``, ``ipc``, and ``shmem``
+connectors), ZeroMQ handles the message framing.
```
chu11, line 349: "as transport" doesn't sound right.  "as a transport"?  "for transport"?

spec_3.rst
garlick: Agreed.  Just pushed a couple of examples.  Helpful?

garlick: oops force pushed again to fix duplicate captions in the image, and add some circles and arrows to further explain.

chu11: should the blue column prefix with `0x`?  Took me awhile to realize it wasn't decimal :-)

garlick: I was tryiing to be consistent throughout and 0x would be kind of ugly - maybe if I just incorporate the footnote into the caption it'll help.  I'll push that - hang on...

chu11 (approved):
LGTM

garlick:
Thanks! I'll set MWP.

PR #348, opened 2022-10-24
rfc3: describe authorization handshake

### LINKED ISSUE #347: rfc 3: document unix domain socket "auth handshake"
I _thought_ remembered there was a little handshake on the local connector that occurs before messages begin flying.  Contrary to what I concluded in #344, it's actually [this](https://github.com/flux-framework/flux-core/blob/master/src/common/librouter/usock.c#L714).

We need to document that, probably along with the local connector framing that was added to RFC 3.

garlick:
Problem: when a client communicates with Flux over a UNIX domain socket, a single byte is returned upon connection that indicates whether the connection is authorized.  This is not described in any RFC.

Rework the Message Framing section as Message Framing and Security, and describe this initial handshake.

Fixes #347

### DIFF
--- a/spec_3.rst
+++ b/spec_3.rst
@@ -350,17 +350,27 @@ Flux messages are specified by the following modified ABNF grammar [#f2]_
    unused          = %x00.00.00.00
 
 
-Message Framing
-~~~~~~~~~~~~~~~
+Message Framing and Security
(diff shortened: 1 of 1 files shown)


chu11 (approved):
LGTM

PR #350, opened 2022-10-25
rfc3: fix message framing misinformation

### LINKED ISSUE #349: rfc3: message part framing only applies to unix domain socket transport
Problem: RFC 3 states that message parts are framed within a message with length fields, but that is only the case when messages are sent via a UNIX domain stream socket.

I had forgotten that we still break up the message into 0MQ "frames" when using 0MQ as the message transport, becuase this is required to be compatible with the native message routing protocols and pub/sub topic strings.  The code that does this is [here](https://github.com/flux-framework/flux-core/blob/master/src/common/libzmqutil/msg_zsock.c#L28).

Sorry it has been a while since I worked on this stuff and it seems to be coming back to me gradually when I'm working on other things!

garlick:
Problem: zeromq handles framing of message parts when it is performing message transport, so the discussion of message part framing must be combined with the section on message framing when a UNIX domain stream socket is used for message transport.

Reorganize the Message Structure and Message Framing and Security sections, and split the figure into one with all the framing fields, and one with none.

Fixes #349


garlick:
@chu11 any chance you could give this a quick once over?

grondo (approved):
LGTM!

garlick:
Thanks!

PR #351, opened 2022-10-26
rfc13: elaborate on PMI_process_mapping

garlick:
Problem: There is no PMI_process_mapping example for "cyclic" task distribution, and the fact that rank to nodeid assignment can be determined is not entirely clear from the prose (but obvious given the goal of finding "cliques" for shmem communication).

Mention that the rank to nodeid assignment can be determined from the mapping.

Add an example of cyclic task distribution, and update formatting and descriptions of the other examples for clarity.


grondo (approved):
Looks great, thanks!

PR #352, opened 2022-11-01
rfc3: rework the rest of the document

garlick:
Problem: RFC 3 was written a long time ago and now reads...poorly.

After adding some missing bits last week, this is an attempt to spruce up the document overall.   It's going to be hard to review at the diff level so maybe a read-through of the rendered preview is the way to go.  However I'll mark it as a WIP for now and plan to come back to it myself after a break before asking for a review by others.


garlick:
I made another pass on this and I think it's in decent shape and ready for a review.

spec_3.rst
```diff
@@ -2,249 +2,325 @@
...
+   The Flux overlay network interconnects Flux brokers with root at rank 0.
 
-Event messages are published such that they are available to subscribers
-throughout the instance. Events are published with a *topic string*
-attached. Subscribers register a list of topic string prefixes
-to filter the set of messages they receive.
+The overlay network is capable of routing messages over the shortest path
+between any two brokers, of "smart host" routing where messages are forwarded
+upstream until a more informed broker knows how to route it, and of
+multi-casting messages to all broker ranks.  These capabilities support remote
```
grondo, line 94: This is fine, though a little difficult to read. No need to change it if it reads well to you, but I'd perhaps split into multiple sentences:
> The overlay network is capable of routing messages using several methods. Messages may be routed over the shortest path between any two brokers, using "smart host" routing, where messages are forwarded upstream until a more informed broker knows how to route it, or by multi-casting to all broker ranks.

Maybe that is only a slight improvement if anything at all.

spec_3.rst
```diff
@@ -60,32 +60,49 @@ the following specific goals:
...
+only between tree levels, and each node has at most one parent.  The tree
+shape, the instance size, and each broker's rank are fixed once the instance
+begins running.
+
+.. figure:: images/tree.png
+   :width: 150
+   :alt: Flux overlay network topology
+   :align: center
 
+   The Flux overlay network interconnects Flux brokers with root at rank 0.
```
chu11, line 89: Perhaps put a "figure 1" or "Example" or something with this footnote.  The way it displays in the web browser, it looks like its part of the normal text of a paragraph.

spec_3.rst
```diff
@@ -60,32 +60,49 @@ the following specific goals:
...
+
+The overlay network is capable of routing messages over the shortest path
+between any two brokers, of "smart host" routing where messages are forwarded
+upstream until a more informed broker knows how to route it, and of
+multi-casting messages to all broker ranks.  These capabilities support remote
+procedure call and publish-subscribe, the two main communication idioms used
```
chu11, line 95: perhaps say "remote procedure call (RPC)", so the "RPC" in the next paragraph has a definition.

spec_3.rst
```diff
@@ -60,32 +60,49 @@ the following specific goals:
...
+   :width: 150
+   :alt: Flux overlay network topology
+   :align: center
 
+   The Flux overlay network interconnects Flux brokers with root at rank 0.
+
+The overlay network is capable of routing messages over the shortest path
+between any two brokers, of "smart host" routing where messages are forwarded
+upstream until a more informed broker knows how to route it, and of
+multi-casting messages to all broker ranks.  These capabilities support remote
```
chu11, line 94: sentence feels a little run on, split into two?  Like "between any two brokers.  Using "smart host" routing, messages are forwarded ...."

spec_3.rst
```diff
@@ -2,249 +2,325 @@
...
+forced to restart.  In a Flux system instance, this restart is managed by
+systemd.
 
-Responses are optional replies to requests. They follow the ZeroMQ
-ROUTER-DEALER message flow, which unwinds the source address route
-accumulated by the request, and uses them to select among peers at each hop.
+Flux messages share a common structure that is strongly influenced by ZeroMQ
+conventions, since ZeroMQ DEALER-ROUTER sockets are used for transport between
+brokers and ZeroMQ PAIR sockets are used between a broker and its dynamically
+loaded modules, and Flux messages must be conformant.  Flux messages may be
```
grondo, line 108: It is unclear what "and Flux messages must be conformant" implies here. Is it supposed to be "and Flux messages must _therefore_ be conformant", or maybe because there is already a "since"  the "and" should just be dropped? Maybe it is supposed to be "...must be conformant to enable these ZeroMQ mechanisms" (Sorry, it is also possible I'm just being dense)

spec_3.rst
```diff
@@ -2,249 +2,325 @@
...
-and SHALL NOT appear in the nodeid slot of an encoded message.
+- Messages SHALL support a *topic string* message part for subscriber
+  filtering.
 
-A node’s rank SHALL be assigned at broker startup and SHALL NOT change
-for the node’s lifetime.
+Message parts SHALL be preserved by message transports;  that is, Flux messages
+are sent and received on the transports as an array of parts.  Message
+transports MAY modify Flux messages if directed to do so.  For example,
+ZeroMQ ROUTER sockets determine the next hop by consuming a message part.
```
grondo, line 138: Does it make sense to say "Message parts SHALL be preserved by message transports" while also saying "Message transports MAY ... consume a message part"?

spec_3.rst
```diff
@@ -2,249 +2,325 @@
...
-Service Addressing
-~~~~~~~~~~~~~~~~~~
+Required Message Parts
+^^^^^^^^^^^^^^^^^^^^^^
 
-A Flux service SHALL be identified in a request by a *topic string*,
-a set of words delimited by periods, in which the first word identifies
-the service, and remaining words represent *methods* within that service.
-For example, "kvs.get" refers to the *get* method of the *kvs* service.
+Flux message are REQUIRED to have one message part that acts as a protocol
```
grondo, line 169:
```suggestion
Flux messages are REQUIRED to have one message part that acts as a protocol
```

spec_3.rst
```diff
@@ -113,170 +113,214 @@ There are four distinct Flux message types:  *request* and *response* messages
...
-Service Addressing
-==================
+Required Message Parts
+^^^^^^^^^^^^^^^^^^^^^^
 
-A Flux service SHALL be identified in a request by a *topic string*,
-a set of words delimited by periods, in which the first word identifies
-the service, and remaining words represent *methods* within that service.
-For example, "kvs.get" refers to the *get* method of the *kvs* service.
+Flux message are REQUIRED to have one message part that acts as a protocol
```
chu11, line 169: message -> messages

grondo (approved):
This looks great! Just a few notes and questions inline.

spec_3.rst
garlick: I like that, thanks.

chu11 (commented):
noticed one extra typo while I was reading but not sure its from your changes, maybe want to fix up while you're at it?  Look for "as as a size" in the message framing & security section.

Also in your last figure, perhaps want to add similar thing to footnote like the other diagram, something like "Integer values are in hex, and for clarity are not converted to network byte order."

spec_3.rst
garlick: No it's a good point.  Does this read any better?

> Flux messages share a common structure that is strongly influenced by ZeroMQ
conventions, since ZeroMQ provides a transport for Flux messages, and certain
ZeroMQ socket types impose structural requirements on messages for routing
and subscription filtering.  Flux messages may be sent over other transports,
however.  For example, regular UNIX domain stream sockets transport messages
between local processes and Flux brokers.

spec_3.rst
garlick: That bugged me too.  How about this?
> The boundary between message parts SHALL be preserved by message transports;
that is, Flux messages sent as an array of parts MUST be received as an array
of parts, not a concatenated blob.
> 
> Message transports MAY modify Flux messages if directed to do so.  For
example, a ZeroMQ ROUTER socket implements source-address routing by adding
a message part in one direction and removing one in the opposite direction.

spec_3.rst
garlick: Ah yeah, I can maybe just shorten the caption and since it is centered and italicized, it will look more like it belongs with the figure.

spec_3.rst
grondo: Yes, sounds great!

spec_3.rst
grondo: Yes, that helps me! Thanks!

spec_3.rst
garlick: @grondo proposed a rewording above that I liked.  What do you think?

chu11: yup, @grondo's looks good.

garlick:
Great suggestions!  I just force pushed those.

garlick:
Thanks!  I'll set MWP.

PR #353, opened 2022-11-03
rfc13: fix PMI_process_mapping example + improve formatting

garlick:
Problem: one of the PMI_process_mapping examples recently expanded has the wrong rank lists.    Also, this document  is egregiously malformatted.

Fix the example and spruce up the formatting a bit.

### DIFF
--- a/spec_13.rst
+++ b/spec_13.rst
@@ -109,17 +109,20 @@ library providing the PMI-1 API implemented using a proprietary protocol.
 Terminology
 -----------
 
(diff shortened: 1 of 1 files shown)


garlick:
The process mapping examples look better in the form presented in RFC 34 (#354) so I switched to that form for the PMI_process_mapping examples presented here.  Also, tweaked the examples to fit in the table columns and to increase signal to noise.

grondo (approved):
LGTM!

PR #354, opened 2022-11-05
rfc34: add flux task map RFC

garlick:
Problem: Flux needs a task map format that scales for both block and cyclic task distributions.

Define a new format in rfc34.

### DIFF
--- a/spec_34.rst
+++ b/spec_34.rst
@@ -1,0 +1,212 @@
+.. github display
+   GitHub is NOT the preferred viewer for this file. Please visit
+   https://flux-framework.rtfd.io/projects/flux-rfc/en/latest/spec_34.html
+
+################
+34/Flux Task Map
+################
+
+The Flux Task Map is a compact mapping between job task ranks and node IDs.
+
+- Name: github.com/flux-framework/rfc/spec_34.rst
+
+- Editor: Jim Garlick <garlick@llnl.gov>
+
+- State: raw
+
+********
+Language
+********
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to
+be interpreted as described in `RFC 2119 <https://tools.ietf.org/html/rfc2119>`__.
+
+*****************
+Related Standards
+*****************
+
+- :doc:`13/Simple Process Manager Interface v1 <spec_13>`
+
+- :doc:`20/Resource Set Specification <spec_20>`
+
+- :doc:`22/Idset String Representation <spec_22>`
+
+**********
+Background
+**********
+
+The task map communicates how a parallel program's tasks are assigned to the
+allocated nodes. Given a node ID, the task map can provide a set of ranks.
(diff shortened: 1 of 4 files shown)


spec_34.rst
```diff
@@ -0,0 +1,121 @@
...
+  [[0,4,4,1]]
+  [[0,4,1,4]]
+
+Here are block and cyclic distributions of 1M tasks over 4K nodes using the
+Flux task map representation:
+
+.. code:: json
+
+  [[0,4096,256,1]]
+  [[0,4096,1,256]]
```
grondo, line 121: Should the last one be:
```json
[[0,4096,1,4096]]
```

grondo: It could also be instructive to give one non-uniform example as in [RFC 13](https://flux-framework--353.org.readthedocs.build/projects/flux-rfc/en/353/spec_13.html#id18), e.g. ``[[0,8,16,1],[8,4,32,1]]``

grondo (approved):
Awesome! Just one question inline.

In the prototype I had made the _repeat_ parameter optional (defaults to 1), but I can see here that just unnecessarily complicates the implementation. Thanks!

spec_34.rst
grondo: Eh, nevermind I think looking back at this _repeat_ of 256 is correct. Sorry!

spec_34.rst
```diff
@@ -0,0 +1,210 @@
...
+This mapping is compact for *block* task distributions, where blocks of
+contiguous task ranks are assigned to nodes in ascending order.  Its
+scalability breaks down for *cyclic* task distributions, where one or more
+task ranks are assigned to nodes in round-robin order. As an example, a PMI
+task map for 1M tasks distributed over 4K nodes in block distribution is
+compact as shown above, but the same job with a cyclic distribution (stride
+of 1) is a string of 2824 characters.
+
+*****
+Goals
```
grondo, line 98: It feels like it would read better if Goals were above the Background section, so the reader can have those goals in mind while reading about the PMI_process_mapping existing standard and other related background. However, I also realize maybe it is idiomatic to put Goals after Background, in that case it is fine. (Another suggestion would be to move the discussion of the naive/raw task map format and PMI_process_mapping format to an "Existing Implementations" or similar section after Goals.

grondo (approved):
LGTM! Thanks. Just a couple trivial questions inline.

spec_34.rst
garlick: Ooh, good idea.  Let's try
- background
- goals
- existing implementations

spec_34.rst
```diff
@@ -0,0 +1,210 @@
...
+     - (vector,(0,6,1),(0,6,1),(4,2,1),(4,2,1))
+     - (vector,(0,6,2),(4,2,2))
+   * - 4096*256
+     - (vector,(0,4096,256))
+     - *(long string)*
+     - *(long string)*
+
+.. note::
+  The cyclic:N distribution for N > 1 is equivalent to Slurm's *plane*
+  distribution.
```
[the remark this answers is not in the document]
grondo: Oh, I somehow missed that on my first read, sorry! The note should stay then :wink:

PR #355, opened 2022-11-09
rfc34: fix test vector

grondo:
Problem: A raw test vector has some commas where semicolons should be.

Fix the typo in the test vector.


garlick (approved):
Oh good catch.  Sorry for the error!

grondo:
No problem, the testsuite caught it! I can barely tell the difference between a comma and a semicolon, so I was going cross-eyed looking at that one.

vsoch:
_c++ cackles maniacally from the shadows…_ 👹

garlick:
Maybe we should have defined the delimiter between nodes as a :carrot: or something that stands out!

vsoch:
_bunny rabbit laughs maniacally from the shadows..._ :carrot: :rabbit:

PR #357, opened 2022-11-30
rfc34: allow empty and document unknown task maps

grondo:
RFC 34 does not indicate how to specify an unknown task mapping, but indication of an unknown mapping is required when, for example, an encoded mapping is too large to distribute or the mapping is otherwise not known.

This PR documents that an empty string is an indicator of an "uknonwn" mapping for `PMI_process_mapping`, as is already documented in RFC 13, and also makes an empty string the official encoding of an unknown mapping in the Flux Task Map format.

Additionally, an "empty" mapping is also allowed by changing the task map requirement from "one or more blocks" to "zero or more". (This is probably only useful in testing, and also is the encoded output from a libtaskmap encoded result from `taskmap_create()` with no further initialization)

### DIFF
--- a/spec_34.rst
+++ b/spec_34.rst
@@ -113,7 +113,10 @@ Implementation
 **************
 
 The Flux task map SHALL be represented as a JSON array to avoid the need
-for a custom parser.  The array MUST contain one or more *map blocks*.
+for a custom parser.  The array MUST contain zero or more *map blocks*.
+
+A Flux task map that contains zero map blocks SHALL indicate that the task
+mapping is unknown.
 
 A Flux task map block is a JSON array with four REQUIRED integer array
 elements:
@@ -182,6 +185,8 @@ Test Vectors
 
    * - raw task map
      - Flux task map
+   * - mapping unknown
+     - []
    * - 0
      - [[0,1,1,1]]
    * - 0;1


grondo:
After some offline discussion with @garlick, this PR has been modified to treat all "empty" task maps as indicating that the mapping is unknown, i.e. both `"[]"` and `""` indicate an unknown mapping.

spec_34.rst
```diff
@@ -100,6 +100,10 @@ of (*nodeid*, *nnodes*, *ppn*).  Some examples are:
   The cyclic:N distribution for N > 1 is equivalent to Slurm's *plane*
   distribution.
 
+RFC 13 also specifies that an empty string is a valid ``PMI_process_mapping``,
+and indicates that the mapping was too large to fit in a PMI KVS value or
+is otherwise unknown.
```
garlick, line 105: It doesn't actually specify that the empty string is valid.  The quote is:

> If the process mapping value is too long to fit in a KVS value, the process manager SHALL return a value consisting of an empty string, indicating that the mapping is unknown.

Maybe better wording would be:
> RFC 13 specifies that process managers must set  ``PMI_process_mapping`` to the empty string to indicate that the mapping was too large to fit in a PMI KVS value, so users should interpret the empty string as "mapping unknown".

spec_34.rst
```diff
@@ -113,7 +117,12 @@ Implementation
 **************
 
 The Flux task map SHALL be represented as a JSON array to avoid the need
-for a custom parser.  The array MUST contain one or more *map blocks*.
+for a custom parser.  The array MUST contain zero or more *map blocks*.
+As a special case an empty string SHALL be treated the same as an zero
+length JSON array, i.e. ``[]``.
```
garlick, line 122: Do we need to define this special case?  The empty string can be treated as valid pmi encoding of "unknown" without defining the empty string as a valid RFC 34 mapping.

spec_34.rst
```diff
@@ -113,7 +117,12 @@ Implementation
...
+
+A Flux task map that is Empty SHALL indicate that the task mapping is
+unknown.
```
garlick, line 125: How about
s/is Empty/contains zero map blocks/

garlick (approved):
Sounds good.  Does it make sense to add `[]` as a test vector?

spec_34.rst
grondo: Ok, I didn't pick up on that semantic subtlety (if an empty string is an invalid mapping, perhaps decoding an empty string should return an error, which should then be treated as "mapping uknown"?)

spec_34.rst
grondo: I guess we don't need this, but this means that a decoder can't tell the difference between an invalid RFC 34 task map or a valid PMI_process_mapping (although I guess you are saying above that the empty string is not a valid PMI_process_mapping). It just seemed easier to make an empty string a valid task map. However, in practice the code in `taskmap_decode()` would be the same, so if the ambiguity doesn't bother you then I'm fine with dropping this.

spec_34.rst
grondo: This seems right if we drop the treatment of an empty string as empty

spec_34.rst
garlick: I think it's OK if we treat the empty string as a valid pmi encoding.  I just didn't want this rfc to state that _that_ rfc said something it didn't (not really).

spec_34.rst
garlick: All valid PMI_process_mappings are invalid RFC 34 task maps, so I think it's OK if one more (the empty string) is treated as valid.  Or did we need to catch that particular case for other reasons?

garlick:
In the RFC could we just say  `mapping unknown` in the left column?

spec_34.rst
grondo: No, the problem is that since you wanted `taskmap_decode()` to also decode valid PMI_process_mapping strings via a heuristic, it seemed like it wasn't possible to determine if `taskmap_decode()` should generate an error on an empty string or not, since the function doesn't know the context from which it is being called. If the string is coming from a context where an RFC 34 task map should be provided, then an empty string should cause an error if an empty string is not a valid taskmap. However, if the string is supposed to be a PMI_process_mapping, then the empty string should be encoded as an "unknown" task map. It seemed reasonable to add a heuristic to `taskmap_decode()` to check for `vector,` and attempt to automatically decode as a PMI_process_mapping, but doing the same for an empty string seemed like going a bit too far.

It seemed to me like that all got cleaned up a bit if an empty string was treated the same in both contexts. However, if having the same special case in RFC 34 as in PMI_process_mapping bothers you, it isn't critical IMO so I can make this change. In the code, any string with `vector,` or an empty string can be assumed to be PMI_process_mapping, so it is effectively the same end result.

spec_34.rst
grondo: Ah, ok, that makes sense.

grondo:
Ok, I've made the suggested changes (I think). I just dropped the addition to the `PMI_process_mapping` section since it was probably only adding unnecessary confusion and duplicating what is already in RFC 13 anyway.

garlick (approved):
This works for me if it works for you.   If this change complicates the implementation and returning to a separate pmi decode function makes sense, I'm fine with that.

grondo:
It doesn't complicate the code, it is just slightly ambiguous that an empty string is not a valid task map, but it will decode without error and will be treated the same as `[]` in `taskmap_decode()`. However, it doesn't bother me too much, and it does simplify the RFC so I'm fine merging this as is.

PR #360, opened 2023-01-18
rfc35: add RFC describing a simple string-based syntax for generating RFC 31 constraint objects

grondo:
A simple "mini language" for generating RFC 31 style JSON "constraint" objects has been proposed in a couple different experimental PRs (flux-framework/flux-core#4677 and flux-framework/flux-core#4849). Therefore it seemed like a good idea to formalize the proposed syntax in an RFC.

This is a WIP to allow others to comment and approve of the general syntax being proposed. It should be noted that the proposal is inspired by existing search/query syntax in tools like Google search, Gmail and [Apache Lucene](https://lucene.apache.org/core/2_9_4/queryparsersyntax.html), which should make it familiar to most users.

### DIFF
--- a/spec_35.rst
+++ b/spec_35.rst
@@ -1,0 +1,177 @@
+.. github display
+   GitHub is NOT the preferred viewer for this file. Please visit
+   https://flux-framework.rtfd.io/projects/flux-rfc/en/latest/spec_31.html
+
+35/Constraint Query Syntax
+==========================
+
+This specification describes a simple string syntax which can be used to
+succinctly express constraints as described in RFC 31. The syntax may
+also be useful in generating search and match expressions for use in other
+parts of Flux.
+
+-  Name: github.com/flux-framework/rfc/spec_35.rst
+-  Editor: Mark A. Grondona <mgrondona@llnl.gov>
+-  State: raw
+
+Language
+--------
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to
+be interpreted as described in `RFC 2119 <https://tools.ietf.org/html/rfc2119>`__.
+
+Related Standards
+-----------------
+
+-  :doc:`31/Job Constraints Specification <spec_31>`
+
+Goals
+-----
+
+-  Describe and define a grammar for a simple text-based string syntax for
+   generating JSON objects in the constraint format defined in RFC 31.
+
+Background
+----------
+
+The JSON constraint format described in RFC 31 is not conducive to use
+on the command line or for other cases where constraints are input by
+humans. The underlying format has also shown promise as a candidate for
+use in other search and matching scenarios, for example when filtering
+jobs. Because of this potential for use throughout Flux, it will be
+useful to define a simple query syntax which can compile to the more
+verbose JSON format.
+
+This RFC describes and defines a simple string syntax which can be used
+by Flux tools and commands to generate RFC 31 compatible JSON constraint
+objects.
+
+Description
+-----------
+
+ * A constraint query string is formed by a series of terms.
+ * A term has the form ``operator:operand``, where ``operator:`` is
+   optional if the implementation defines a default operator. For example,
+   if an implementation defines a default operator of ``name``, then
+   then string ``myname`` is equivalent to ``name:myname``.
+ * Terms may optionally be joined with boolean operators and parenthesis
+   to allow the formation of more complex constraints or queries.
+ * Boolean operators include logical AND (``&``, ``&&``, or ``and``),
+   logical OR (``|``, ``||``, or ``or``), and logical negation (``not``).
+   The shorthand operators allow terse strings for some implementations
+   e.g. ``a|b|c``.
+ * Terms separated by whitespace SHALL be joined with logical AND by default.
+ * Quoting of terms SHALL be supported to allow whitespace and other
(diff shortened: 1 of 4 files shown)


grondo:
BTW, I do have a very simple parser for this proposal implemented with `python-ply` (chosen because it is already a requirement of `python-pycparser` and thus flux-core. So this would not add a new dependency.

If anyone is interested, I have that parser parked on my `cparser` branch, and you can run the parser directly via the module with:
```console
$ flux python -m flux.constraint.parser --default-op=properties -- testprop -host:badhost
{"and": [{"properties": ["testprop"]}, {"not": [{"host": ["badhost"]}]}]}
```
e.g. to simulate how this parser might interpret an argument to the `flux-mini --requires` option.

spec_35.rst
```diff
@@ -0,0 +1,174 @@
...
+Description
+-----------
+
+ * A constraint query string is formed by a series of terms.
+ * A term has the form ``operator:args``. An implementation MAY define a
+   default term operator, in which case a term may simply be a word.
+ * Terms may optionally be joined with boolean operators and parenthesis
+   to allow the formation of more complex constraints or queries.
+ * Boolean operators include ``&``, ``&&``, ``and``, ``|``, ``||``, ``or``,
+   and ``not``. The shorthand
```
garlick, line 59: since (&, &&, and) and (|, ||, or) are all aliases, you might consider indicating this so people don't have to wonder if there are subtle distinctions as in C.

garlick (approved):
Just a couple of suggestions, not at all show stoppers.  This is good!

spec_35.rst
```diff
@@ -0,0 +1,174 @@
...
+This RFC describes and defines a simple string syntax which can be used
+by Flux tools and commands to generate RFC 31 compatible JSON constraint
+objects.
+
+Description
+-----------
+
+ * A constraint query string is formed by a series of terms.
+ * A term has the form ``operator:args``. An implementation MAY define a
+   default term operator, in which case a term may simply be a word.
```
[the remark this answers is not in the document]
grondo: Good point! The thing I was trying to get across is that a term is `[operator:]operand` where the operator may be optional if a the implementation defines a default operator, and `operand` may be expanded to one or more `values` in the generated RFC 31 object.

I realize I didn't quite get that across here, so I'm happy accept alternate wording and/or try again. The operand or values is definitely not optional `foo:` is a syntax error.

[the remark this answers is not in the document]
grondo: Hm, maybe just what I said above is slightly better than the existing wording, along with your suggestion to use `operand` instead of `args`?

spec_35.rst
grondo: Ah, yeah, great idea. What about something like this?

>  * Boolean operators include logical AND (``&``, ``&&``, or ``and``),
   logical OR (``|``, ``||``, or ``or``), and logical negation (``not``).
   The shorthand operators allow terse strings for some implementations
   e.g. ``a|b|c``.

spec_35.rst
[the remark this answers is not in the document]
grondo: This any better?

>  A term has the form ``operator:operand``, where ``operator:`` is
   optional if the implementation defines a default operator. For example,
   if an implementation defines a default operator of ``name``, then
   then string ``myname`` is equivalent to ``name:myname``.

grondo:
I've removed WIP since it seems like we've got some agreement on at least the basics and are now just working on details.

grondo:
BTW, another well-known interface that uses a similar syntax to what is being proposed here is [Github search](https://docs.github.com/en/search-github/getting-started-with-searching-on-github/understanding-the-search-syntax). They seem to call search/query terms `qualifiers`.

grondo:
Ok, pushed a fixup commit to address some of the comments above.

garlick:
It seems good to me FWIW!

grondo:
Squashed the fixups and pushed the result.

spec_35.rst
```diff
@@ -0,0 +1,177 @@
...
+   optional if the implementation defines a default operator. For example,
+   if an implementation defines a default operator of ``name``, then
+   then string ``myname`` is equivalent to ``name:myname``.
+ * Terms may optionally be joined with boolean operators and parenthesis
+   to allow the formation of more complex constraints or queries.
+ * Boolean operators include logical AND (``&``, ``&&``, or ``and``),
+   logical OR (``|``, ``||``, or ``or``), and logical negation (``not``).
+   The shorthand operators allow terse strings for some implementations
+   e.g. ``a|b|c``.
+ * Terms separated by whitespace SHALL be joined with ``and`` by default.
```
garlick, line 64: Might read slightly better to say "SHALL be joined with logical AND by default"?

spec_35.rst
```diff
@@ -0,0 +1,177 @@
...
+ * Quoting of terms SHALL be supported to allow whitespace and other
+   reserved characters in ``operand``, e.g. ``foo:'this is args'``
+ * A negation operator is supported in front of terms such that ``-op:value``
+   is equivalent to ``not op:value``. Negation MAY NOT be used in front
```
garlick, line 68: For consistency, these two occurrences of "value" should probably be changed to "operand".

garlick (approved):
Had another quick readthrough and noticed a couple of small things.  Otherwise looks good!

spec_35.rst
grondo: Agreed, thanks!

spec_35.rst
grondo: Good catch, thanks!

grondo:
Thanks! I made the suggested changes and will set MWP.

PR #361, opened 2023-01-26
RFC 3: clarify byte order of 32-bit integers in PROTO frame

garlick:
Problem: RFC 3 is wishy washy about byte order within the PROTO frame.

Drop an incorrect statement from a figure caption, and replace incomplete comments in the ABNF wtih a definitive statement in the prose.


grondo (approved):
:+1:

PR #362, opened 2023-01-26
rfc31: update Job Constraint Specification

grondo:
Now that we're close to supporting conditional operators and `hostlist` and `ranks` in Fluxion and sched-simple, update the RFC to require these operators be supported for a scheduler to be RFC 31 compliant.

### DIFF
--- a/spec_31.rst
+++ b/spec_31.rst
@@ -62,8 +62,7 @@ several advantages:
  * The format allows nesting to support complex constraints
  * Simple cases can be expressed simply
 
-In this version of the RFC, only the following constraint operators SHALL be
-supported
+The following constraint operators SHALL be supported
 
  - ``properties``: The set of values SHALL designate a set of required
    properties on execution targets. As a special case, if a property value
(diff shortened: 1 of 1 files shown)


spec_31.rst
```diff
@@ -62,7 +62,7 @@ several advantages:
  * The format allows nesting to support complex constraints
  * Simple cases can be expressed simply
 
-In this version of the RFC, only the following constraint operators SHALL be
+In this version of the RFC, the following constraint operators SHALL be
 supported
```
garlick, line 66: "In this version of the RFC" seems like it could be dropped?

garlick (approved):
LGTM!  Just one comment, not really related to the proposed change but in a line that was touched.  Feel free to ignore.

spec_31.rst
grondo: Good point.

grondo: Ok, dropped that phrase and will set MWP.

PR #364, opened 2023-02-01
rfc6: improve diagrams

garlick:
Problem: rfc6 includes one lame ascii diagram with no detail.

Add three diagrams that illustrate various RPC types in greater detail.


grondo (approved):
:+1:

PR #367, opened 2023-02-14
rfc36: fix invalid example

grondo:
Problem: One of the examples in RFC 36 is invalid because it does not include the directive sentinel for all lines of a multiline quoted string.

Fix the invalid example to comply with the RFC.

Also fix a minor typo in another example.


chu11 (approved):
LGTM!

grondo:
whew, thanks!

PR #368, opened 2023-02-19
WIP: rfc37: add Flux JSON Archive specification

grondo:
This is a first rough draft specification for storing an "archive" of files in a JSON object. The spec is purposefully very simple for now.

The idea here is that once this RFC is official, we can specify an `attributes.system.inputs` or similar key in RFC 14 as a generic way to encode files that are used as input to a job, e.g. the batch script, an instance config file etc. The job shell would then treat all these inputs in the same way, extracting the "archive" to the job's TMPDIR and allowing the job to find all inputs in a standard way, instead of using a specific method or plugin for each file.

An archive can contain other directories in this specification, though perhaps that was a misstep since that isn't a known use case at this time. The content of regular files must be encoded as utf-8, as we may want to discourage the placement of large binary blobs in JSON. (we have the stage-in/out plugin for this). However, an `encoding` key could be added in the future to allow small binary files, e.g. to allow encrypted or compressed files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ Table of Contents
 - [34/Flux Task Map](spec_34.rst)
 - [35/Constraint Query Syntax](spec_35.rst)
 - [36/Batch Script Directives](spec_36.rst)
+- [37/Flux JSON Archive](spec_37.rst)
 
 Build Instructions
 ------------------
--- a/index.rst
+++ b/index.rst
@@ -251,6 +251,12 @@ JSON objects in the format described in RFC 31.
 This specification defines a method for embedding job submission options
 and other directives in files.
 
+:doc:`37/Flux JSON Archive <spec_37>`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This specification defines a method for archiving a directory of files in
+JSON.
+
 .. Each file must appear in a toctree
 .. toctree::
    :hidden:
@@ -290,3 +296,4 @@ and other directives in files.
    spec_34
    spec_35
    spec_36
+   spec_37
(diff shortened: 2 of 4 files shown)


garlick:
> An archive can contain other directories in this specification, though perhaps that was a misstep since that isn't a known use case at this time

IMHO it may be a mistake to define the archive object recursively if the goal is just to be able to represent a hierarchy of files.   The KVS metadata defined in [RFC 11](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_11.html) is recursive, and unwinding and composing those objects is more complex than dealing with a flat list of objects like a tar archive.  In a tar archive, each archived object includes a full path (relative to the archive root) and directory objects have no contents (just metadata).  This is because tapes are sequential devices, but it turns out to be kind of handy anyway!

I did define a [fileref object](https://github.com/flux-framework/flux-core/blob/master/src/common/libutil/fileref.c) (not to be confused with the KVS one in RFC 11) to represent a file object for the `content-mmap.list` response (used in stage-in). It fits the "tar" model and supports files, symlinks, and directories.  Files < 4K are base64 encoded.  Larger files are externally referenced in the content store.  If you agree that this is the right approach, maybe we tweak that object slightly (and rename) so it fits here too?

There is even code in `stage-in`, and sadly duplicated in `flux-fileref` to pass a list of those objects to libarchive for extraction into $FLUX_JOB_TMPDIR.

grondo:
> IMHO it may be a mistake to define the archive object recursively if the goal is just to be able to represent a hierarchy of files. T

Ok, that sounds like the right advice. I think the libutil fileref object is overkill in this scenario, and would possibly pollute jobspec with unnecessary file attributes in the common case (we are not really archiving files but representing a _possible_ hierarchy of input files in the main use case at this point).

Given that the "specification" then becomes "a dictionary where keys are paths and values are objects with keys `mode` and `value`", at least the content of this draft is not useful and possible there is no necessity for an RFC. (The specification can be made in-place in the jobspec RFC as we've done for other special keys in jobspec.)

Closing for now.

PR #369, opened 2023-02-28
rfc37: add File Archive Format RFC

garlick:
Problem: flux-filemap(1) and the stage-in shell plugin use a first draft of a JSON file archive format, and now we need something similar for embedding input files in jobspec.

Add an RFC proposing a format that would work for both.


grondo (approved):
Looks great!

garlick:
OK, I have completed a rough conversion of the [flux-filemap(1)](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man1/flux-filemap.html) and it went fairly well so  I'll drop the wip and set MWP.  Thanks!

PR #371, opened 2023-03-21
add command line jobspec

vsoch:
Problem: we cannot define a shape on the command line to map into a job specification resources section. Solution: first add a command line jobspec that defines this shape.

I've never written one of these before - let me know what I should expand / contract or fix!


garlick:
Slightly concerned about muddying the terminology here since we already have a jobspec and it contains a different set of information.  Maybe we could call this a "short form resource request" or maybe a "job shape specification"?  Also, the fact that it fits conveniently on the command line could just be one item in the goals section rather than in the title.

One thing I thought turned out well in some of our other RFCs is to have a Test Vectors section at the end that specifies a table of inputs and outputs.  If this is a strict translation from "new thing" to  the jobspec v1 Reources section, then maybe that would be applicable here?

Finally just wanted again to mention [OAR](https://oar-3.readthedocs.io/en/latest/_static/hierarchical_resources.png) as I think their `--resources` option is similar and their choice of characters like using resource=count instead of resource[count] seems like it might be a little easier for users to work with.  I do like their use of `/` to denote hierarchy also.

vsoch:
Okay I’ll ping - @trws - he suggested the command line job spec (which makes sense to be different I think) and we talked about the design of the shape in slack today.

garlick:
OK!  Those were just my off the cuff first impressions, so take them with appropriate grains of salt etc

vsoch:
yeah totally! Here is count as an attribute:

```
# If it's allowed to go first (and then later)
Node[4]:Slot[count=1+label=default]:Core[count=2]
Node[4]:Slot[label=default+count=1]:Core[count=2]
```
If I had to guess, the rationale for just using the number is that it's common enough that we shouldn't have to write out `count=` every time. It's a single number (either present or not) at the beginning of the declaration.

PR #373, opened 2023-04-12
rfc5: update to cover recent flux-core changes

garlick:
Problem: the RFC is out of date with current broker behavior with regard to broker module loading.

Update some of the introductory paragraphs.

Drop the requirement for the 'mod_name' symbol.

Drop the protocol definitions for module management.  Now that "submodules" are not supported, modules no longer need to implement the protocol, so it can be private.

Add a section on module attributes covering how to obtain the uuid and module name from within a module.

Add a section on the newly added multiple loading feature.


grondo (approved):
LGTM!

PR #374, opened 2023-04-14
rfc20: add basic R example

chu11:
A few times I've wanted a simple example and end up digging one up out of an active Flux instance instead.  Add a basic R example into rfc20.

Would be great if there were a few more advanced ones with more of the optional keys, but not too familiar with them.  Hopefully this is a starting point.


grondo (approved):
Sorry, missed reviewing this one when it was proposed.
Seems like a good addition!

Issue #376, opened 2023-05-01
rfc31: need to define an "always match" / "never match" constraint object


chu11:
In rfc31 "not specifying" a constraint object would mean "match everything" or "there are no constraints".

But there may be circumstances we would like to define a constraint object that "matches everything" or indicates "there are no constraints".  For example, lets say we had the function:

`json_t *build_me_constraint_object (param1, param2, param3);`

This function could return NULL to indicate "no constraint object" given the inputs, but if it might be convenient if it could always return an object.

https://jsonlogic.com/ proposes that json `true` and `false` be defined for "always match" and "never match".  But given our use case, we feel this is not a good use in flux.

After some offline discussion with @grondo, we think defining an empty json object, i.e. `{}` might be very suitable for "no constraints" and "always matches".

But we do hit the issue of what is the "never matches" case?  I admit I'm not 100% sure.

Given my comments in #375, an alternate proposal might be:

"always match" :  `{}` or `{ "and": [] }` or `{ "or": [] }`.
"never match": `{ "not": [] }`

chu11:
minor correction in what jsonLogic says:

```
Always and Never

Sometimes the rule you want to process is “Always” or “Never.” If the first parameter passed to jsonLogic is a non-object, non-associative-array, it is returned immediately.

//Always
jsonLogic.apply(true, data_will_be_ignored);
// true

//Never
jsonLogic.apply(false, i_wasnt_even_supposed_to_be_here);
// false
```

it's not that true/false means "true/false" its just that the code just processes what was passed in if it's not an object.

grondo:
Keep in mind that RFC 31 is "inspired" by JsonLogic, not meant to be compatible or comply with JsonLogic.
Without thinking about too hard, it makes sense to have an empty constraint `{}` be legal and treated the same as no constraint, which would match anything.

For match nothing, `{"not": []}` works for me, though I do wonder if we could come up with something clearer.

Issue #375, opened 2023-05-01
rfc31: conditionals corner case definition


chu11:
While working on https://github.com/flux-framework/flux-core/issues/4914 an interesting corner case came up.  I think we just have to pick a definition and define it.  For empty conditionals, it makes sense that.

`{ "and": [] }`

and

`{ "or": [] }`

always return true / "match"

But what about 

`{ "not": [] }`

My initial feeling was to define this as `true`, because my feeling was that if the user specified no constraint operations, even though a `not` was put in front, it should still evaluate to `true` ... i.e. "nothing" -> true, "not nothing" -> true.  i.e. "nothing" is some magical "between true and false" value.

But an offline discussion with @grondo, he felt it should be `false`.  With the logic, no constraint means "match", so "not no-constraint" => "not match" => false.

That also makes sense to me.  My opinion is waffling.  I don't have super strong opinions on this, dunno if anyone wants to throw in a vote if they have strong opinions.

PR #377, opened 2023-05-01
rfc31: add clarification on several corner case / use cases

chu11:
This is a proposal for clarifications on #375 and #376.  I'm not sure if this will be liked or not, but could atleast start a discussion.

### DIFF
--- a/spec_31.rst
+++ b/spec_31.rst
@@ -95,6 +95,12 @@ The following constraint operators SHALL be supported
 
  - ``and``: Simple logical ``and``.
 
+If a constraint operator does not list any ``values``, behavior is operator
+dependent.  The operator may return a match, not a match, or report an error.
+However, when no ``values`` are listed for the conditional operators listed above,
+``{ "or": [] }`` and ``{ "and": [] }`` are defined to always return true and
+match anything.  ``{ "not": [] }`` is defined to return false and match nothing.
+
 Examples
 --------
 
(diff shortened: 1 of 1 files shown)


chu11:
Hmmm, actually i'm already regretting this:

```
If a constraint object does not list any ``values``, generally speaking
the constraint will always return true and match anything.  The lone exception
to this is the ``not`` operation.  ``{ "not": [] }`` is defined to return false
and match nothing.
```

it should not be for all constraint operations, (note I say "object" above, I meant operation).  Maybe we define this only for `and` and `or`, but everything else is dependent.  i.e. like it can be an error for certain operations.

chu11:
re-pushed with some tweaks, ehhhh?

chu11:
hmmmm, actually should `{ "or": [] }` be false by default?  b/c implementation wise this would be far more common.

```
foreach value in values array:
     if test match == true:
         return true
return false
```

with and being the opposite of course.

```
foreach value in values array:
     if test match == false:
         return false
return true
```

grondo:
Yeah, we shouldn't focus on possible implementations when defining the spec.
This is all a bit of an academic exercise, but if `{ "and": [] }` is true then that implies that empty args is true, which means `{ "or": [] }` should also be true. Perhaps the spec itself is a bit to blame here, maybe empty args shouldn't be allowed to these operations, an empty object is "no constraint" and matches everything, and a constraint that matches nothing isn't specified, but could easily be supplied on a per-case basis with something like `{"name": ["noexist"]}`.

grondo (approved):
Seems reasonable and a nice clarification.

PR #378, opened 2023-05-04
rfc24: add repeat key to data event

garlick:
Problem: when a job produces many identical lines of output, such as when repeatedly encountering the same error, we have no way of representing that efficiently in the output eventlog.

Add a repeat key to data events.

Identical data items can be de-duplicated when added to the eventlog, and re-duplicated when converted back into a stream.

### DIFF
--- a/spec_24.rst
+++ b/spec_24.rst
@@ -60,6 +60,8 @@ Goals
 
 -  Support shell error and debug logging.
 
+-  Support de-duplication of stream contents, where applicable.
+
 
 Implementation
 --------------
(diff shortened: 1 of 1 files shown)


chu11:
Do we want to mention something about the timestamp range that is acceptable for repeats?  I don't think we should define an actual number like "repeats within X seconds", but it could simply say "dependent on implementation".  Or maybe not worth mentioning at all.  Perhaps it's kind of obvious.

chu11:
oh wait, should the "repeat window" be documented in the header for the stream? in fsd format?

garlick:
I guess the repeat window (good term!) would be useful for interpreting timestamps of repeat events when looking at raw logs, but I'm not sure that's likely to be all that useful.  We could always add it in later if needed?

I did waffle on at least mentioning that there's a time range to consider but ultimately it seemed like it would be wasted words so I left it out.

chu11:
>  We could always add it in later if needed?

That seems fair.  I suppose adding it to the header was only some "document it for sake of documenting it", but you're right, it perhaps has little actual value.

chu11 (approved):
LGTM

garlick:
Seem to have hit he same mergify snafu as in flux-framework/flux-core#5150.  I'll just press the button.

PR #381, opened 2023-05-16
rfc19: add emoji FLUID encoding

grondo:
This update to RFC 19 describes a method for encoding FLUIDs to a string of emoji using a table of 576 selected emoji.

Note that since the emoji all have a common 2 prefix bytes, a prefix character is not necessary and thus it was dropped from the proposed standard. We could add that back if it is desired.


spec_19.rst
```diff
@@ -152,6 +152,39 @@ by speaking, e.g. over the phone.
...
+
+In order to encode a FLUID using a minimal number of printable characters,
+and to increase visual appeal when displaying FLUIDs in various settings,
+a conformant implementation MAY implement an encoding of FLUIDs to a string of
+`Unicode emoji <https://www.unicode.org/reports/tr51/>`__ characters.
+
+The emoji implementation SHALL consist of the following 576 emoji:
+
+.. literalinclude:: data/spec_19/basemoji.h
+   :language: C
```
chu11, line 166: forgot to add file?

grondo: yep :-(

grondo: try again?

chu11: it's glorious!

chu11, line 166: perhaps move this chunk below after the `emoji-ordering.txt` chunk.  Say the resulting emoji table is this.

grondo: Ok, that's how I had it at first, but it seemed awkward. I'll give it another shot though

grondo: Does it seem any better now?

garlick:
Nice!  No 🇫 prefix?

spec_19.rst
chu11: yeah, i think it flows a little better.

chu11:
I think  🇫 prefix would be nice, and it seems to be from the 2010 set and starts with 0xF0 0x9F.

grondo:
Yeah, note as stated in the description, I dropped it because it wasn't necessary. We can add it back.

Note that 🇫  is a "Regional Indicator' and thus is _not_ in the [Basic_Emoji](https://www.unicode.org/reports/tr51/#def_basic_emoji_set) spec. In fact, the spec states:

> Other components (U+20E3 COMBINING ENCLOSING KEYCAP, Regional Indicators, tag characters, ZWJ, and VS16) should never have an emoji presentation in isolation, but do occur as part of emoji sequences. These are not included in Basic_Emoji.

The "should never have an emoji presentation in isolation" gave me pause given that an 🇫 turns out to be unnecessary and adds an extra 4 bytes and one glyph to the presentation.

However, in practice obviously the 🇫 seems to work just fine so we can easily add it as required here.

Here's the difference in output for comparison:
```
😄😹🎇📥🏧🙉🔞
😄😁💤📞🍀💞😃
🈹🍶🏀🕕🍒🌼
🉐🎂💆🍎🈲👘
🉐🍳🍚💌🏦🏃
🚹💂🈳💰🎩🏃
🚹💂🌛💄💧🌼
🚹💂🙌😳💱🏃
🚹🙇💄👎🌄🍹
🚹💁🍏🍛🎢🕓
🚹🙎💥🏮🙉🔞
🚹👩🔱🍦💁📓
🏧🌏🚑🌼🔖💞
```

```
🇫😄😹🎇📥🏧🙉🔞
🇫😄😁💤📞🍀💞😃
🇫🈹🍶🏀🕕🍒🌼
🇫🉐🎂💆🍎🈲👘
🇫🉐🍳🍚💌🏦🏃
🇫🚹💂🈳💰🎩🏃
🇫🚹💂🌛💄💧🌼
🇫🚹💂🙌😳💱🏃
🇫🚹🙇💄👎🌄🍹
🇫🚹💁🍏🍛🎢🕓
🇫🚹🙎💥🏮🙉🔞
🇫🚹👩🔱🍦💁📓
🇫🏧🌏🚑🌼🔖💞
```

garlick:
Oh sorry I missed that in the description.  It just seemed nice, as with F58, to have a prefix that indicates "this is a job ID",
but I'm fine with dropping it (or selecting a different prefix) if the 🇫 isn't kosher here.

grondo:
I'm fine with a prefix, though we should probably then remove the statement "minimal characters" from the design criteria for the spec. It seems to work fine in practice, so I'm not sure my worry about using regional indicator characters alone is an issue. There are no other regional indicators allowed so it isn't like we'll get `🇫` + `🇮` next to eachother and end up with `🇫🇮` (which is what happens when you put two regional indicators together to form a country code)

Note also that F58 only uses a prefix to distinguish it from other encodings (decimal, words, dotted hex do not have prefixes either).

I could go either way on this one. I like the prefix, but dislike that it adds unnecessary length to the encoding. However, if @chu11 also votes to require a prefix I'll go ahead and include it.

garlick:
It's a good point that F58 needs a disinguisher while emoji encoding does not.  What do you think @chu11, are you convinced one way or the other?  I'm fine either way at this point.

chu11:
ya know, when I saw a single emoji encoding before, I liked the 🇫 prefix, b/c it reminded me of the the F58.  But seeing the example above with all of the emojis listed, which is what we may have in `flux-jobs`, it seems a little out of place and unnecessary.  I think I prefer the wall of "random" emojis instead :-)  So I'd lean to nix it.

vsoch:
But where is the plucky poop bomb? I don't know if I can move forward without it! :laughing:

garlick:
> But where is the plucky poop bomb?

Did you mean the standard poomoji?  Row 4 column 5:  https://flux-framework--381.org.readthedocs.build/projects/flux-rfc/en/381/spec_19.html

grondo:
> Did you mean the standard poomoji? Row 4 column 5

For the purposes of this RFC, that emoji SHALL now be known as "numeral poo".

grondo:
> xample above with all of the emojis listed, which is what we may have in flux-jobs, it seems a little out of place and unnecessary. I think I prefer the wall of "random" emojis instead :

This was my feeling as well when seeing many FLUIDs in this encoding...I'll propose an implementation in a PR to flux-core and we can go from there.

chu11:
TIL that originally poomojis did not have eyes, they were just piles of poo.  They were later  anthropomorphized, apparently drawing inspiration from poop-boy (https://drslump.fandom.com/wiki/Poop-Boy).

chu11 (approved):
it seems we're not going with the prefix, so approved!

grondo:
Thanks!

PR #382, opened 2023-05-30
rfc14: add `files` section to jobspec

grondo:
Problem; There is no standardized method for embedding files in jobspec.  Currently, `flux batch` and the job shell use an ad-hoc mechanism to store the batch script under `attributes.system.batch.script`, but it would be useful to formalize this case and open the door for the storage of other simple files in jobspec, with automated extraction by the job shell.

Document `attributes.system.files` as an RFC 37 file archive, which shall be extracted by the job shell into the job's temporary directory. Reserve two files: `script` for any job batch script, and `conf.json` for a subinstance config file.

This PR is currently a WIP while a prototype of the job shell and cli interface are developed. We can also start a discussion to make sure all use cases are covered with this simple proposal.

### DIFF
--- a/spec_14.rst
+++ b/spec_14.rst
@@ -378,6 +378,21 @@ Some common system attributes are:
       for the job shell. A job shell and its plugins are free to define
       what keys and values should go into ``options``.
 
+**files**
+   The ``files`` key SHALL consist of a dictionary in RFC 37 File Archive
+   Format containing files that SHALL be made available to the job.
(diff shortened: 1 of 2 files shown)


grondo:
I renamed the jobspec key for this feature from `inputs` to `files`, which seemed more clear and specific. Also reworded the prose a bit.

garlick (approved):
This sounds good to me.

grondo:
prototype is going well with this simple format, so I think I'll go ahead and set MWP.

PR #386, opened 2023-07-19
rfc21: make exception note required

chu11:
Problem: Currently the exception note is defined as optional.  Making the exception note optional makes parsing exceptions difficult.

Solution: Make an exception note required.  If one is not desired, it should be set to an empty string.

---

per discussion in https://github.com/flux-framework/flux-core/pull/5336

I went with the language "should" set it to empty string.  Because implementation could choose something different, eg. "no note specified", etc.

### DIFF
--- a/spec_21.rst
+++ b/spec_21.rst
@@ -548,10 +548,12 @@ severity
    (integer) Specify the severity of the exception, in range of 0 (most severe)
    to to 7 (least severe).
 
-The following keys are OPTIONAL:
-
 note
-   (string) Brief human-readable explanation of the exception.
+
+   (string) Brief human-readable explanation of the exception.  If no explanation
+   is given, an empty string SHOULD be set for the note.
+
+The following keys are OPTIONAL:
 
 userid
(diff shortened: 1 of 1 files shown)


spec_21.rst
```diff
@@ -548,10 +548,12 @@ severity
    (integer) Specify the severity of the exception, in range of 0 (most severe)
    to to 7 (least severe).
 
-The following keys are OPTIONAL:
-
 note
-   (string) Brief human-readable explanation of the exception.
+
+   (string) Brief human-readable explanation of the exception.  If no explanation
+   is given, an empty string should be set for the note.
```
garlick, line 554: The "should" should be capitalized since it's an RFC 2119 key word.

garlick (approved):
LGTM! Just one minor quibble, but you can set MWP after fixing that I think!

chu11:
@garlick Thanks.  re-pushed with that tweak.  I'll hold off on setting MWP once we all agree that what is in https://github.com/flux-framework/flux-core/pull/5336 is the way we want to go.

PR #387, opened 2023-08-16
rfc27: add sched.expiration RPC

garlick:
Problem: there is no way for the job manager to notify the scheduler that a job's expiration has been adjusted, nor for the scheduler to fail such a request if it would interfere with an advance reservation.

Add the sched.expiration RPC.

### DIFF
--- a/spec_27.rst
+++ b/spec_27.rst
@@ -103,6 +103,8 @@ Design Criteria
 - Capture scheduler specific job annotations for display by the job listing
   tool (e.g. start time estimates).
 
+- Allow the expiration time of a resource allocation to be adjusted.
+
 
 Implementation
 --------------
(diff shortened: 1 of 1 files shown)


spec_27.rst
```diff
@@ -541,6 +543,37 @@ No response is sent to the ``sched.prioritize`` request.
...
+- The job ID is invalid or does not currently have an allocation.
+
+- The new expiration time would invalidate an advance reservation.
+
+- The scheduler does not implement ``sched.expiration``.
+
+.. note::
+    The job manager SHALL interpret an ENOSYS error response to
+    ``sched.expiration`` as "not implemented" and proceed as though the
+    request was successful, causing the new expiration time to be enforced.
```
chu11, line 573: i assume enforcement is managed by the job-manager?  Perhaps can just add that small snippet of clarification in here.  Should it be "SHALL be enforced by the job-manager"??

garlick: I pictured the job manager making the `sched.expiration` request, then notifying the systems that do enforcement (e.g. job-exec etc).  Is this better?
```  
    The job-manager SHALL interpret an ENOSYS error response to
    ``sched.expiration`` as "not implemented" and process the new
    expiration time as though the request were successful.  Schedulers
    that require accurate expiration times SHOULD implement this RPC to
    avoid making schedules that are based on outdated information.
    Note that enforcement of expiration times is the responsibility of the
    execution system, not the scheduler.

```

spec_27.rst
```diff
@@ -541,6 +543,38 @@ No response is sent to the ``sched.prioritize`` request.
...
+
+The response consists of an empty payload on success.
+
+The request MAY fail for the following reasons:
+
+- The job ID is invalid or does not currently have an allocation.
+
+- The new expiration time would invalidate an advance reservation.
+
+- The scheduler does not implement ``sched.expiration``.
```
grondo, line 568: Just double checking that this is not meant to be an exhaustive list? If so, then another case in which the scheduler might fail the request is if the instance lifetime is less than the new expiration. (IOW, the current expiration of R from which the scheduler is allocating resources is less than requested expiration)

grondo (approved):
LGTM!

spec_27.rst
garlick: That wasn't meant to be an exhaustive list so I'll fix the wording, but that particular failure case is important so I'll add it.

garlick:
I'll go ahead and set MWP here since this was approved earlier.

PR #389, opened 2023-08-18
rfc38: add rfc for flux-security key-value encoding

garlick:
Problem: the key-value format used by flux-security ostensibly for auditability is not documented.

Add an RFC.

### DIFF
--- a/spec_38.rst
+++ b/spec_38.rst
@@ -1,0 +1,159 @@
+.. github display
+   GitHub is NOT the preferred viewer for this file. Please visit
+   https://flux-framework.rtfd.io/projects/flux-rfc/en/latest/spec_38.html
+
+###################################
+38/Flux Security Key Value Encoding
+###################################
+
+The Flux Security Key Value Encoding is a serialization format
+for a series of typed key-value pairs.
+
+- Name: github.com/flux-framework/rfc/spec_38.rst
+
+- Editor: Jim Garlick <garlick@llnl.gov>
+
+- State: raw
+
+********
+Language
+********
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to
+be interpreted as described in `RFC 2119 <https://tools.ietf.org/html/rfc2119>`__.
+
+*****************
+Related Standards
+*****************
+
+- :doc:`15/Independent Minister of Privilege for Flux: The Security IMP <spec_15>`
+
+**********
+Background
+**********
+
+The flux-security project requires simple objects to be encoded for
+communication in the following security-sensitive situations:
+
+- To receive *J* (signed jobspec plus metadata).
+
+- When communicating options between privileged and unprivileged sections
+  of the IMP.
+
+*****
+Goals
+*****
+
+- Replace JSON for security-sensitive use cases where auditability of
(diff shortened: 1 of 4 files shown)


spec_38.rst
```diff
@@ -0,0 +1,152 @@
...
+     - ``DOUBLE_INF\0dinf\0``
+   * - DBL_MIN
+     - double
+     - 2.2250738585072014e-308
+     - ``DBL_MIN\0d0.000000\0``
+   * - DBL_MAX
+     - double
+     - 1.7976931348623158e+308
+     - ``DBL_MAX\0d179769313486231570814527423731704356798070567525844996598917476803157260780028538760589558632766878171540458953514382464234321326889464182768467546703537516986049910576551282076245490090389328944075868508455133942304583236903222948165808559332123348274797826204144723168738177180919299881250404026184124858368.000000\0``
+   * - FALSE
```
grondo, line 141: Are negative doubles supported? if so `-DBL_MAX` might be a good test vector since it adds one character to the string representation. (Sorry I didn't notice this PR until after your flux-security PR)

grondo (approved):
This LGTM. I'd like to get @dun to take a look at it and approve if that's ok.

dun (commented):
What's the maximum length of a key name or value string? How do you protect against something sending an endless string?

garlick:
No max key length or max overall length is currently defined.  Key max would be trivial.  Not sure about overall ...

spec_38.rst
garlick: Ah good one.  Yeah I guess they should be.  I'll add that.

garlick:
A max value length would need to handle all possible serialized jobspecs, which contain a user's environment.  So maybe 1MB?  I think the environment is limited to something on the order of 128K so that would give some leeway.  Perhaps we could just say the overall encoded length is limited to 1MB and then not set limits on key length or string value length?

Presumably preventing someone from running the IMP out of memory by passing it some bad input is one part of a defense in depth strategy?

garlick:
Pushed with -DBL_MAX as a test vector.

@grondo any thoughts on limiting the size of various things?

grondo:
No, but remember jobspec can now have (small) files embedded. Having a size limit seems reasonable but it would be annoying to users and admins to have jobs fail when they are started even after passing other validation (but it would be easy to place and limit and enforce that on the validator side?) Sorry I'm having trouble coming up with any coherent thought on this one...

dun:
I think you're ok in having an overall length of 1MB or whatever for the total encoded length. This limit could even be specified in the imp config file or somesuch (so maybe this rfc isn't the place for it to be defined). You just need to protect against bad input exhausting resources.

garlick:
I had forgotten about the files embedded in jobspec.  1MB may be on the small side.  I guess the main thing is we don't want to be able to trigger a malloc failure.  Heck a 1GB limit would do that.  Could we set the limit at 1GB and ignore it in the rest of the system?

If we don't currently impose any limits at all on the size of generated jobspec, we might want to add something that could reject jobs at submit time to prevent abuse of the input file thing.

garlick:
Ah I didn't see your comment @dun.  Maybe the RFC should just say that the size SHALL be limited and not specify what the limit is.

garlick:
Added:
> An implementation defined limit SHALL be imposed on the maximum overall size
of a serialized object to avoid resource exhaustion.

dun (approved):
LGTM!

PR #390, opened 2023-08-19
rfc38: clarify disposition of null characters in keys and string values

garlick:
Problem: rfc38 doesn't say anything about (valid) UTF-8 strings containg the NULL character.

Add clarification.

### DIFF
--- a/spec_15.rst
+++ b/spec_15.rst
@@ -29,6 +29,8 @@ Related Standards
 
 -  :doc:`12/Flux Security Architecture <spec_12>`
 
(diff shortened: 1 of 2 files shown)


grondo (approved):
Ooh, this is a nice improvement, I had forgotten NUL is valid utf-8..

LGTM

grondo:
One pedantic comment: is it technically the NUL character vs null, which I think is a pointer value? I think it is probably clear either way though...

garlick:
Oops, apologies!  I've been called out on the NUL vs NULL before and so should have known better. Will fix!

PR #391, opened 2023-08-22
add RFC for Flux security signatures

garlick:
Problem: flux security signature format is undocumented.

This is a work in progress RFC that attempts to document the format.

### DIFF
--- a/spec_39.rst
+++ b/spec_39.rst
@@ -1,0 +1,193 @@
+.. github display
+   GitHub is NOT the preferred viewer for this file. Please visit
+   https://flux-framework.rtfd.io/projects/flux-rfc/en/latest/spec_39.html
+
+##########################
+39/Flux Security Signature
+##########################
+
+The Flux Security Signature is a NUL terminated string that represents
+content secured with a digital signature.
+
+- Name: github.com/flux-framework/rfc/spec_39.rst
+
+- Editor: Jim Garlick <garlick@llnl.gov>
+
+- State: raw
+
+********
+Language
+********
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to
+be interpreted as described in `RFC 2119 <https://tools.ietf.org/html/rfc2119>`__.
+
+*****************
+Related Standards
+*****************
+
+- :doc:`14/Canonical Job Specification <spec_14>`
+
+- :doc:`15/Independent Minister of Privilege for Flux: The Security IMP <spec_15>`
+
+- :doc:`38/Flux Security Key Value Encoding <spec_38>`
+
+**********
+Background
+**********
+
+The Flux Security IMP is the privileged component of Flux responsible for
+launching tasks in a multi-user Flux installation.  Each job request is
+signed by the submitting user and verified by the IMP as a precondition for
+performing the user transition and starting tasks.  Signing deters tampering
+when the job passes through the unprivileged Flux instance.  This is the
+primary use case for the Flux Security Signature.
+
+A job request is expressed as *jobspec* (RFC 14).  Importantly, jobspec
+contains the proposed command line to run as well as the current working
+directory and environment.  Signing the jobspec ensures that these parameters
+are passed to the IMP exactly as specified by the user.
+
+RFC 15 refers to the signed jobspec as *J*.  The content of *J* is thus fully
+specified by this document and RFC 14.
+
+*****
+Goals
+*****
+
+- The signature mechanism SHOULD be configurable so that a site's existing
+  security infrastructure can be used, if applicable.
+
+- The signature payload MUST accommodate the maximum size of a serialized
+  RFC 14 jobspec.
+
+- The signature MUST be a UTF-8 string that does not contain NUL characters.
+
+- The signature SHOULD support a configurable time-to-live.
+
+**************
+Implementation
+**************
+
+The Flux Security Signature is structured like the JSON Web Signature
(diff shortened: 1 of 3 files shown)


garlick:
This is ready for a review when someone has a moment.

spec_39.rst
```diff
@@ -0,0 +1,185 @@
...
+is not important, such as in a single-user Flux instance.
+
+The header *mechanism* key SHALL be set to the string ``none``.  No additional
+mechanism-specific keys are required.  The *none* mechanism does not support
+a time-to-live.
+
+Signing SHALL consist of the following steps:
+
+#. Set the signature component to the string ``none``.
+
```
grondo, line 175: For the curious, include how _header_ and _payload_ are constructed for sign-type=none?

spec_39.rst
```diff
@@ -0,0 +1,185 @@
...
+Verification SHALL consist of the following steps:
+
+#. Check that the signature component is the string ``none``.
+
+#. Check that the real user id matches the header userid.
+
```
grondo, line 181: Include how payload is extracted?

grondo (approved):
LGTM! Just a couple suggestions for maybe fleshing out the sign-type=none section. If you don't feel that's necessary then I'm fine with this RFC as is!

Might also be good to have @dun give his approval, so I'll request it.

spec_39.rst
garlick: The header and payload are the same regardless of mechanism and can be read or written in a mechanism independent manner.  Maybe I'd better call that out explicitly!

Well that statement is almost true - the header can optionally contain mechanism specific values, although for *none* and *munge* it does not.

grondo: Oh duh, sorry you do explicitly state that above but it may be helpful to re-state it in the 'none' section. Sorry!

spec_39.rst
grondo: Sorry, you do state that the payload is always base64 encoded, my bad!

spec_39.rst
```diff
@@ -0,0 +1,189 @@
...
+- The *signature* SHALL be a mechanism-dependent UTF-8 string that represents
+  a signature over the other two components including the delimiter.
+  The signature string SHALL NOT contain the NUL or period (``.``) characters.
+
+The header SHALL contain the following keys:
+
+version (integer)
+  The Flux Security Signature version (currently must be set to 1).
+
+mechanism (string)
```
dun, line 95: How is this string encoded? NUL terminated?

Or are integer and string encodings defined in another RFC?

spec_39.rst
```diff
@@ -0,0 +1,189 @@
...
+
+The header *mechanism* key SHALL be set to the string ``munge``.  No
+mechanism-specific header keys are required.
+
+Signing SHALL consist of the following steps:
+
+#. Compute the SHA-256 hash digest over HEADER.PAYLOAD
+
+#. Construct a MUNGE payload by concatenating a single byte *hash type* field
+   of *1* (indicating SHA-256) and the SHA-256 hash digest.
```
dun, line 124: How is the *1* represented? Is this the ASCII character `1`? Or a single byte signed or unsigned int?

How is the SHA-256 hash digest represented? As a 64-character string? NUL terminated? Or a 32-byte binary encoding?

Would a delimiter between *hash type* and *hash digest* be helpful? If they're both in ASCII, a 65-character string of `[0-9A-F]` might cause confusion at some point. If they're both in binary, it probably doesn't matter.

dun (approved):
LGTM. I just had a few minor questions on some things that were unclear to me.

spec_39.rst
```diff
@@ -0,0 +1,189 @@
...
+It SHALL contain three components concatenated with period (``.``) delimiters:
+
+HEADER.PAYLOAD.SIGNATURE
+
+- The *header* SHALL be formatted using Flux Security Key Value Encoding
+  (RFC 38), then base64 encoded.  The REQUIRED keys are listed below.
+  The header is not encrypted.
+
+- The *payload* is an arbitrary sequence of zero or more bytes that
+  SHALL be base64 encoded.  The payload is not encrypted.
```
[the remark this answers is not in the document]
garlick: Great point.  Looks like we're using [rfc4648](https://www.rfc-editor.org/rfc/rfc4648#section-4) "standard base64".  I'll call that out.

In flux-security, we use [libsodium SODIUM_BASE64_VARIANT_ORIGINAL](https://libsodium.gitbook.io/doc/helpers#base64-encoding-decoding)

In flux-core we use [ccan base64_maps_rfc4648](https://github.com/rustyrussell/ccan/blob/master/ccan/base64/base64.h).

spec_39.rst
```diff
@@ -0,0 +1,189 @@
...
+
+- The *signature* SHALL be a mechanism-dependent UTF-8 string that represents
+  a signature over the other two components including the delimiter.
+  The signature string SHALL NOT contain the NUL or period (``.``) characters.
+
+The header SHALL contain the following keys:
+
+version (integer)
```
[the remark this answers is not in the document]
garlick: This is defined in [RFC 38](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_38.html).

spec_39.rst
garlick: The *1* is a single byte and the SHA-256 digest is the binary encoding.  I'll try to make those more clear.

> If they're both in binary, it probably doesn't matter.

This.   Although looking back, I'm not sure why it didn't go in the structured HEADER portion.

garlick:
Thanks for the excellent comments @dun - just pushed some tweaks to address them.

garlick:
Setting MWP, thanks @grondo and @dun!

PR #396, opened 2023-08-30
rfc21: allow jobspec-update to any key

### LINKED ISSUE #395: rfc21: allow jobspec-update event for other top-level keys
In flux-framework/flux-coral2#92 the jobspec of a job with a DWS spec is modified in place by the dws jobtap plugin.

It would be better to use an RFC 21 `jobspec-update` event to make this modification, but currently RFC 21 forbids updating any keys that do not start with `attributes.`.

I'm not sure the rationale behind limiting `jobspec-update` events to only the `attributes` section of jobspec. Perhaps we can expand this to allow updates to `resources.` as well since there is this use case necessary in flux-coral2?

garlick:
Problem: the jobspec-update event is constrained to only allow updates to the `attributes` section, but updates to the resources section are needed to implement emerging requirements in flux-coral2.

Allow any key to be updated.

Add some clarity on JSON key "paths" and what is supposed to happen if the key already exists.

Fixes #395

### DIFF
--- a/spec_21.rst
+++ b/spec_21.rst
@@ -210,9 +210,11 @@ The ``submit`` event SHALL be the first event posted for each job.
 Jobspec-update Event
 ^^^^^^^^^^^^^^^^^^^^
 
-Set jobspec attributes after job submission.  The event context object SHALL
-consist of a dictionary of period-delimited keys beginning with ``attributes.``
-and MUST contain at least one entry.
+Change jobspec after job submission.  The event context object SHALL consist
+of a dictionary of period-delimited keys that SHALL be interpreted as a
+hierarchical JSON path.  The dictionary MUST contain at least one entry.
+If the key already exists in the jobspec object, the old value SHALL be
+replaced with the new value.
 
 Example:
 


chu11:
while we're at it, do we want to update `resource-update`?  Or shall we leave as is?

chu11:
another thought, perhaps we should add a clarification / note on how to delete something?  set the value to json null?  
OR perhaps we shouldn't allow?

garlick:
I think it would probably be good to handle `resource-update` separately in another PR, when needed.

Regarding deleting something, say a queue has a default duration.  If you delete the duration, should the jobspec have no duration or should it revert to the default duration?  I think the latter but that may be a bit tricky to accomplish if the default was assigned by the frobnicator at ingest...

chu11:
> Regarding deleting something, say a queue has a default duration. If you delete the duration, should the jobspec have no duration or should it revert to the default duration? I think the latter but that may be a bit tricky to accomplish if the default was assigned by the frobnicator at ingest...

Ahh good point.  My initial thought was on optional fields (job name) but there's a lot of trickery there.  

Perhaps just want to mention no fields can be removed?

chu11:
I was just coding up fixes to `job-list` for this update and it occurred to me, 

> dictionary of period-delimited keys that SHALL be interpreted as a hierarchical JSON path

tasks and resources in the jobspec is a list / json array.  The period delimited path syntax doesn't account for this I think.  should we support an array indexing?  like ` "tasks[0].command[0]" : "sleep" ` for an update?  

Or perhaps this is the original reason why it was `attributes` dictionary only.

grondo:
I think you'd have to read/update/replace the whole array in these cases. That probably makes it a lot simpler to implement the updates in the job manager.

grondo (approved):
Thanks! This looks good to me. I can create a PR for flux-core that modifies `validate_jobspec_updates()` to remove the constraint that updates begin with `attributes.` if nobody else is already doing that.

garlick:
Thanks.  I'll set MWP and I have not started that change in flux-core if you want to put up a PR.

PR #397, opened 2023-09-02
rfc8: drop outdated document

garlick:
Problem: RFC 8 is an early design document that is no longer very useful, and in fact could be confusing.

Drop the document.
Remove references from other documents.


grondo (approved):
LGTM, Thanks!

(The tests are calling out one more spec_8 reference in toctree)

garlick:
Thanks!  Fixed and setting MWP.

PR #398, opened 2023-09-08
rfc23: note default fsd is in seconds

chu11:
Problem: In commit

7eca38c882c8acc6b8e0a5c61cbab41a548060af

a list of suffixes was converted into a table.  This conversion accidentally dropped a note indicating that the default units of FSD are in seconds.

Solution: Add a sentence after the suffix table indicating the default is in seconds if no suffix is specified.


grondo (approved):
Good clarification, thanks!

PR #399, opened 2023-10-18
 rfc13: fix formatting of C API docs

garlick:
Problem: PMI-1 API documentation formats function parameters as literals, rendered in red monospace.
    
Use `:c:var:` instead.
    
Fix up a couple of other minor formatting problems along the way.

### DIFF
--- a/spec_13.rst
+++ b/spec_13.rst
@@ -72,8 +72,7 @@ protocol versions. Apart from version negotiation and the common
 fundamentals, PMI version 2 a different protocol and not covered here.
 
 PMIx ("x" for exascale, from the OpenMPI community) is a separate effort
-that provides compatible PMI version 1 and 2 APIs but uses an incompatible
-wire protocol. PMIx is not covered here.
+that is not covered here.
 
 PMIX ("X" for extension), is as set of extensions to PMI-2. The PMIX
 extensions are not covered here.
--- a/spec_13.rst
+++ b/spec_13.rst
@@ -74,7 +74,7 @@ fundamentals, PMI version 2 a different protocol and not covered here.
 PMIx ("x" for exascale, from the OpenMPI community) is a separate effort
 that is not covered here.
 
-PMIX ("X" for extension), is as set of extensions to PMI-2. The PMIX
+PMIX ("X" for extension), is a set of extensions to PMI-2. The PMIX
 extensions are not covered here.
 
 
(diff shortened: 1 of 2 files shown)


spec_13.rst
```diff
@@ -72,8 +72,7 @@ protocol versions. Apart from version negotiation and the common
 fundamentals, PMI version 2 a different protocol and not covered here.
 
 PMIx ("x" for exascale, from the OpenMPI community) is a separate effort
-that provides compatible PMI version 1 and 2 APIs but uses an incompatible
-wire protocol. PMIx is not covered here.
+that is not covered here.
```
grondo, line 75: Unrelated to this PR, but I noticed a typo below this change:

> PMIX ("X" for extension), is _as_ set ...

Should be "is _a_ set"?

spec_13.rst
```diff
@@ -274,7 +274,7 @@ Initialization
 .. c:function:: int PMI_Init (int *spawned)
 
 Initialize the PMI library for this process. Upon success, the value
-of ``spawned`` (boolean) SHALL bet set to (1) if this process was created
+of :c:var:`spawned` (boolean) SHALL bet set to (1) if this process was created
 by :c:func:`PMI_Spawn_multiple`, or (0) if not.
```
grondo, line 278: what is the difference between, `:func:` and `:c:func:` used here? (Above `PMI_Spawn_multiple` is referenced with plain `:func:`).

grondo (approved):
Thanks! Nice improvement (I didn't know about `:c:var` and `:func:`, I bet there are lots of other docs that could use this update).

spec_13.rst
garlick: I can't see any difference in the rendering but we should be consistent. Let me see if I can dig a little deeper in the docs and see what the default domain (correct term?) is for `:func:`.

grondo: Oh, good find. You know, I looked around for a bit and had trouble finding the correct documentation for this. I wonder if there's somewhere we should keep a reference to RFC best practices or something? Another RFC? :shrug:

(maybe it hasn't risen to that level yet, but I find reStructuredText and sphinx to be fairly complicated and usually just refer to existing examples when writing docs, so apologies in advance when I mess this up in the future)

garlick: OK to set MWP?

PR #401, opened 2023-11-08
rfc40: pull fluxion scheduling key def out of rfc20 and reformat rfc20

garlick:
Problem: the detailed description of Fluxion's scheduling key format is inappropriate in RFC 20 which describes the general Rv1 format.

Pull it out to a new RFC 40.

Also, improve the markup in RFC 20 to make it more readable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ Table of Contents
 - [37/File Archive Format](spec_37.rst)
 - [38/Flux Security Key Value Encoding](spec_38.rst)
 - [39/Flux Security Signature](spec_39.rst)
+- [40/Fluxion Resource Set Extension](spec_40.rst)
 
 Build Instructions
 ------------------
--- a/index.rst
+++ b/index.rst
@@ -260,6 +260,13 @@ for a series of typed key-value pairs.
 The Flux Security Signature is a NUL terminated string that represents
 content secured with a digital signature.
 
+:doc:`40/Fluxion Resource Set Extension <spec_39>`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This specification defines the data format used by the Fluxion scheduler
+to store resource graph data in RFC 20 *R* version 1 objects.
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -49,6 +49,7 @@ Table of Contents
 - [37/File Archive Format](spec_37.rst)
 - [38/Flux Security Key Value Encoding](spec_38.rst)
 - [39/Flux Security Signature](spec_39.rst)
+- [40/Fluxion Resource Set Extension](spec_40.rst)
```
grondo, line 52: Commit message nit: capitalization of Fluxion is inconsistent.

grondo (approved):
Oh, that turned out nice! Thanks!

garlick:
Thanks - fixed that and also a heading level problem that was messing up the table of contents tab.

Note the `attributes` entry in RFC 20.  If you remember, at one point we observed that Fluxion was setting the queue name in `attributes.system.scheduler`.  I'm not seeing that now in my testing but I also didn't immediately find a Fluxion commit message that says anything about dropping that.  For now I left it, but we might want to revisit at some point.  I'm not sure we even want/need the whole `attributes` section?

grondo:
>  For now I left it, but we might want to revisit at some point. I'm not sure we even want/need the whole `attributes` section?

Yeah, I was wondering the same thing on a read-through. It seems like if it is necessary for Fluxion it should go in the opaque `scheduling` key. It's utility may have been obviated, though, now that we have `jobspec-update` events which add a default queue to Flux's copy of jobspec.

There is still code in Fluxion to emit the attributes section in the RV1 "writer" here:

https://github.com/flux-framework/flux-sched/blob/22088217e0592ed2380b0a82e7c45e93daf50f0a/resource/writers/match_writers.cpp#L941-972

But I don't know when it is used.

grondo:
Those attributes seem to be required by the `t1009-recovery-multiqueue.t` test in Fluxion, but it seems like there's a disconnect in Fluxion's concept of queues and that supported by flux-core.

PR #402, opened 2023-11-08
rfc20: drop the attributes section

garlick:
Problem: the Rv1 attributes section is used by the early Fluxion queue implementation to store the assigned queue name in the system.scheduler key, but queues are now handled differently per RFC 33.

Drop the attributes section as discussed in #401

### DIFF
--- a/spec_20.rst
+++ b/spec_20.rst
@@ -106,16 +106,13 @@ Resource Set Format Definition
 
 The JSON documents that conform to the *R* format SHALL be referred
 to as *R* JSON documents or in short *R* documents.
-An *R* JSON document SHALL consist of a dictionary with four
-keys: :data:`version`, :data:`execution`, :data:`scheduling` and
-:data:`attributes`.  It SHALL be valid if and only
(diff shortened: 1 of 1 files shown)


grondo:
In case it wasn't clear in #401, this section is still required by Fluxion for multi-queue restart (on restart presumably Fluxion does not have access to jobspec to get the queue name, which for some reason it needs). I'll open an issue in flux-sched and refer to this PR.

garlick:
Yes that is what I was thinking.  Move RFCs forward, let code catch up.  And it should have access to the jobspec...oh maybe you meant in the place where it's using the queue name.

grondo:
FYI: https://github.com/flux-framework/flux-sched/issues/1108

grondo:
>  And it should have access to the jobspec...oh maybe you meant in the place where it's using the queue name.

Yeah, I'm guessing this is in the scheduler "hello" protocol.  I think only R objects are returned by this protocol. Stashing the queue name in R may have been an optimization to avoid fetching the jobspec for every R? (I actually don't know, just guessing)

PR #403, opened 2023-11-09
rfc20: rework content

garlick:
RFC 20 contained a lot of prose that didn't really add clarity, and missed explicitly mentioning some requirements.

Rework the RFC so that it's more useful as a quick reference and includes clarifying points that were missing.

### DIFF
--- a/spec_20.rst
+++ b/spec_20.rst
@@ -111,80 +111,77 @@ Design Goals
 Implementation
 **************
 
-Resource Set Format Definition
-==============================
+.. note::
 
-The JSON documents that conform to the *R* format SHALL be referred
-to as *R* JSON documents or in short *R* documents.
-An *R* JSON document SHALL consist of a dictionary with four
-keys: :data:`version`, :data:`execution`, :data:`scheduling` and
-:data:`attributes`.  It SHALL be valid if and only
-if it contains the :data:`version` key and either or both the :data:`execution`
-and :data:`scheduling` keys. The value of the :data:`execution` key SHALL
-contain sufficient data for the execution system to perform its core tasks. The
-value of :data:`scheduling` SHALL contain sufficient data for schedulers.
-Finally, the value of :data:`attributes` SHALL provide optional information
-including but not being limited to data specific to the scheduler used to
-create this JSON document.
+  In Flux documentation, the terms "node rank" and "execution target" are
+  sometimes used interchangeably to refer to the Flux broker rank used to
+  launch tasks on a given resource set.  When Flux launches a new Flux
+  instance on a subset of resources, the new broker ranks do not match those
+  of the the enclosing instance and the inherited *R* must be re-ranked
+  before it can be brought into inventory.  Therefore, to avoid implying
+  that a node rank uniquely identifies a physical node across instances,
+  execution target is preferred in this document.
+
+R Format
+========
+
+*R* SHALL be defined as a JSON dictionary with the following keys:
 
 .. data:: version
 
-  The value of the :data:`version` key SHALL contain 1 to indicate
-  the format version.
+  (*integer*, REQUIRED) The *R* specification version.
+
+  For this specification the value is always 1.
 
 .. data:: execution
 
-  The value of the :data:`execution` key SHALL contain at least the keys
-  :data:`R_lite`, and :data:`nodelist`, with optional keys :data:`properties`,
-  :data:`starttime` and :data:`expiration`. Other keys are reserved for future
-  extensions.
+  (*dictionary*, REQUIRED) The resource set.  It SHALL have following keys:
 
   .. data:: R_lite
 
-    :data:`R_lite` is a strict list of dictionaries each of which SHALL contain
-    at least the :data:`rank` and :data:`children` keys.
+    (*array of dictionary*, REQUIRED) A list that identifies one or more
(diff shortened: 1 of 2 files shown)


spec_20.rst
```diff
@@ -41,6 +41,8 @@ Related Standards
 
 -  :doc:`29/Hostlist Format <spec_29>`
 
+-  :doc:`31/Job Constraints Specification <spec_31>`
+
```
grondo, line 45: Was this line meant for a different commit?

spec_20.rst
```diff
@@ -111,80 +111,76 @@ Design Goals
...
-  :data:`R_lite`, and :data:`nodelist`, with optional keys :data:`properties`,
-  :data:`starttime` and :data:`expiration`. Other keys are reserved for future
-  extensions.
+  (*dictionary*, REQUIRED) The resource set.  It SHALL have following keys:
 
   .. data:: R_lite
 
-    :data:`R_lite` is a strict list of dictionaries each of which SHALL contain
-    at least the :data:`rank` and :data:`children` keys.
+    (*array of dictionary*, REQUIRED) A list that identifies one or more nodes
```
grondo, line 142: Based on the added note above, should "node" here be replaced with "execution target", or is execution target only meant to be a synonym for "node rank"? Either way I think it's fine.

spec_20.rst
```diff
@@ -111,80 +111,76 @@ Design Goals
...
+      (*string*, REQUIRED) An RFC 22 idset representing one or more execution
+      targets that contain the resources in :data:`children`.
 
     .. data:: children
 
-      The :data:`children` key encodes the information about certain compute
-      resources contained within this compute node. The value of this key SHALL
-      contain a dictionary with two keys: :data:`core` and :data:`gpu`. Other
-      keys are reserved for future extensions.
+      (*dictionary*, REQUIRED) It SHALL have the following keys:
```
grondo, line 153: Just a touch more description than "It" could be useful here, though I know you are trying to keep the document terse. E.g. maybe "The :data:`children` key identifies the resources assigned to execution targets encoded in :data:`rank`" (too much?)

spec_20.rst
```diff
@@ -111,80 +111,76 @@ Design Goals
...
   .. data:: properties
 
-    The optional :data:`properties` key SHALL be a dictionary where each key
-    maps a single property name to a RFC 22 idset string. The idset string SHALL
-    represent a set of execution target ranks. A given execution target
-    rank MAY appear in multiple property mappings. Property names SHALL
-    be valid UTF-8, and MUST NOT contain the following illegal characters::
+    (*dictionary*, OPTIONAL) Each key
+    maps a single property name to a RFC 22 idset string. The idset string
+    SHALL represent a set of execution targets. A given node rank MAY appear in
```
grondo, line 181: Should "node rank" just be "rank" or "target" (again, going back to the note that says "execution target" will be used throughout the document, though I actually think what is here now is find if you'd like to leave it alone)

spec_20.rst
```diff
@@ -201,30 +197,39 @@ create this JSON document.
...
+    (*number*, OPTIONAL) The end or expiration time of the resource set,
+    after which it becomes invalid.
+
+    The value SHALL be expressed as the number of seconds elapsed since the
+    Unix Epoch (1970-01-01 00:00:00 UTC) with optional microsecond precision.
+
+    If :data:`starttime` is also set, :data:`expiration` MUST be greater than
+    :data:`starttime`.
+
+    If :data:`expiration` is unset, the resource set has no specified end time.
```
grondo, line 219: While not called out here, I think `0.` is valid for both `starttime` and `expiration` and is equivalent to unset. An important point is that `expiration` can be less than `starttime` if it is `0.`.

grondo (approved):
Nice improvement! Just noticed some minor things (questions really) on a review.

Probably the only thing that needs to be changed is to document the meaning of `0.` for `starttime` and `expiration`. (Something that appears to have been missing before as well)

spec_20.rst
garlick: Oh yeah, "meant" might be a strong word but yes it should be split out :-)

spec_20.rst
garlick: Yeah I think you're right, will fix.

garlick: What do you think about

> A list that identifies one or more execution targets and the specific cores and GPUs they control.

grondo: That sounds great to me. (I assume this was meant for the other comment though)

garlick: It kind of overlaps what's in the text mentioned in the other comment.  I'll try to tune them both in one fixup and you can have a look at the flow.

grondo: Ok, I see that now (wasn't apparent just reading the diff).

spec_20.rst
garlick: OK I added
> A value of `0.` SHALL be interpreted as "unset".

to both keys.  I think it all works out then because the bit about "must be greater than" says to "if set".

garlick:
OK, made those changes. I left the changes to the implementation section broken out as fixups since I thought it might easier to see what changed.  The other changes were squashed.

grondo (approved):
Great! LGTM!

garlick:
Thanks!  Squashed the fixups and I'll set MWP.

PR #404, opened 2023-11-30
misc cleanup

garlick:
This cleans up a few long standing issues

- fix inconsistent use of heading levels
- move repeated Language section to an include file
- let sphinx x-ref fill in titles for referenced RFCs instead of restating the titles
- make the Name/Editor/State block pretty with a table


grondo (approved):
Nice cleanup!

PR #406, opened 2023-12-19
rfc42: add subprocess protocol RFC

garlick:
Problem: the subprocess execution protocol is not documented.

Add a new RFC.

(Marking as a WIP since I think I need to make another pass double-checking this against code before unleashing it on others)


vsoch:
oh neat! Is this what I was doing with `flux exec` (and wasn't aware what it was called)?

garlick:
Yep :-)

garlick:
I've made a pass checking this and it seems OK so removing WIP.

grondo (approved):
LGTM!

PR #407, opened 2023-12-20
rfc41: add new lookup flags

chu11:
Problem: The job-info.lookup RPC is missing new flags that are supported.

Document the new flags.

---

WIP as we discuss https://github.com/flux-framework/flux-core/pull/5633 ("update" vs "updated" as flag name?)

also .. this specific spec change assumes https://github.com/flux-framework/flux-core/pull/5635 would go in too.  Could break into two commits.

### DIFF
--- a/spec_41.rst
+++ b/spec_41.rst
@@ -127,7 +127,18 @@ The RPC payloads are defined as follows:
 
   .. object:: flags
 
-    (*integer*, REQUIRED) Flags, reserved for future use.  Set to zero.
+    (*integer*, REQUIRED) A bitfield comprised of zero or more flags:
+
+    json_decode (1)
+      For lookups of R or jobspec, return the field as a decoded
+      JSON object instead of a string.  This flag has no effect on
+      other keys.
+
+    current (2)
+      For lookups of R or jobspec, return the current version. The
+      current version SHALL be computed by applying any
+      resource-update or jobspec-update events that have been posted
+      to the job eventlog, as described in RFC 21.
 
 .. object:: job-info.lookup response
 
@@ -141,7 +152,9 @@ The RPC payloads are defined as follows:
 
     Additional keys correspond to the keys in the request.
 
-    Values are the KVS values associated with the keys, encoded as strings.
+    Values are the KVS values associated with the keys.  Values are
+    encoded as strings, except in special cases indicated by flags
+    in the request.
 
 Eventlog Watch
 ==============
--- a/spell.en.pws
+++ b/spell.en.pws
@@ -489,3 +489,4 @@ userids
 parsable
 bitmasks
 DoS
+lookups


spec_41.rst
```diff
@@ -127,7 +127,15 @@ The RPC payloads are defined as follows:
...
-    (*integer*, REQUIRED) Flags, reserved for future use.  Set to zero.
+    (*integer*, REQUIRED) A bitfield comprised of zero or more flags:
+
+    decode (1)
+      For lookups of R or jobspec, return the field as a decoded
+      JSON object instead of a string.
+
+    update (2)
+      For lookups of R or jobspec, return an updated version which
+      has applied updates from the eventlog.
```
garlick, line 138: Suggestion: rename `update` to `current`.

Also, make it clear that the flag has no effect on other keys as opposed to causing an error.

spec_41.rst
```diff
@@ -127,7 +127,15 @@ The RPC payloads are defined as follows:
 
   .. object:: flags
 
-    (*integer*, REQUIRED) Flags, reserved for future use.  Set to zero.
+    (*integer*, REQUIRED) A bitfield comprised of zero or more flags:
+
+    decode (1)
+      For lookups of R or jobspec, return the field as a decoded
+      JSON object instead of a string.
```
garlick, line 134: The response payload states that all values are returned as strings so it needs an update.

Maybe it would be slightly clearer to say
> return the value as a JSON object instead of a JSON object encoded as a JSON string.

The flag name `decode` feels a bit ambiguous.  What about `objectify`? :shrug:

grondo: Would `json` be explicit enough?

garlick: but a JSON object and a JSON object encoded as a JSON string are both JSON?

Gosh I hate naming things! :laughing:

grondo: Hm, it seemed clear to me, but if we need to be pedantic, `decode` or `json_decode` is probably better than `objectify`, for which the primary definition has negative connotations.

garlick: I could live with `json` on second thought.  I didn't mean to be pedantic - just trying to come up with an intuitive name and failing.  I don't think it matters that much.  How about `json`?

grondo: I also meant to say as long as it is clearly documented, the actual flag name isn't too important to me either.

chu11: I like `json_decode`, i dunno about just `json` ... because the returned RPC object is already json?

garlick: Works for me I guess.

chu11:
re-pushed, going with `json_decode` and `current`.

spec_41.rst
```diff
@@ -127,7 +127,17 @@ The RPC payloads are defined as follows:
...
+
+    json_decode (1)
+      For lookups of R or jobspec, return the field as a decoded
+      JSON object instead of a string.  This flag has no effect on
+      other keys.
+
+    current (2)
+      For lookups of R or jobspec, return the current version which
+      has applied respective updates from the eventlog.  This flag has
+      no effect on other keys.
```
garlick, line 140: Maybe the eventlog events ought to be explicitly called out?  E.g. say
> For lookups of R or jobspec, return the current version.  The current version SHALL be computed by applying any resource-update  or jobspec-update events that have been posted to the job `eventlog`, as described in RFC 21.

Still need to update the response description, which states that all returned values are strings.

chu11:
re-pushed, with some tweaks per comments above

chu11:
removed WIP, I think this RFC and the associated PR https://github.com/flux-framework/flux-core/pull/5633 are ready for review.

grondo (approved):
This ready to go in? Seems good to me.

PR #408, opened 2024-01-10
rfc43: add new RFC for job list service

chu11:
Problem: job-list services are not documented.

Add an RFC to document current RPCs.

---

first run through, I might have some rst-isms wrong. But mostly went back and forth on various wording and various layout.  Lets see what initial reaction is :P

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -278,6 +278,10 @@ information for guest users.
 The subprocess server protocol is used for execution, monitoring, and
 standard I/O management of remote processes.
 
+:doc:`spec_43`
+~~~~~~~~~~~~~~
+
+The Flux Job List Service provides read-only summary information for jobs.
 
 .. Each file must appear in a toctree
 .. toctree::
@@ -323,3 +327,4 @@ standard I/O management of remote processes.
    spec_40
    spec_41
    spec_42
+   spec_43
--- a/spec_43.rst
+++ b/spec_43.rst
@@ -1,0 +1,393 @@
+.. github display
+   GitHub is NOT the preferred viewer for this file. Please visit
+   https://flux-framework.rtfd.io/projects/flux-rfc/en/latest/spec_43.html
+
+43/Job List Service
+###################
+
+The Flux Job List Service provides read-only summary information for jobs.
+
+.. list-table::
+  :widths: 25 75
+
+  * - **Name**
+    - github.com/flux-framework/rfc/spec_43.rst
+  * - **Editor**
+    - Albert Chu <chu11@llnl.gov>
+  * - **State**
+    - raw
+
+Language
+********
+
+.. include:: common/language.rst
+
+Related Standards
+*****************
+
+- :doc:`spec_18`
+- :doc:`spec_20`
+- :doc:`spec_21`
+- :doc:`spec_25`
+- :doc:`spec_26`
+- :doc:`spec_27`
+- :doc:`spec_29`
+- :doc:`spec_31`
+- :doc:`spec_41`
+
+Background
+**********
+
+The job list service provides a simple job query service that collates
+information from several sources, including the job info service
+described in RFC 41.  Some use cases include:
+
+- See which jobs are pending, running, or inactive
+- See what jobs are running on specific nodes
+- Get general information about a job, such as a job's exit code
+- See the order in which jobs were submitted
+- See how many jobs are pending in the queue before a specific one
+
+Goals
+*****
+
+- Provide read-only access to job information
+
+- limit guest access to sensitive job information such as job stdout.
+
+- Hide the complexity of parsing or collating data from multiple sources for commonly accessed information.
+
+- Provide ways to query jobs callers are interested in.
+
+Implementation
+**************
+
+The job list service SHALL provide callers the ability to request specific job attributes.  See :ref:`spec_43_job_attributes` below for details.
+
+The job list service SHALL provide a RFC 31 constraint syntax for filtering jobs.  See :ref:`spec_43_constraint_operators` below for details.
+
+.. _spec_43_job_attributes:
+
+Job Attributes
+==============
+
+The job list service SHALL support the following attributes.
+
+.. list-table::
+   :header-rows: 1
+
+   * - Attribute
+     - Description
+     - Value Encoding
+   * - ``id``
+     - job id
+     - integer
+   * - ``userid``
+     - userid of job submitter
+     - integer
+   * - ``urgency``
+     - job urgency
+     - integer
+   * - ``priority``
+     - job priority
+     - integer
+   * - ``t_submit``
+     - time job was submitted
+     - real
+   * - ``t_depend``
+     - time job entered depend state
+     - real
+   * - ``t_run``
+     - time job entered run state
+     - real
+   * - ``t_cleanup``
+     - time job entered cleanup state
+     - real
+   * - ``t_inactive``
+     - time job entered inactive state
+     - real
+   * - ``state``
+     - current job state
+     - integer
+   * - ``name``
+     - job name
+     - string
+   * - ``cwd``
+     - job current working directory
+     - string
+   * - ``queue``
+     - job queue
+     - string
+   * - ``project``
+     - job project
+     - string
+   * - ``bank``
+     - job bank
+     - string
+   * - ``ntasks``
+     - job task count
+     - integer
+   * - ``ncores``
+     - job core count
+     - integer
+   * - ``nnodes``
+     - job node count
+     - integer
+   * - ``ranks``
+     - ranks a job ran on
+     - integer
+   * - ``nodelist``
+     - nodes assigned to a job in RFC 29 Hostlist format
+     - string
+   * - ``duration``
+     - job duration in seconds
+     - real
+   * - ``expiration``
+     - time job was marked to expire
+     - real
+   * - ``success``
+     - true if job was successful
+     - boolean
+   * - ``result``
+     - integer indicating job success or failure type
+     - integer
+   * - ``waitstatus``
+     - status of job as returned by waitpid(2)
+     - integer
+   * - ``exception_occurred``
+     - true if exception occurred
+     - boolean
+   * - ``exception_type``
+     - if exception occurred, exception type
+     - string
+   * - ``exception_severity``
+     - if exception occurred, exception severity
+     - integer
+   * - ``exception_note``
+     - if exception occurred, exception note
+     - string
+   * - ``annotations``
+     - annotations as described in RFC 27
+     - object
+   * - ``dependencies``
+     - current job dependencies
+     - array of string
+
+Job attributes SHALL be returned via an object where the keys are the requested job attributes.  The values are the attribute values, each encoded as described in the above table.
+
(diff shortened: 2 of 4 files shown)


garlick:
Nice!  Very good to get this protocol documented IMHO.   Just a couple of quick thoughts on a skim:
-  a table might be better presentation for attributes and/or constraint operators
- no need to require the service to be loaded on rank 0.  perhaps a future implementation using the same protocol will have distributed caching of some sort.
- I'd reference the job info RFC when you mention the job info service

spec_43.rst
```diff
@@ -0,0 +1,290 @@
...
+   and match jobs with those job names.
+
+ - ``queue``: The set of *string* values SHALL designate one or more queues
+   and match jobs with those job queues.
+
+ - ``states``: The set of *integer* values SHALL designate one or more job states
+   and match jobs in those job states.
+
+ - ``results``: The set of *integer* values SHALL designate one or more job results
+   and match jobs with those results.
```
grondo, line 147: Question: Can the `states` and `results` values be a bitmask of states to match? I.e. is `{"states": [48, 14]}` equivalent to `{"states": [62]}`? If so, it might be useful to call that out specifically here.

chu11: yes!  and i forgot that string inputs are ok too (i.e. "depend")

chu11:
re-pushed based on comments above, minor tweaks forthcoming once i see how it looks on read the docs

chu11:
removed WIP, I think it's at an ok point to read with more seriousness :-)

index.rst
```diff
@@ -278,6 +278,12 @@ information for guest users.
 The subprocess server protocol is used for execution, monitoring, and
 standard I/O management of remote processes.
 
+:doc:`spec_43`
+~~~~~~~~~~~~~~
+
+The Flux Job List Service provides summary information for jobs in the
+system.  It provides read-only access.  Several ways to find / filter
+jobs is also supported.
```
grondo, line 286: Could simplify to "The Flux Job List Service provides read-only summary information for jobs."

grondo: Also, IMO the slash (`/`) should be avoided in the more formal prose of RFCs. An "and" could work fine here.

https://style.mla.org/slashed-terms-mla-style/

spec_43.rst
```diff
@@ -0,0 +1,362 @@
+.. github display
+   GitHub is NOT the preferred viewer for this file. Please visit
+   https://flux-framework.rtfd.io/projects/flux-rfc/en/latest/spec_41.html
+
+43/Job List Service
+###################
+
+The Flux Job List Service provides summary information for jobs in the
+system.  It provides read-only access.  Several ways to find / filter
+jobs is also supported.
```
grondo, line 10: See above comment.

spec_43.rst
```diff
@@ -0,0 +1,362 @@
...
+- :doc:`spec_27`
+- :doc:`spec_29`
+- :doc:`spec_31`
+- :doc:`spec_41`
+
+Background
+**********
+
+Users are interested in seeing jobs that have been submitted to the
+scheduler.  Some reason may include:
```
grondo, line 44: Minor, but jobs are technically submitted to the job manager not the scheduler (a scheduler need not be loaded to submit and query or list jobs). 

Also "Some reason may include:" :arrow_right: "Some reasons may include"

spec_43.rst
```diff
@@ -0,0 +1,362 @@
...
+- See which jobs are pending, running, or inactive
+- See what jobs are running on specific nodes
+- Get general information about a job, such as a job's exit code
+- See the order in which jobs were submitted
+- See how many jobs are pending in the queue before a specific one
+
+While the job info service described in RFC41 is capable of providing job owners information about their
+own jobs, it has several limitations:
+
+- job information may not be easily parsed / collated from multiple sources into one easily parsable format
```
grondo, line 55: Again, I'd replace `/` with `or` here.

spec_43.rst
```diff
@@ -0,0 +1,362 @@
...
+- information from multiple jobs is not collated into a simple to parse list
+- information about non-owned jobs is not available
```
grondo, line 57: Suggestion: something like?
```suggestion
- information about the jobs of other users in a multi-user instance is not available
```

spec_43.rst
```diff
@@ -0,0 +1,362 @@
...
+
+Goals
+*****
+
+- Provide read-only access to non-sensitive information for all jobs.
+
+- Hide the complexity of parsing or collating data from multiple sources for commonly accessed information.
+
+- Provide ways to find and/or filter jobs callers are interested in.
```
grondo, line 66: The difference between find and filter may not be clear. Maybe replace "find and/or filter" with just "filter" or "query"? I don't have a strong opinion on this one though

spec_43.rst
```diff
@@ -0,0 +1,362 @@
...
+     - job core count
+     - integer
+   * - nnodes
+     - job node count
+     - integer
+   * - ranks
+     - ranks a job ran on
+     - integer
+   * - nodelist
+     - nodes a job ran on, may accept RFC29 Hostlist
```
grondo, line 144:
```suggestion
     - nodes assigned to a job in RFC29 Hostlist format
```
Unless I'm misunderstanding what is meant by "may accept" here (maybe this was meant to describe the constraint not the attribute?)

spec_43.rst
```diff
@@ -0,0 +1,362 @@
...
+    Designate one or more queues (*string*) and match jobs submitted to those job queues.
+
+``states``
+    Designate one or more job states (*string* or *integer*) and match jobs in those job states.  Both bitmasks (including multiple states) and string names of the states SHALL be accepted.
+
+``results``
+    Designate one or more job results (*string* or *integer*) and match jobs with those results.  Both bitmasks (including multiple results) and string names of the results SHALL be accepted.
+
+``t_submit``, ``t_depend``, ``t_run``, ``t_cleanup``, ``t_inactive``
+    Designate one timestamp with a prefixed comparison operator (*string*).  The accepted comparison operators SHALL be `>`, `<`, `>=`, and `<=`, for greater than, less than, greater than or equal, or less than or equal.  Match jobs where the respective timestamp matches against the submitted timestamp and comparison operator.
```
grondo, line 216: Suggestion: 
> Match jobs where the respective timestamp matches against the submitted timestamp and comparison operator.

Maybe reword with RFC syntax ("A timestamp operator SHALL match...") and replace "submitted timestamp" (could be confused with t_submit or job submission) with "provided timestamp"?

grondo (commented):
Very nice! First pass up through the Constraint Operators section. Just some minor suggestions so far.

spec_43.rst
chu11: ahhh yeah I messed up meaning the constraint input

chu11:
thanks @grondo, re-pushed with some fixups based on your comments.  Also fixed a few minor English issues.

spec_43.rst
```diff
@@ -0,0 +1,360 @@
...
+    Designate one or more queues (*string*) and match jobs submitted to those job queues.
+
+``states``
+    Designate one or more job states (*string* or *integer*) and match jobs in those job states.  Both bitmasks (including multiple states) and string names of the states SHALL be accepted.
+
+``results``
+    Designate one or more job results (*string* or *integer*) and match jobs with those results.  Both bitmasks (including multiple results) and string names of the results SHALL be accepted.
+
+``t_submit``, ``t_depend``, ``t_run``, ``t_cleanup``, ``t_inactive``
+    Designate one timestamp with a prefixed comparison operator (*string*).  The accepted comparison operators SHALL be `>`, `<`, `>=`, and `<=`, for greater than, less than, greater than or equal, or less than or equal. A timestamp operator SHALL match jobs where the respective timestamp matches against the provided timestamp.
```
grondo, line 214: Suggestion: "Designate one timestamp with an _optional_ prefixed comparison operator" ?

grondo (approved):
LGTM! Just one more suggestion. Not sure if @garlick wants to have another pass too.

spec_43.rst
chu11: actually it's not optional ... I think we decided to go with that to avoid confusion of "what is the default?"

garlick:
Yes I'll have a run through, thanks :-)

garlick (approved):
Looks good - thanks for doing this!

I just had a few cosmetic comments that aren't super important.  Up to you what you want to take or leave.

spec_43.rst
```diff
@@ -0,0 +1,360 @@
...
+- Provide read-only access to non-sensitive information for all jobs.
+
+- Hide the complexity of parsing or collating data from multiple sources for commonly accessed information.
+
+- Provide ways to query jobs callers are interested in.
+
+Implementation
+**************
+
+The job list service SHALL provide callers the ability to read job information via identifier keys, which will be called *attributes*.  See `Job Attributes` below for details.
```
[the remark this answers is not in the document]
chu11: Hmmm I tried the section references earlier but it didn’t work.   Lemme try again.

spec_43.rst
```diff
@@ -0,0 +1,360 @@
...
+
+The job list service SHALL provide a RFC31 constraint syntax for filtering jobs.  See `Constraint Operators` below for details.
+
+Job Attributes
+==============
+
+Job information is defined by the following *attribute* keys listed below.
```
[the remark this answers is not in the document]
chu11: ehhh, it feels empty without something before the table.  Perhaps a simpler sentence: `The job list service SHALL support the following attributes.`

spec_43.rst
```diff
@@ -0,0 +1,360 @@
...
+- :doc:`spec_27`
+- :doc:`spec_29`
+- :doc:`spec_31`
+- :doc:`spec_41`
+
+Background
+**********
+
+Users are interested in seeing jobs that have been submitted to the
+job-manager.  Some reasons may include:
```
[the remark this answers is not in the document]
chu11: Sounds good, although I think your wording suggests this is a complete set of use cases.   I used the phrase "Some use cases include:" instead.

chu11:
re-pushed with tweaks per recommendations above.  possible tweaks coming depending on how the final output looks. (oops, I see `make check` already failed)

chu11:
fixed a few minor typos, but overall looks good to me if you guys wanna take one more skim

garlick:
It's looking good!  So what do we do with attributes that are not available, just omit them or set the attribute to NULL or?  Might be good to clarify that.

The only other thing I can think of that might be helpful is to show a JSON example of the list request and response like we've done in some of the other RFCs.

Great work!

chu11:
> It's looking good! So what do we do with attributes that are not available, just omit them or set the attribute to NULL or? Might be good to clarify that.

Yup, the text I put in is "If an attribute has not been set for a job, it SHALL NOT be returned in the returned data object."

> The only other thing I can think of that might be helpful is to show a JSON example of the list request and response like we've done in some of the other RFCs.

Good idea, I'll stick one in at the end.

chu11:
repushed, stuck an example at the end of the rfc, just one for `job-list.list`, didn't think necessary to have one example for each rpc

garlick:
Looks good!

PR #410, opened 2024-01-16
rfc43: support hostlist constraint

chu11:
Problem: It would be useful to fitler jobs based on the nodes assigned to the job.

Support a "hostlist" constraint to filter jobs by specific hosts.

----

per https://github.com/flux-framework/flux-core/pull/5656

The idea to make max hostlist input size of instance OR 1024 is there is potential for the size of the cluster to change over time (like EOL and cluster size shrinking).  So giving it a minimum of 1024 is to give it atleast an ok minimum no matter what.


chu11:
closing, i believe a "comparison limit" in constraints overall is a better solution

spec_43.rst
```diff
@@ -209,6 +209,9 @@ based on the following constraint operators.
 ``results``
     Designate one or more job results (*string* or *integer*) and match jobs with those results.  Both bitmasks (including multiple results) and string names of the results SHALL be accepted.
 
+``hostlist``
+    Designate one or more nodes in RFC 29 Hostlist format (*string*) and match jobs assigned to those nodes.  The number of nodes specified in the ``hostlist`` constraint SHALL NOT exceed the size of the instance or 1024, whichever is larger.
```
grondo, line 213: Instead of specifying a precise limit here, maybe couch this by saying the "job list module MAY limit the number of entries in a hostlist constraint to prevent long constraint match times" or similar. 

The 1024 "minimum maximum" makes some sense, but since it doesn't handle instances larger than 1024 that shrink, I'm not sure that specific number should be encoded in RFC. Also, maybe when there's a backend DB there will be more efficient hostlist matching and a limit won't be necessary.

chu11: > The 1024 "minimum maximum" makes some sense, but since it doesn't handle instances larger than 1024 that shrink, I'm not sure that specific number should be encoded in RFC.

Understood.  I picked 1024 somewhat randomly as a "reasonable minimum maximum".  But agree, we could leave the number out.

chu11:
re-pushed, tweaking per comment above

spec_43.rst
```diff
@@ -209,6 +209,9 @@ based on the following constraint operators.
 ``results``
     Designate one or more job results (*string* or *integer*) and match jobs with those results.  Both bitmasks (including multiple results) and string names of the results SHALL be accepted.
 
+``hostlist``
+    Designate one or more nodes in RFC 29 Hostlist format (*string*) and match jobs assigned to those nodes.  The job list module MAY limit the number of entries in a hostlist constraint to prevent long constraint match times
```
grondo, line 213: nit: missing period.

grondo (approved):
LGTM.

PR #412, opened 2024-03-11
rfc27: add R to sched.free request

garlick:
Problem: flux-framework/flux-core#5783 adds R to the sched.free request to avoid an extra lookup.

Add R to the sched.free request payload definition in RFC 27.


garlick:
What the heck?
```
2024-03-11T18:01:45.6831528Z spec_43.rst:16: : Spell check: Chu: Albert Chu <.
```

grondo:
The sphinx spellcheck implementation is supposed to collect contributor names directly from git and automatically ignore them in the spellcheck. I wonder why it failed for Al's name?

garlick:
Ah I didn't know that was a thing. I did however see other contributor names in the pws file so I went and and added Al's.

grondo (approved):
Thanks! LGTM!

PR #413, opened 2024-03-11
rfc27: eliminate sched.free response

garlick:
Problem: sched.free requires a response, but this adds complexity to the job manager and has little benefit.

Code that drops the sched.free response from flux-core is proposed in flux-framework/flux-core#5786.

Drop the free response.


grondo (approved):
Looks good!

PR #414, opened 2024-05-13
rfc: add notification service design doc

wihobbs:
This is a RFC-style design document being put up to document the design we have for the flux job notification service.

I'm looking to get feedback on the clarity of this document and make sure we have as many edge cases as possible covered. If this isn't an appropriate RFC for the whole project it can be moved somewhere else, but the github PR interface provides a nice way for us to engage in a feedback loop. 

I wanted to add a flow chart detailing some of the interactions but am still working on that. I think there's enough here for a first pass review, though.

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -283,6 +283,12 @@ standard I/O management of remote processes.
 
 The Flux Job List Service provides read-only summary information for jobs.
 
+:doc:`spec_44`
+~~~~~~~~~~~~~~
+
+This specification describes the Flux service that allows users to
+receive external notifications for events in a Flux job.
+
 .. Each file must appear in a toctree
 .. toctree::
    :hidden:
@@ -328,3 +334,4 @@ The Flux Job List Service provides read-only summary information for jobs.
    spec_41
    spec_42
    spec_43
+   spec_44
--- a/spec_44.rst
+++ b/spec_44.rst
@@ -1,0 +1,204 @@
+.. github display
+   GitHub is NOT the preferred viewer for this file. Please visit
+   https://flux-framework.rtfd.io/projects/flux-rfc/en/latest/spec_28.html
+
+44/Flux Library for Adaptable Notifications Version 1
+###########################################################
+
+This specification describes the Flux service that allows users to
+receive external notifications for events in a Flux job.
+
+.. list-table::
+  :widths: 25 75
+
+  * - **Name**
+    - github.com/flux-framework/rfc/spec_44.rst
+  * - **Editor**
+    - William Hobbs <wihobbs@llnl.gov>
+  * - **State**
+    - raw
+
+Language
+********
+
+.. include:: common/language.rst
+
+Related Standards
+*****************
+
+- :doc:`spec_14`
+- :doc:`spec_21`
+- :doc:`spec_25`
+
+Background
+**********
+
+The Flux Library for Adaptable Notifications (FLAN) provides event-driven
+functionality that sends external notifications of job events. 
+
+Terminology
+***********
+
+These terms may have broader meaning in other RFCs or the Flux project. To
+avoid confusion, below is a glossary of terms as they apply in this document.
+
+Notification
+  An email or other notification triggered by FLAN but whose ultimate delivery
+  is handled by an external service.
+
+Notification-enabled jobs
+  Jobs that include a jobspec attribute requesting a notification for certain
+  events in the job's life cycle. For a more detailed definition of job events,
+  refer to :doc:`spec_21`.
+
+Requirements
+************
+
+ - By default, do not notify a user of any job events.
+ - Allow the user to override this default with a jobspec attribute,
+   ``system.notify``.
+ - Support notification after any event of the job, where events are defined in
+   :doc:`spec_21`.
+ - Support email for end user notification delivery.
+ - Allow for extensibility via plugins to support more end user notification 
+   delivery services, such as Slack and Mattermost. The implementation of
+   plugins for any service other than email is not a requirement.
+ - Utilize as few resources as possible in the Flux job-manager. Under no
+   circumstances will a notification block any stage or event of a Flux job.
+ - Provide configurable rate-limiting to ensure users can never be overwhelmed
+   by notifications.
+
+Implementation 
+**************
+
+FLAN SHALL be implemented by a service that MAY be started under ``systemd``.
+
+Introduction
+============
+
+The Flux job-manager journal of events (JoE) is an interface that streams job
+events in real-time for jobs in a Flux instance. The JoE can be configured to
+send all completed events in addition to streaming real-time events. The JoE
+includes annotations such as jobspec and R where appropriate. 
+
+The Flux Library for Adaptable Notifications (FLAN) provides a server that
+opens a streaming RPC request to the JoE, receives events from the JoE, stores
+jobspec, event logs, and resource sets by jobid for all active jobs, and allows
+for clients to asynchronously perform operations (such as send emails) based on
+these events. 
+
+FLAN implements an event dispatcher to handle batches of events
+based on a timer, allowing for a massive number of events to be handled
+semi-synchronously with rate limiting for massive job throughput. Since
+FLAN is run as a separate process alongside a Flux instance, it can never
+block the job-manager or other critical Flux services. FLAN must be run under 
+the instance owner credentials but needn't be run on the same node as the
+rank 0 Flux broker.
+
+Initial Request
+===============
+
+FLAN SHALL open a streaming RPC request to the JoE. FLAN SHALL request the full
+journal, including completed events.
+
+Initial Response(s)
+===================
+
+An "initial response" is any response prior to the JoE's "sentinel," which
+indicates that the backlog has completed transmission. 
+
+Initial responses are per-jobid and can include multiple events. FLAN SHALL
+store the annotations (jobspec, R) per jobid and process each event.
+
+Event Dispatcher
+================
+
+Instead of handling each event sequentially, events shall be queued and handled
+in batches by the event dispatcher. The event dispatcher SHALL contain a queue
+of events to process. The event dispatcher SHALL process these events after 
+receiving a signal from the reactor's timer watcher. The timer watcher SHALL 
+have a configurable delta (time between wake-ups).
+
+Upon waking up, the event dispatcher shall determine if an event is "of
+interest," and process the event if so. Only the most recent event for a given
+job SHALL be processed. Processing of the event involves clients of the FLAN
+server completing any process they specify, such as sending an email.
+
+On the initial run of the event dispatcher, FLAN SHALL compare the events in
+its queue to a record in the KVS of "handled" events, and ignore "handled"
+events. The initial run of the event dispatcher SHALL block subsequent runs.
+For each subsequent iteration of the event dispatcher, FLAN SHALL write to the
+KVS a record of the events it has processed before the event dispatcher goes to
+sleep. 
+
+Subsequent Responses
+====================
+
+Subsequent responses from the JoE shall be queued in the event dispatcher in
+real-time, and processed when the dispatcher wakes up. A record of jobspec, R,
+and eventlog for each event SHALL be stored, and the record removed by the
+event dispatcher when it receives the ``clean`` event for a job.
+
+User Interface
+**************
+
+Users SHALL create notification-enabled jobs by specifying an attribute in their
+job's jobspec. Jobspec attributes are defined in :doc:`spec_25`.
+
+Basic Use Case
+==============
+
+Users SHALL add the following attribute to their jobspec:
+
+.. code-block:: json
+
+  {
+    "attributes": { 
+      "system": {
+        "notify": "default"
+      }
+    }
+  }
+
+The default behavior SHALL be to send a notification via email to the user
+when the job reaches the START and FINISH events.
+
+A future update to the jobspec API will make this jobspec attribute easily
+accessible via a single argument to a command, ``--notify``.
+
+Advanced Use Cases
+==================
+
+Only the basic use case SHALL be supported in v1.
+
+The ``system.notify`` jobspec attribute SHALL accept a dictionary containing some
+or all of the following values:
+
+.. code-block:: json
+
+  {
+    "attributes": { 
+      "system": {
+        "notify": {
+          "service": "slack",
+          "handle": "elvis",
+          "include": ["R", "eventlog", "return_code"],
+          "states": ["start", "prolog_finish"]
+        }
+      }
+    }
+  }
+
+Edge Cases
+**********
+
+These edge cases MAY be supported in FLAN v1.
+
+Expiration of notifications
+===========================
+
+In certain cases, a restart of the service may be delayed such that events of interest
+on notification-enabled jobs are long past. FLAN MAY support an "expiration" setting
+which would stop any notification from final delivery if a set amount of time had
+passed since the event.
+
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ Table of Contents
 - [41/Job Information Service](spec_41.rst)
 - [42/Subprocess Server Protocol](spec_42.rst)
 - [43/Job List Service](spec_43.rst)
+- [44/Flux Library for Adaptable Notifications](spec_44.rst)
 
 Build Instructions
 ------------------
--- a/spell.en.pws
+++ b/spell.en.pws
@@ -8,6 +8,8 @@ org
 tarball
 tarballs
 adoc
+Mattermost
+JoE
 api
 env
 github


spec_44.rst
```diff
@@ -0,0 +1,235 @@
...
+
+Notification-enabled jobs
+  Jobs that include a jobspec attribute requesting a notification for certain
+  events in the job's life cycle.
+
+The python driver
+  A python process used for tracking notification-enabled jobs through the job
+  life cycle. Started by the flux user on the node containing the rank 0 broker
+  in a cluster, it asynchronously monitors the events for all jobs in the cluster
+  requesting notification. It attaches callbacks to certain events and sends
```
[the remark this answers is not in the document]
wihobbs: "all jobs in the cluster requesting notification" -> notification-enabled jobs 
would maybe be even more concise, and in line with the rest of the doc

spec_44.rst
```diff
@@ -0,0 +1,235 @@
...
+Requirements
+************
+
+ - By default in a system-instance, do not notify a user of any job events.
+   Allow the user to override this default with a jobspec attribute,
+   system.notify.
+ - Support notification after any event of the job, where events are defined in
+   :doc:`spec_21`.
+ - Support email notifications, as well as a driver capable of sending POST
+   requests to any chat service, provided they have an API capable of accepting
```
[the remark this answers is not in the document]
wihobbs: Do you think its necessary to have the definition of chat services in the Terminology section? On reflection, it seems redundant to have a definition of chat services as mattermost and slack and then reiterate that later in the doc too. I think I can drop the "Chat services" definition in terminology...? Thoughts?

spec_44.rst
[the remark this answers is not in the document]
cmoussa1: sounds good to me!

spec_44.rst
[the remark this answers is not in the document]
grondo: Part of the reason the design of the notification service is split into a jobtap plugin and Python service is to make it easier to make the notification types _extensible_ via plugins (I assume that is why the term "adaptable" appears in the name here).

It might be clearer to make the (perhaps eventual) extensibility of the notification a first class design goal in this document, and mention this is done "in order to allow easy support of alternate notification platforms, such as Slack or Mattermost, via Python plugins" and leave it at that.

Then, make the only _requirement_ that the notification service SHALL support notification via email.

index.rst
```diff
@@ -283,6 +283,12 @@ standard I/O management of remote processes.
 
 The Flux Job List Service provides read-only summary information for jobs.
 
+:doc:`spec_44`
+~~~~~~~~~~~~~~
+
+The Flux Library for Adaptable Notifications (FLAN) provides a connection to
+external notification services (such as email) for steps in a batch job.
```
garlick, line 290: Excellent name :+1: 

Did you mean "job state transitions" rather than "steps"  (technically, we don't have steps in flux)

spec_44.rst
```diff
@@ -0,0 +1,235 @@
...
+- :doc:`spec_25`
+
+Background
+**********
+
+Towards the goal of supporting users who run batch jobs with variable end time
+dependent on queues, runtime, and other variable factors, Flux SHALL provide the
+Flux Library for Adaptable Notifications (FLAN). FLAN SHALL be capable of
+sending email notifications to users upon the completion of their batch job.
+FLAN shall be a shared library jobtap plugin loaded in the Flux job manager,
```
garlick, line 41: You could probably relax the RFC 2119 language in the Background section.

Also, the last sentence seems like an implementation detail that doesn't belong in this section?

spec_44.rst
```diff
@@ -0,0 +1,235 @@
...
+  A python process used for tracking notification-enabled jobs through the job
+  life cycle. Started by the flux user on the node containing the rank 0 broker
+  in a cluster, it asynchronously monitors the events for all jobs in the cluster
+  requesting notification. It attaches callbacks to certain events and sends
+  notifications.
+
+The jobtap plugin
+  A shared library based on the API defined in 
+  `flux-jobtap-plugins(7) <https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man7/flux-jobtap-plugins.html#jobtap-plugin-names>`_
+  which streams the jobids of notification-enabled jobs to the python driver.
```
garlick, line 73: Again, maybe it's too early to define the implementation in terms of jobtap plugin and python driver since we're not to the requirements yet?

spec_44.rst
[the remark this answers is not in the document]
wihobbs: As an aside, "adaptable" in the name is also meant to support the goal of making the notifications flexible to user demands, i.e., a user can specify what job metadata and formatting would be relevant and useful to them, and the service will include those fields/comply with that formatting in the notification. (This is the feature described, albeit breifly, in "user interface/advanced use cases" for possible support in v2.)

I like the idea of having to only support emails as the requirement for v1.

spec_44.rst
```diff
@@ -0,0 +1,235 @@
...
+ - Utilize as few resources as possible in the Flux job-manager. Under no
+   circumstances should a notification block any stage or event of a Flux job.
+ - Provide configurable rate-limiting to ensure users can never be overwhelmed
+   by a deluge of notifications, regardless of the number of jobs they submit.
+
+Implementation
+**************
+
+After the jobtap plugin has been loaded in the job-manager, the python driver 
+SHALL send a ``notify.enable`` streaming RPC request at initialization.
```
garlick, line 96: Here is where I would begin with the SHALLs and description of the architecture.

garlick (commented):
Excellent!

I had a few minor suggestions about the flow of the document.  Feel free to ignore if it doesn't make sense to you!

index.rst
wihobbs: Ah, this is good feedback. I think I meant to say _events_ actually. Is that appropriate here?

wihobbs: And glad you like the name! I will have to learn how to make one when the work is finished. 🍰

grondo: BTW, thought of a good tagline for this project: 

 flux-flan: sweet job notifications for flux

spec_44.rst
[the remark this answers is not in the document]
wihobbs: Dropped due to comment below

spec_44.rst
```diff
@@ -0,0 +1,235 @@
...
+The jobtap plugin SHALL keep a hash table of jobids that are ACTIVE and
+notification-enabled.
+
+jobid
+  As defined in :doc:`spec_19`, a single jobid for a notification-enabled job.
+
+.. note::
+  The hash table is intended to ensure that, should the python driver crash,
+  upon restart it can "catch up" with all of the jobs that have been submitted
+  and send users' the notifications they have requested.
```
chu11, line 115: not sure if this is too early in design process, but need an option to say "don't send me old stuff"?  Like hypothetically request for notifications is down for the day (say plugin to e-mail service is down for some reason) and when it comes back up, bam, a bajillion notifications and bajillion e-mails are sent out.

spec_44.rst
```diff
@@ -0,0 +1,235 @@
...
+--------------
+
+Users SHALL add the following attribute to their jobspec:
+
+.. literalinclude:: data/spec_44/example2.yaml
+   :language: yaml
+
+The default behavior SHALL be to send a notification to the users' primary email
+address, as provided by an LDAP query, when the job reaches the START and FINISH
+events.
```
chu11, line 173: I initially read this and thought this implementation was way too specific.  But then I thought maybe you're just giving a use case example.  Perhaps the words "The default behavior" threw me off.  Maybe "For example, the default behavior ..."

chu11 (commented):
Just a side question, is the python driver typically going to be associated with system instance?  So we'd run this python driver as a side daemon.

Or is the python driver something the user would start with their subinstance?  How would they launch the python driver?

spec_44.rst
```diff
@@ -0,0 +1,235 @@
...
+  The hash table is intended to ensure that, should the python driver crash,
+  upon restart it can "catch up" with all of the jobs that have been submitted
+  and send users' the notifications they have requested.
+
+Additional Responses
+--------------------
+
+The jobtap plugin SHALL continue to send responses to the initial 
+``notify.enable`` RPC request whenever notification-enabled jobs enter the
+DEPEND state. The jobtap plugin shall add these job's jobids to its hash
```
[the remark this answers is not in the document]
wihobbs: Sentence is clearer if you drop "jobs" altogether and just say "jobids" :-)

spec_44.rst
wihobbs: I added an "edge case" called "Expiration of notifications" and said that we MAY support this. Will push up soon.

spec_44.rst
wihobbs: I liked having definitions for these, since "jobtap plugin" etc. has a broader meaning outside this RFC, so I moved these to below "Implementation" and kept two of the existing definitions. On a second pass let me know if that formatting change is helpful/harmful to the flow.

spec_44.rst
```diff
@@ -0,0 +1,235 @@
...
+Flux does not currently support restarting with running jobs. However, on a system
+restart, all events for all ACTIVE jobs are replayed. This means that when each
+notification-enabled active job reaches the DEPEND event, the jobtap plugin SHALL
+send a streaming RPC response and insert the job's jobid into its hash table. The
+python driver, upon receiving a new jobid MUST ensure that the jobid does not have
+a previous entry in the KVS. Since the KVS is reloaded on a restart, any outstanding
+notifications shall have corresponding keys there. If a jobid received by the python
+driver already has a KVS subdirectory, the python driver shall ignore the job's 
+event notification requests in the jobspec and only send notifications for that
+correspond with the keys in the KVS. This prevents a double-notification of the user
```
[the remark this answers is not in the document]
wihobbs: Oh I see. It's that "for that correspond" -> "that correspond"

spec_44.rst
wihobbs: The default behavior is supposed to be detailing what happens when a user says `flux batch --setattr=system.notify=default`. The eventual goal is to support doing crazy things like `--setattr=system.notify.service=slack --setattr=system.notify.events=validate,epilog-start --setattr=system.notify.include="{id.f58} {R} {eventlog}`, etc. so I wanted to give the behavior in the general case (which is probably all we'll actually support in v1.)

I do think that maybe specifying LDAP is a tad specific. Though I've generally tried to be as specific as possible in this RFC since its more of a design document for a service than a traditional RFC...

spec_44.rst
```diff
@@ -0,0 +1,250 @@
...
+
+Initial Request
+---------------
+
+After the jobtap plugin has been loaded in the job-manager, the python driver 
+SHALL send a ``notify.enable`` streaming RPC request at initialization.
+
+The ``notify.enable`` request has no payload.
+
+At initialization the python driver SHALL create a kvs subdirectory, ``notify``.
```
grondo, line 98: What if the directory already exists? :-)

spec_44.rst
```diff
@@ -0,0 +1,250 @@
...
+
+Initial Response
+----------------
+
+Multiple responses may be sent to the initial ``notify.enable`` RPC request.
```
grondo, line 103: The RPC is already defined as multi-response above. This line could probably be removed.

spec_44.rst
```diff
@@ -0,0 +1,250 @@
...
+---------------
+
+After the jobtap plugin has been loaded in the job-manager, the python driver 
+SHALL send a ``notify.enable`` streaming RPC request at initialization.
+
+The ``notify.enable`` request has no payload.
+
+At initialization the python driver SHALL create a kvs subdirectory, ``notify``.
+
+Initial Response
```
grondo, line 100: I think it would help to title this "Initial Responses" since the first request could result in multiple responses. I got a bit confused looking for a special single "Initial Reponse".

spec_44.rst
```diff
@@ -0,0 +1,250 @@
...
+----------------
+
+Multiple responses may be sent to the initial ``notify.enable`` RPC request.
+The jobtap plugin SHALL keep a hash table of jobids that are ACTIVE and
+notification-enabled. On initialization, all of the jobids in the hash table
+SHALL be sent as individual responses to the Python driver.
```
grondo, line 106: Probably unnecessary to define how the jobtap plugin keeps track of active, notification-enabled jobids. Just define the required response and leave out the specific implementation details.

spec_44.rst
```diff
@@ -0,0 +1,250 @@
...
+
+jobid
+  As defined in :doc:`spec_19`, a single jobid for a notification-enabled job.
+
+.. note::
+  The hash table is intended to ensure that, should the python driver crash,
+  upon restart it can "catch up" with all of the jobs that have been submitted
+  and send users the notifications they have requested.
```
grondo, line 114: s/the hash table/the initial responses/?

spec_44.rst
```diff
@@ -0,0 +1,250 @@
...
+For each response received by the python driver, the driver SHALL create a 
+KVS subdirectory, ``notify.<jobid>``. In this directory the driver SHALL
+insert keys representing the job events for which users have requested a 
+notification. These keys values SHALL be empty. The key SHALL be deleted
+after the corresponding notification is sent. 
+
+The python driver MUST then asynchronously monitor the job as it reaches
+events of interest.
+
+When the job reaches an event of interest, FLAN SHALL generate an email
```
grondo, line 133: Why the switch from "Python driver" to FLAN here?

spec_44.rst
```diff
@@ -0,0 +1,250 @@
...
+-------------------------
+
+Due to the recursive launch feature of Flux, users may wish to have notifications 
+for states of batch jobs that are not at the system-instance level. This MAY NOT 
+be supported in FLAN v1.
+
+Invalid jobspec attributes
+--------------------------
+
+FLAN MAY eventually provide a frobnicator plugin for validating the advanced use
```
grondo, line 247: validator plugins validate jobspec, not frobnicators (these modify jobspec)

spec_44.rst
```diff
@@ -0,0 +1,250 @@
...
+Advanced Use Cases
+------------------
+
+Only the basic use case SHALL be supported in v1.
+
+The ``system.notify`` jobspec attribute SHALL accept a dictionary containing some
+or all of the following values:
+
+.. literalinclude:: data/spec_44/example1.yaml
+   :language: yaml
```
grondo, line 191: I'm thinking it would be better to include JSON here instead of having YAML kept separately. (This will be much easier to track and evolve as we modify the jobspec contents)

spec_44.rst
```diff
@@ -0,0 +1,250 @@
...
+
+For System Administrators
+-------------------------
+
+The webhooks and other secrets required to connect to chat services SHALL be included
+in a ``config.toml`` file. The path to this file MUST be provided to the FLAN 
+python driver on initialization. Note that best practice for managing webhooks is 
+to keep them secret.
```
grondo, line 199: It is good you're thinking about this, but it is an implementation detail and probably should just be excised for now.

Instead, I think this document should just state that the Python driver SHALL support alternate notification mechanisms via plugins, so that that third parties can extend the set of supported backend notifications.

Once some work is done on that plugin system perhaps a config file will become necessary and at that time could be further fleshed out here?

grondo (commented):
Ok, I just took a quick first pass :+1:

index.rst
```diff
@@ -283,6 +283,12 @@ standard I/O management of remote processes.
 
 The Flux Job List Service provides read-only summary information for jobs.
 
+:doc:`spec_44`
+~~~~~~~~~~~~~~
+
+The Flux Library for Adaptable Notifications (FLAN) provides a connection to
```
vsoch, line 289: This is an excellent name. Although I do not like the texture of flan, +1 from me!

spec_44.rst
```diff
@@ -0,0 +1,250 @@
...
+*****************
+
+- :doc:`spec_19`
+- :doc:`spec_21`
+- :doc:`spec_25`
+
+Background
+**********
+
+Towards the goal of supporting users who run batch jobs with variable end time
```
vsoch, line 36: Nit: [Toward](https://www.grammarly.com/blog/toward-towards/#:~:text=The%20only%20difference%20between%20toward,is%20the%20more%20common%20spelling.)

spec_44.rst
```diff
@@ -0,0 +1,250 @@
...
+  events in the job's life cycle. For a more detailed definition of job events,
+  refer to :doc:`spec_21`.
+
+
+Requirements
+************
+
+ - By default in a system-instance, do not notify a user of any job events.
+   Allow the user to override this default with a jobspec attribute,
+   system.notify.
```
vsoch, line 62: Why does it go under `system` ?

spec_44.rst
```diff
@@ -0,0 +1,250 @@
...
+ - Support notification after any event of the job, where events are defined in
+   :doc:`spec_21`.
+ - Support email as the primary end user notification delivery.
+ - Build a driver capable of sending POST requests to chat services for
```
vsoch, line 66: Where do you see tokens or credentials for these posts going? I assume it's on the level of the user? So the user needs to setup their own slack app to make it work? But what about an organization or group - how would that be handled, permissions wise for the credentials?

spec_44.rst
```diff
@@ -0,0 +1,250 @@
...
+
+Implementation
+**************
+
+FLAN SHALL be implemented in two parts:
+
+The jobtap plugin
+  A shared library based on the API defined in 
+  `flux-jobtap-plugins(7) <https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man7/flux-jobtap-plugins.html>`_
+  which streams the jobids of notification-enabled jobs to the python driver.
```
vsoch, line 82:
```suggestion
  that streams the jobids of notification-enabled jobs to the python driver.
```
Nit: which only comes after a comma.

spec_44.rst
```diff
@@ -0,0 +1,250 @@
...
+
+The python driver
+  A python process used for tracking notification-enabled jobs through the job
+  life cycle. Started by the flux user on the node containing the rank 0 broker
```
vsoch, line 86: Is this saying the username `flux` or "the user of Flux" as in the flux user? I think if you capitalize for the latter it would be more clear, otherwise (if it is username == `flux`) maybe add the code formatting as I've done here.

spec_44.rst
```diff
@@ -0,0 +1,250 @@
...
+  in a cluster, it asynchronously monitors the events of all notification-enabled
+  jobs. It attaches callbacks to certain events and sends notifications.
```
vsoch, line 88: How is uniqueness of an event determined? You said there would be rate limiting, but I think generally we want to only notify of each event once, and have a means to determine "this is an event that the user already heard about" (and do not notify again.

spec_44.rst
```diff
@@ -0,0 +1,250 @@
...
+   system.notify.
+ - Support notification after any event of the job, where events are defined in
+   :doc:`spec_21`.
+ - Support email as the primary end user notification delivery.
+ - Build a driver capable of sending POST requests to chat services for
+   notification delivery, provided they have an API capable of accepting such 
+   requests. Examples include but are not limited to Mattermost and Slack.
+ - Utilize as few resources as possible in the Flux job-manager. Under no
+   circumstances will a notification block any stage or event of a Flux job.
+ - Provide configurable rate-limiting to ensure users can never be overwhelmed
```
vsoch, line 71: Can you explain what you mean by rate limiting here? Arguably we can have it, but we shouldn't ever find ourselves in a scenario where "the same" event is triggering notifications so many times as to need it. When would this happen/

spec_44.rst
```diff
@@ -0,0 +1,250 @@
...
+The python driver
+  A python process used for tracking notification-enabled jobs through the job
+  life cycle. Started by the flux user on the node containing the rank 0 broker
+  in a cluster, it asynchronously monitors the events of all notification-enabled
+  jobs. It attaches callbacks to certain events and sends notifications.
+
+Initial Request
+---------------
+
+After the jobtap plugin has been loaded in the job-manager, the python driver
```
vsoch, line 93: Make sure to be consistent in capitalization. Python is capitalized later but not here.

spec_44.rst
vsoch: I would say the plugin should have a mechanism to crash and restore state, but without explicitly saying it has to be a hash table. Why does it?

spec_44.rst
```diff
@@ -0,0 +1,250 @@
...
+  The hash table is intended to ensure that, should the python driver crash,
+  upon restart it can "catch up" with all of the jobs that have been submitted
+  and send users the notifications they have requested.
+
+Additional Responses
+--------------------
+
+The jobtap plugin SHALL continue to send responses to the initial 
+``notify.enable`` RPC request whenever notification-enabled jobs enter the
+DEPEND state. The jobtap plugin SHALL add these jobids to its hash table
```
vsoch, line 121: Why DEPEND?

spec_44.rst
```diff
@@ -0,0 +1,250 @@
...
+The jobtap plugin SHALL continue to send responses to the initial 
+``notify.enable`` RPC request whenever notification-enabled jobs enter the
+DEPEND state. The jobtap plugin SHALL add these jobids to its hash table 
+of ACTIVE, notification-enabled jobs.
+
+For each response received by the python driver, the driver SHALL create a 
+KVS subdirectory, ``notify.<jobid>``. In this directory the driver SHALL
+insert keys representing the job events for which users have requested a 
+notification. These keys values SHALL be empty. The key SHALL be deleted
+after the corresponding notification is sent.
```
vsoch, line 128: Is there any case when you might see the same event for an equivalent jobid) and it will be sent again? Or is the idea that when the key is deleted it's not going to trigger again? I'm trying to imagine what happens if the plugin crashes, the responses come in again, and the keys are re-created and the user spammed with notifications. Is that possible?

spec_44.rst
```diff
@@ -0,0 +1,250 @@
...
+
+The python driver MUST then asynchronously monitor the job as it reaches
+events of interest.
+
+When the job reaches an event of interest, FLAN SHALL generate an email
+and send it to the user. (FLAN MAY eventually support other means of 
+notification delivery, such as chat services.) FLAN SHALL subsequently
+delete the corresponding key in the KVS, ``notify.<jobid>.<state>``.
+
+The ``notify.<jobid>`` KVS subdirectory SHALL be deleted when the job reaches
```
vsoch, line 138: Is it possible that this removal can be missed, and you have jobs that are done INACTIVE but the user keeps getting notifications?

spec_44.rst
```diff
@@ -0,0 +1,250 @@
...
+
+The ``notify.<jobid>`` KVS subdirectory SHALL be deleted when the job reaches 
+an INACTIVE state. If the ``notify.<jobid>`` directory is non-empty upon 
+reaching the INACTIVE state, this indicates some notifications have been missed.
+The python driver SHALL send a final notification to the user documenting 
+that their notification-enabled job has reached an INACTIVE state.
+
+.. note::
+  This design is intended to ensure that no double-notifications are sent upon
+  the restart of the Python script, the jobtap plugin, or the job-manager.
```
vsoch, line 146: Unless the directory deletion is missed?

spec_44.rst
```diff
@@ -0,0 +1,250 @@
...
+jobtap plugin is not loaded in the job-manager. The python driver SHALL exit 
+immediately, and print an appropriate error message.
+
+Disconnect Request
+------------------
+
+If a disconnect request is received by the jobtap plugin, this indicates the
+python driver has exited. The jobtap plugin SHALL continue to add notification-
+enabled jobs to its hash table as they enter the DEPEND state. When the python
+driver reconnects, the jobtap plugin SHALL respond to its initial ``notify.enable``
```
vsoch, line 161: What if it doesn't reconnect?

spec_44.rst
```diff
@@ -0,0 +1,250 @@
...
+Basic Use Case
+--------------
+
+Users SHALL add the following attribute to their jobspec:
+
+.. literalinclude:: data/spec_44/example2.yaml
+   :language: yaml
+
+The default behavior SHALL be to send a notification to the users' primary email
+address, as provided by an LDAP query, when the job reaches the START and FINISH
```
vsoch, line 179: What if there isn't LDAP? Isn't this an implementation detail for a plugin that requires (sends to) email, and out of scope for the high level spec?

spec_44.rst
```diff
@@ -0,0 +1,250 @@
...
+or all of the following values:
+
+.. literalinclude:: data/spec_44/example1.yaml
+   :language: yaml
+
+For System Administrators
+-------------------------
+
+The webhooks and other secrets required to connect to chat services SHALL be included
+in a ``config.toml`` file. The path to this file MUST be provided to the FLAN
```
vsoch, line 197: I would either place this in a directory that has a meaningful path (for the developer to see) or give it name that is more specific to describe what it is. And I am wondering who has ownership of this file - a single user or if a center needs to take responsibility, for example, to provide a single credential for all users to use (is that a good idea)?

vsoch: Likely both use cases could be wanted (the "I am a user with my own custom plugin and private credential" and "I am a center providing this to my users."

spec_44.rst
```diff
@@ -0,0 +1,250 @@
...
+
+These edge cases MAY be supported in FLAN v1.
+
+Restarting the job-manager
+--------------------------
+
+In the event the job-manager crashes or is shut down the python driver SHALL exit
+immediately and log an error. 
+
+Flux does not currently support restarting with running jobs. However, on a system
```
vsoch, line 217: Trying to restore state sounds messy - is there any reason to not recreate notifications on restart so there aren't any mismatches between state brought up and state known to the plugin?

spec_44.rst
wihobbs: Ah, it's been too long, I ought to go reread the [AP stylebook...](https://twitter.com/APStylebook/status/484339629920100352?lang=en)

spec_44.rst
wihobbs: There's a longer explanation in RFC [14/Canonical Job Specification](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_14.html#attributes) about `system` vs. `user` jobspec attributes. The TLDR as I understand it is that attributes _Flux_ needs go under `system` (such as changing your bank in flux-accounting, etc.) while things that the user wants to propagate to their application go under `user`. 

A benefit of having jobspec attributes under system is that's the default if nothing is specified in `setattr`, i.e. `--setattr=notify=default` expands to `--setattr=system.notify=default`. Another is that under `system` we can make `notify` a reserved word which isn't possible under `user`.

I had a longer conversation with Mark about this that got eaten by the Slack monster :( where he had a more detailed explanation for why the notification service attributes would go under `system` .

grondo: That's a pretty good summary @wihobbs.

spec_44.rst
wihobbs: Actually, I had envisioned credentials/tokens being stored in a config file on the management node. The only thing a user would need to pass to the jobspec would be their handle, i.e. hobbs17 on the LLNL slack. The slack admins would set up a bot to handle delivering the message to the user, and the sysadmins would store the webhooks for this bot in a toml file on the management node.

An edge case I haven't considered yet is "what if we need to support multiple 'teams' on Slack," i.e. LLNL and llnl-performance.

I should probably go read the Slack/Mattermost API docs and make sure this approach will actually work. I've built very little with MM and nothing with Slack...

spec_44.rst
wihobbs: Good point. I was thinking the `flux` user, but really I meant "the instance owner of the system instance" (which we usually configure to be `flux`.)

spec_44.rst
grondo: In any event I would probably place configuration location of third party tokens, etc. outside the scope of this document, which should focus on how the service monitors jobs for events of interest to turn them into generic notifications. (i.e. the protocol between Python service and a jobtap plugin, how restarts are handled, etc)

spec_44.rst
grondo: we may want to think through the ergonomics of requesting notifications before codifying into an RFC. Requiring a user to specify multiple `--setattr` options on the command line might not go over that well..

grondo: On second thought, it would be fine to have something in here, it can always be updated in the future. However, just keep in the back of your mind that users would appreciate a single `--setattr` so perhaps comma delimited options or other ways to provide options in a single go should be considered.

spec_44.rst
wihobbs: As a first revision I'm just going to always capitalize Python. However, now that I think about it, I wonder if it's necessary to be constantly saying "the Python driver." If the term "the Python driver" -> "the driver" in the definitions under Implementation, and the definition itself mentioned Python, maybe we could just call it "the driver" throughout the RFC...

spec_44.rst
wihobbs: It doesn't. In my prototypes, it's been a `std::map` in C++, could be a vector, could be ...

Accepting this substitution and removing future references to "hash table." I'll refer to it more generally as "The jobtap plugin SHALL keep a record..."

spec_44.rst
wihobbs: "reconnect" is probably the wrong word. The Python driver if disconnected from the jobtap plugin will exit immediately and log an error. Eventually, systemd or an admin will start a new Python driver which sends a new initial `notify.enable` RPC request. No notifications can be sent while there's not an active Python driver process.

spec_44.rst
wihobbs: Excised all references to "hash table" (explanation below)

wihobbs:
Note that ^ push is mostly grammar/clarity fixes and I'm still working through the [excellent, but involved :) ] design feedback.

spec_44.rst
wihobbs: The Python driver can traverse the existing directory for jobid keys.

If the jobtap plugin could _insert_ and _remove_ jobid keys itself, this eliminates the need for the record of jobids in the jobtap plugin ("hash table"). The jobtap plugin could have an INACTIVE callback to remove, and add the key in DEPEND. Currently the KVS is only being used to ensure uniqueness of the notification being sent.

Thoughts on this design?

spec_44.rst
wihobbs: Oh, maybe for the "center providing for users" case this needs an `/etc/flux/notify.d/` or similar. Probably an implementation detail outside this RFC, but I had not thought about standardizing the config file path. Good idea!

spec_44.rst
wihobbs: Notifications would be recreated internally in the Python driver, since losing its connection to Flux is a fatal error (and that would happen in a system-instance restart). 

Restoring state is messy...but I want to try for it anyway :). Here's an idea for how:

1. Python driver crashes because job-manager is shut down.
2. Job-manager is restarted (and so is the KVS). KVS comes back; jobtap plugin is loaded on start. All job eventlogs are replayed.
3. All of the jobtap plugin depend callbacks are hit again. Jobtap plugin records these jobids.
4. A "new" Python driver starts and gets a bunch of jobids sent to it. It watches the events [that may be somewhat stale.] (This was @chu11's point in a message above.) It does not add jobid folders to the KVS because they are already there.
5. Before it sends a notification, it checks the KVS `notify.<jobid>` folder to see if __ (event) key is there. If the key is _not_ there, that means a notification has already been sent. (On reflection, maybe it should be flipped -- i.e. if the key is there, that means a notification has _already_ been sent.)
6. Stop! Do not pass go, do not collect $200! Because we know the job is being watched, but a notification has already been sent, keep chugging and watching for others that may _not_ have been sent.

grondo: It is a bit unclear here what is meant by the "depend callbacks are hit again".

Note that there are two jobtap callbacks related to dependencies and the DEPEND state:
 - `job.state.depend`: One of the generic `job.state.*` callbacks, called when the job reaches the DEPEND state.
 - `job.dependency.SCHEME`: Called for each dependency `SCHEME` in jobspec of a submitted job

I don't think the `job.state.*` callbacks are called on reload. You can test this by writing a jobtap plugin that subscribes to all these states and prints what it sees on a restart.

The `job.dependency.SCHEME` callbacks _are_ specifically issued on a restart for jobs with dependencies defined in jobspec.

I know we talked about this before, but I don't think I properly understood you were talking about one of these callback types and not the other.

Probably the best way for plugins to be notified of new jobs is to use the `job.new` callback, which is called for new jobs after they've been validated, on plugin reload, and on job manager restart.

wihobbs: For lack of a better place to document this (tagging @garlick as well):

It appears the `job.state.priority` callback is re-triggered when `flux start` is run in recovery mode with pending jobs, but no other `job.state.*` callbacks.

```c
#include <flux/core.h>
#include <flux/jobtap.h>
#include <stdio.h>
#include <jansson.h>
#include <stdlib.h>

static int print_cb (flux_plugin_t *p,
                     const char *topic,
                     flux_plugin_arg_t *args,
                     void *arg)
{
    flux_jobid_t id;
    long unsigned int state;
    flux_t *h = flux_jobtap_get_flux(p);
    json_t *event;

    if (flux_plugin_arg_unpack(args, 
                               FLUX_PLUGIN_ARG_IN,
                               "{s:I s:i s:o}",
                               "id", &id,
                               "state", &state,
                               "entry", &event) < 0)
        return -1;
    
    if (id) {
        char* json_string = json_dumps(event, JSON_ENCODE_ANY);
	    flux_log(h, 0, "jobid %ld has entered state %s\n", id, json_string);
    }

    return 0;
}

static const struct flux_plugin_handler tab[] = {
    {"job.state.*", print_cb, NULL},
    {0},
};

int flux_plugin_init (flux_plugin_t *p) {
    if (flux_plugin_register(p, "jobtap-print", tab) < 0) {
        return -1;
    }
    return 0;
}
```
And on shutdown/restart:
```
  auk108 ~/scripts/jobtap-test $ flux start -o,--config-path=$(pwd)/config.toml
(s=1,d=0)  auk108 ~/scripts/jobtap-test $ flux submit --urgency=0 hostname
May 14 15:19:03.251059 job-manager.emerg[0]: jobid 330880253952 has entered state 140651589009410 {"timestamp": 1715725143.2404392, "name": "validate"}
May 14 15:19:03.251083 job-manager.emerg[0]: jobid 330880253952 has entered state 140651589009412 {"timestamp": 1715725143.2510674, "name": "depend"}
May 14 15:19:03.251104 job-manager.emerg[0]: jobid 330880253952 has entered state 140651589009416 {"timestamp": 1715725143.251091, "name": "priority", "context": {"priority": 0}}
ƒ9h7knoh
(s=1,d=0)  auk108 ~/scripts/jobtap-test $ flux shutdown --dump=$(pwd)/dump.tar
  auk108 ~/scripts/jobtap-test $ flux start -o,--config-path=$(pwd)/config.toml --recovery=$(pwd)/dump.tar
May 14 15:22:46.751219 job-manager.emerg[0]: jobid 330880253952 has entered state 139796890517508 {"timestamp": 1715725366.7511699, "name": "flux-restart"}
May 14 15:22:46.751270 job-manager.emerg[0]: jobid 330880253952 has entered state 139796890517512 {"timestamp": 1715725366.7512314, "name": "priority", "context": {"priority": 0}}
+-----------------------------------------------------
| Entering Flux recovery mode.
| All resources will be offline during recovery.
| Any rc1 failures noted above may result in
|  reduced functionality until manually corrected.
|
| broker.rc1_path    /home/hobbs17/flux-core/../bin/etc/flux/rc1
| broker.rc3_path    /home/hobbs17/flux-core/../bin/etc/flux/rc3
| config.path        /home/hobbs17/scripts/jobtap-test/config.toml
| statedir           changes will not be preserved
|
| Exit this shell when finished.
+-----------------------------------------------------
```

I accept the suggestion to tie the streaming RPC response to `job.new`, which is replayed on restart (tweaked the test above and verified).

We will have to make the ability to restart the service dependent on a _requirement_ that this plugin be included in the `job-manager`'s `plugins` table in the config.

grondo: `job.new` is also called for every active job when a plugin is loaded (on the assumption that this is the method that "introduces" jobs to jobtap plugins). This should allow the plugin to be reloaded (e.g. to get a new version) and get back to the same place it was before it was unloaded.

grondo: Nice job on the test plugin, btw. You could tweak it to register callbacks for `*` if you want to see everything.

spec_44.rst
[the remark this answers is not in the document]
vsoch: > such as Slack or Mattermost, via Python plugins" and leave it at that.

+1 to provide these are "examples of such plugins" and not make any promises. Indeed a notification service on the level of a user is going to have a different level of permission than one that goes to an entire channel, or a single email. And the design you choose for each is going to be subtly different. I've made bots for Twitter, slack, Mastodon, Discord, GitHub, Reddit, email, and niche ones like Uservoice or other help desks, and they are all special snowflakes. They change their APIs too (I'm looking at you, Twitter)!

index.rst
vsoch: Love me those sweet, sweet notifications! :cake:

spec_44.rst
vsoch: For longer term, maybe a design that is readable and simple, something like `--notify.service=slack` I can read from left to right.

spec_44.rst
vsoch: :heavy_plus_sign:

spec_44.rst
vsoch: I think the answer is: https://devguide.python.org/documentation/style-guide/

spec_44.rst
vsoch: My question then is - what if the admin doesn't start a new Python driver? The hash table keeps filling up and explodes?

vsoch:
@wihobbs you've probably already exceeded most of us in knowledge of jobtap plugins, but in case you are interested we are having a little (hopefully fun) hackathon this week on Thursday - grandmaster @trws is going to show us the ropes for making plugins, and :exclamation: I think there might be rust involved! If you are interested, even for the fun or sharing your expertise, I can send along the invite.

spec_44.rst
wihobbs: > DEPEND callbacks get re-triggered on a restart, when the event log is replayed. It's also early enough in the job life cycle that it can hand event monitoring off to the Python driver, and take up few resources in the job-manager.

I am wrong here! (longer explanation below)
Changing this to the `job.new` [callback topic.](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man7/flux-jobtap-plugins.html#job-callback-topics)

spec_44.rst
wihobbs: This is a really good point. I hadn't thought of this. Maybe the jobtap plugin should remove jobids when they reach an inactive state... the notifications would be missed, though. Hmm. 🤔 

There might be a way to log a warning reminding the user/admins to start the Python driver, but I'll have to look into that more.

grondo: Let's add this to the list of things to investigate. It shouldn't be an issue to keep a hash/list of all jobids with notifications, but we should not assume there is not a better way at this point in the design.

spec_44.rst
wihobbs: Included a JSON code block (mostly) similar to what was in the YAML files. Resolving this, but the ergonomics of advanced use cases remains TBD.

wihobbs:
Tons of great feedback on this. Thanks to everybody for participating and sharing your expertise.

I tried to close or consolidate duplicate issues. I apologize in advance if something was inadvertently closed -- please reopen if so.

I'm making a high-level summary of the outstanding design comments.

- Recovery of state on a crash. Note that there are three things that can crash, and we should possibly have a different strategy for handling each:
  - Crash of the job-manager (+Python driver+jobtap plugin) -- essentially restoring the state of everything.
  - Crash of the Python driver, but jobtap plugin and job-manager persist.
  - Crash of the jobtap plugin, but not the job-manager (+?Python driver)
- Ensuring uniqueness of events before sending notifications (related: no double-or-worse notifications, and no missed notifications)
- Rate limiting: why do we need it and how will we implement it?
- Extensibility of the service via plugins (for external services such as Slack). Implementation of individual services is not a requirement.
- Ergonomics of the "advanced use case," where we allow a user to specify services other than email, non-standard things to include, and states other than `START` and `FINISH` in which to trigger notifications.

spec_44.rst
trws: Given this, my question above is becoming a bit more urgent.  Is there a reason to keep these things separate? If so, is there a reason the jobtap plugin shouldn't manage the lifetime of the driver as a subprocess or systemd unit?

spec_44.rst
```diff
@@ -0,0 +1,258 @@
...
+**************
+
+FLAN SHALL be implemented in two parts:
+
+The jobtap plugin
+  A shared library based on the API defined in 
+  `flux-jobtap-plugins(7) <https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man7/flux-jobtap-plugins.html>`_
+  that streams the jobids of notification-enabled jobs to the Python driver.
+
+The Python driver
```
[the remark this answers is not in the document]
grondo: It's largely because the jobtap plugin resides in the job manager and we want it to do as little as possible since it would be in the critical path.

The jobtap plugin (at the time) seemed like the most efficient way to avoid having to fetch and examine the jobspec of every job since it can just peek at the jobspec that the job manager already has in memory and send the jobid along to the Python process only when the user has requested notification.

_However_, I'm glad you asked because I wonder if there's a better way to do this using the job-manager journal or some other more widely applicable method. It sure would be nice if it was easy for people to develop services like this and not have to install yet another jobtap plugin to identify jobs of interest.

Something to consider...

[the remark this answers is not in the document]
trws: Ah, that's a good point. Blocking the job manager would be unfortunate.

The journal is a good thought, or maybe even a pub-sub event from the plugin so anyone could subscribe to it? Not sure, there are definitely tradeoffs here, but it sounds like the current approach has some potential to cause lost or repeated notification as well, especially racing with the event delivery and kvs removal vs jobtap replay if flux is restarted.

spec_44.rst
wihobbs: @grondo and I discussed this yesterday.

A user could submit 1000+ jobs at the system instance and just reuse jobspecs for all of them, not realizing they had requested notification. We don't want to spam them with an individual email for each of these, or overwhelm the email server.

Rate limiting could be implemented with a reactor-like design (maybe an "event dispatcher" is the term). As notifications are generated, they could be pushed into a list/queue. The dispatcher could wake up at a set interval, check for new notifications, and batch all of the user's jobs into just one notification. It's not critical that the notifications be instantaneous.

wihobbs: Take 1000+ to mean "some large number for which we don't want to send individual notifications."

wihobbs:
I need to reread some of the previous comments before asking for another set of reviews. This is a pretty substantial update to the proposed design. 

Some concerns will stay and some will go, but a major advantage of subscribing to the journal of events is that it absolutely cannot block the job manager, because it's a separate process completely.

I also noticed that we've ended up in a race over who gets RFC no. 44 😉😆

spec_44.rst
wihobbs: Glad I posted this because I lost it in my home directory somewhere over the last 3 months!

wihobbs: With a dispatcher handling actual notification transmission there's an opportunity a few notifications would be missed on an abrupt shutdown, but maybe that's not so bad for a first pass.

wihobbs:
Ideas that were discussed, but not included in the RFC:

- Posting an event to the eventlog of the job when a notification is sent. It seems easier to keep the record of what's been sent all in one place that's quickly accessible by the driver, rather than dissociated in each job's eventlog.
- A serialized hash table, hashing each notification before sending, so that a collision (sent notification) can be easily detected. I determined that this was an implementation detail and too deep for an RFC.
- Using systemd freeze and thaw for storing the state of the driver. I don't think this will work in the way we thought, and we have access to the KVS, so I figured we'd just use that for storing state.

grondo:
> Posting an event to the eventlog of the job when a notification is sent. It seems easier to keep the record of what's been sent all in one place that's quickly accessible by the driver, rather than dissociated in each job's eventlog.

Haven't read through the new RFC yet, but just as a counterpoint here: There are some nice benefits to using an event in the eventlog. For example when jobs are purged those events won't appear on the next restart (the notification service will have to purge its own records otherwise, adding unnecessary complexity). Also, you're already getting all job events through the journal, so this would limit the number of sources the notification service has to deal with, again simplifying the implementation.

That being said, I think we discussed that perhaps an event posted by the notification service would appear in the job eventlog _after_ it is needed (e.g. the `start` event is posted, the service sends a notification, then posts a notification event -- when restarting the service would see the `start` event before the `notification` event so it may not be helpful for preventing duplicate notifications) IMO, this is a more compelling argument to drop the suggestion that events be posted by the notification service in job eventlogs than because it _seems_ easier.

PR #415, opened 2024-05-28
rfc42: note response requirements for a successful exec stream, plus cleanup

garlick:
Problem: while tracking down flux-framework/flux-core#5956, it became evident that RFC 42 does not specify which responses must be expected in a successful exec stream.

Add a paragraph that clarifies this.

Also add a couple of other minor cleanups.

### DIFF
--- a/spec_42.rst
+++ b/spec_42.rst
@@ -34,8 +34,7 @@ Related Standards
 Background
 **********
 
-The subprocess server protocol is implemented in two distinct Flux
-components:
+The subprocess server protocol is implemented in three Flux components:
 
 .. list-table::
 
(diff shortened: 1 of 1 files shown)


spec_42.rst
```diff
@@ -59,6 +62,8 @@ The primary use cases are:
 
 #. The instance owner runs arbitrary processes with :program:`flux exec`.
 
+#. Tool daemon launch using :program:`flux exec`.
```
chu11, line 65: i assume this refers to parallel debug stuff?  Maybe can just mention that?

chu11 (approved):
LGTM, just that one nit comment

spec_42.rst
garlick: Well it's potentially a bit broader than that but I'll go ahead and mention it as an example.

garlick: Fixed that and forced a push.

PR #416, opened 2024-05-28
rfc42: exec.write should use matchtag not pid

garlick:
Problem: requiring the PID as a parameter to `write` complicates client code which must support writing/closing a stream before the `started` response has been received.

Use the `exec` request matchtag instead.

Note: this came up while discussing flux-framework/flux-core#6002


grondo (approved):
LGTM!

PR #417, opened 2024-06-18
rfc43: add job-list 'ranks' constraint

grondo:
Problem: A hostlist constraint operator is defined in the set of supported job-list constraint operators, but there is no corresponding ranks constraint.  The ranks constraint is useful not only in test instances where multiple ranks may share the same hostname, but also in cases where a specific rank is to be queried without dependence on hostname (e.g. rank 0).

Define a supported `ranks` constraint operator in RFC 43.


chu11 (approved):
oh yeah, that does seem like a good idea.

grondo:
Thanks! I've set MWP.

PR #420, opened 2024-08-06
rfc14: add bank and project as optional jobspec attributes

grondo:
Problem: The canonical jobspec attributes for `bank` and `project` are not specified in RFC 14. Even though these attributes only apply for multi-user instances with an accounting plugin configured, they are common enough concepts that it would be useful to standardize their name, location in jobspec, and purpose.

Define `bank` and `project` attributes in RFC 14. Add a note that these attributes may have no effect unless an accounting package is installed and configured.

### DIFF
--- a/spec_14.rst
+++ b/spec_14.rst
@@ -334,6 +334,18 @@ Some common system attributes are:
    The value of the ``queue`` attribute is a string containing the name of
    the job queue this job should be submitted to.
 
+**bank**
+   The value of the ``bank`` attribute is a string containing the name of
+   an accounting bank requested for this job. Support for the ``bank``
+   attribute MAY depend on installation and configuration of optional
+   framework components, such as an accounting or other plugin.
+
+**project**
+  The value of the ``project`` attribute is a string containing the
+  name of an accounting project requested for this job. Support for the
+  ``project`` attribute MAY depend on installation and configuration of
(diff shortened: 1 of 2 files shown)


spec_14.rst
```diff
@@ -334,6 +334,18 @@ Some common system attributes are:
...
 
+**bank**
+   The value of the ``bank`` attribute is a string containing the name
+   of an accounting bank requested for this job. Actual support for the
+   ``bank`` attribute MAY depend on installation and configuration of
+   optional framework components, such as an accounting or other plugin.
+
+**project**
+  The value of the ``project`` attribute is a string containing the name
+  of an accounting project requested for this job. Actually support for
```
cmoussa1, line 345: "Actually" --> "Actual" here?

cmoussa1 (approved):
LGTM! Thanks for adding these. One minor spelling comment

spec_14.rst
garlick: IMHO I'd drop the word completely, e.g. just say "Support for the project attribute..."

garlick (approved):
LGTM too!

spec_14.rst
grondo: Removed that from both stanzas and will set MWP.

PR #422, opened 2024-08-08
rfc5: update module managment topic strings

garlick:
Problem: flux-framework/flux-core#6203 proposes to change the topic strings used for the module management RPCs.

Update the topic strings mentioned (somewhat superfluously) in the RFC.


grondo (approved):
LGTM!

PR #425, opened 2024-08-19
rfc27: document job feasibility service

### LINKED ISSUE #384: RFC 27: define sched.feasibility RPC
Problem: the `sched.feasibility` RPC is not defined in an RFC

Consider adding it to [RFC 27:  Flux Resource Allocation Protocol Version 1](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_27.html)

grondo:
This is a first cut at documenting the ``feasibilty`` service proposed in #384.

Fixes #384

### DIFF
--- a/spec_27.rst
+++ b/spec_27.rst
@@ -26,6 +26,7 @@ Language
 Related Standards
 *****************
 
+- :doc:`spec_3`
 - :doc:`spec_14`
 - :doc:`spec_16`
 - :doc:`spec_20`
@@ -98,6 +99,8 @@ Design Criteria
   tool (e.g. start time estimates).
 
 - Allow the expiration time of a resource allocation to be adjusted.
+
+- Detect unsatisfiable job requests at submission time.
 
 Implementation
 **************
@@ -644,3 +647,27 @@ a ``sched.free`` request.
 
 If the scheduler is monitoring job exceptions, it SHOULD NOT react in ways
 that might conflict with the job manager's actions.
+
+Feasibility
+===========
+
+A scheduler or other entity MAY register a generic ``feasibility`` service
+name through which unsatisfiable jobs may be detected at job submission.
+
+The ``feasibility`` service MAY be registered on one or more nodes to
+distribute the load of feasibility checks. The job ingest validator (the
+main user of the ``feasibility`` service) runs on all ranks and issues the
+``feasibility.check`` RPC with ``FLUX_NODEID_ANY``. The request is routed to
+a local ``feasibility`` service if available, or an upstream one per RFC 3.
+
+To determine the feasibility of a job request, a ``feasibility.check``
+request is sent with the following REQUIRED key:
+
+jobspec
+   (object) The jobspec for which feasibility should be checked.
+
+If the included jobspec could not ever be satisfied, even if all resources
+were available and ready, then the ``feasibility.check`` service SHALL
+respond with a error response including a human readable error string.
+
+The response SHALL consist of an empty payload on success.
--- a/conf.py
+++ b/conf.py
@@ -66,6 +66,7 @@ source_suffix = '.rst'
 linkcheck_ignore = [
     "https://help.github.com/en/pull-requests", # 403 Forbidden
     r"https://blog\.twitter\.com/.*", # 403 Forbidden
+    r"https://github.com/pmodels/mpich/commit/.*", # 406 client not acceptable
 ]
 
 
--- a/spell.en.pws
+++ b/spell.en.pws
@@ -491,3 +491,4 @@ bitmasks
 DoS
 lookups
 chu
+unsatisfiable


spec_27.rst
```diff
@@ -99,6 +99,8 @@ Design Criteria
 
 - Allow the expiration time of a resource allocation to be adjusted.
 
+- Allow out of band detection of unsatisfiable job requests.
```
garlick, line 102: Would it be more to the point to say "Detect unsatisfiable job requests at submission time"?

spec_27.rst
```diff
@@ -644,3 +646,29 @@ a ``sched.free`` request.
 
 If the scheduler is monitoring job exceptions, it SHOULD NOT react in ways
 that might conflict with the job manager's actions.
+
+Feasibility
+===========
+
+A scheduler or other entity MAY register a generic ``feasibility``
+service name through which the feasibility or satisfiability of jobs may
+be determined out of band, for example during job submission.
```
garlick, line 655: Like my previous comment, maybe "may be determined during job submission"?

spec_27.rst
```diff
@@ -644,3 +646,29 @@ a ``sched.free`` request.
...
+
+If the included jobspec could not ever be satisfied, even if all resources
+were available and ready, then the ``feasibility.check`` service SHALL
+respond with a error response including a human readable error string.
+
+If the job could ever be satisfied, then success SHALL be indicated via
+a response with the following REQUIRED key:
+
+errnum
+   (integer) An integer error number zero (0).
```
garlick, line 674: Oh weird,  I'm guessing the existing protocol requires `errnum` in the success response payload?  I wonder why we did that.

garlick (commented):
some quick comments

spec_27.rst
grondo: Good suggestion. How about "Allow the detection of unsatisfiable job requests at submission time" since the feasibility service and corresponding validation plugin are optional? If you'd prefer the shorter version and don't think it implies the detection of unsatisfiable jobs is required by the RFC, then I'm fine with that.

spec_27.rst
grondo: Sounds good. Note that the reason I left this statement more open is that there is nothing preventing use of the feasibility service and validator plugin outside of job submission (e.g. via `flux job-validator --jobspec-only`), so I was being a bit wary of specifying only at job submission in the RFC.

spec_27.rst
grondo: I think it was copying the requirements for the validator workers. Shall we change it for the `feasibilty.check` service? (Also if you have another suggestion for the name of the service, feel free to share it. I was just throwing something up to get this moving)

grondo: eh, looking at the `feasibility` validator plugin,  I don't think the `errnum` is actually required. It just happens to be in the `sched-simple` implementation for some reason.

https://github.com/flux-framework/flux-core/blob/dc2ef05918c8df0a9324d7fa36d5d6c85da98088/src/modules/sched-simple/sched.c#L703-L705

I'll remove it here since it doesn't seem necessary, and open a PR to remove it in sched-simple's implementation.

spec_27.rst
garlick: Sure that works for me.

spec_27.rst
garlick: Yup sounds good.

spec_27.rst
grondo: I just took your suggestions verbatim. I think shorter is better at this point, and if gets confusing in the future (which it probably won't) then it is easy to update the RFC.

grondo:
Ok, I've pushed changes that I believe address @garlick's comments (also had to add a URL to linkcheck ignore). This could use another pass.

garlick (approved):
LGTM!

I did have one thought but I'm not sure how to resolve it.  You mentioned that feasibility might run on more than one rank for scalability but since the feasibility client isn't discussed it's left open ended how that would get used.  Would it make sense to just tack on an explanatory  sentence that mentions that feasibility is normally called by the ingest validator which runs on all ranks and makes the RPC with FLUX_NODEID_ANY so it would work whether loaded on all ranks or just rank 0?   Or maybe as a boxed note?

Or maybe not at all!  It just seemed a little open ended to me.  I leave it to you to decide.

grondo:
I was wondering about that too. I wasn't sure if it was appropriate to go into that much detail in this RFC. I like your suggestion.

grondo:
> so it would work whether loaded on all ranks or just rank 0? 

Side note: in the case of the Fluxion feasibility service, there will be a non-trivial memory hit for running this service module, so I think the intent is not to run it on every rank, but possibly rank 0 and the login nodes? I wonder if the sentence "all ranks or just rank 0" should be modified to state something along the lines of

> ...the service SHALL be loaded on one or more ranks that are upstream of any rank where job submission is expected.

Or something?

garlick:
Works for me.

grondo:
Uhh, I took a stab at it, but not too happy with the result. Feel free to suggest editing.

garlick:
> , therefore the feasibility service SHALL be loaded on one or more ranks upstream of any rank where job submission is expected.

maybe instead

> .  The request is routed to a local `feasibility` service if available, or an upstream one per RFC 3.

grondo:
Thanks, I took your suggestion verbatim (sorry, having trouble with my prose today) and force-pushed the result.

grondo:
I guess I'll set MWP here since this was approved. We can always update with future PRs.

PR #428, opened 2024-10-10
rfc42: rename flags to forwarding_flags

chu11:
Problem: The exec request uses a field named "flags" to indicate what should be forwarded to the client.  The use of the generic term "flags" can be ambiguous, as it could be easily confused for other "flags" used within subprocess code.

Rename the field to "forwarding_flags".

----

side note, this is mostly b/c in the future we would like to pass the actual "subprocess" flags to the remote server.  So this is just mini-prep for that.  

Obviously can debate the name.  I considered "callback_flags" as well.

### DIFF
--- a/spec_42.rst
+++ b/spec_42.rst
@@ -108,7 +108,7 @@ are defined as follows:
 
     See `Command Object`_ below.
 
-  .. object:: flags
+  .. object:: forwarding_flags
 
     (*integer*, REQUIRED) A bitfield comprised of zero or more flags:
 


grondo:
Sorry if this was discussed elsewhere, but should we take a step back before breaking the protocol? Would it make sense to keep the current rexec flags and add a new rexec_flags member or mask the subprocess flags into the current `flags` member, rather than renaming?

Maybe my point is moot because we're breaking the protocol anyway?

chu11:
> Sorry if this was discussed elsewhere, but should we take a step back before breaking the protocol? Would it make sense to keep the current rexec flags and add a new rexec_flags member or mask the subprocess flags into the current flags member, rather than renaming?

Yeah, I guess I should have said that this is 100% optional.  Nothing stops us from leaving it as is and just adding a new field for the subprocess flags.    

We can leave things as is for now.  If we are going to break protocol, then we can reconsider this.

garlick:
I'd vote for adding something like `xflags` (for "extended" flags) and leaving the existing flags as is.

chu11:
per discussion, closing

PR #429, opened 2024-10-17
rfc15: describe IMP signal handling + minor updates

garlick:
Problem:  RFC 15 only describes `flux imp kill` but we need the IMP to forward signals.

Take a stab at describing this behavior, admittedly without a lot of detail as yet - perhaps we can fix that by coming to some agreement on those details.  But this is a start anyway.

### DIFF
--- a/spec_15.rst
+++ b/spec_15.rst
@@ -122,8 +122,8 @@ design
    which MAY be installed with setuid permissions in cases where multi-user
    Flux is required.
 
--  The IMP SHALL accept and process data using stdin, to avoid putting
-   sensitive data on the command line or environment.
+-  The IMP SHALL avoid putting sensitive data on the command line or
+   environment.
 
 Implementation of the IMP as a separately installed, setuid executable
 allows sysadmin control over where and how the IMP is enabled. If the
--- a/spec_15.rst
+++ b/spec_15.rst
@@ -338,9 +338,11 @@ IMP post-verification execution
 ===============================
 
 After verification of input is complete, the ``flux-imp`` executable
-invokes required job setup code as the superuser. This setup code SHALL
+invokes required job setup code as the superuser. This setup code MAY
 be implemented as system-installed and verified plugins, and MAY include
 such things as
+
+-  Start a PAM session on behalf of the guest
 
 -  Execution of some sort of job prolog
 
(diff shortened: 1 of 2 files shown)


spec_15.rst
```diff
@@ -372,13 +372,25 @@ A multi-user instance of Flux not only requires the ability to execute
 work as a guest user, but it must also have privilege to monitor and
 kill these processes as part of normal resource manager operation.
 
-Signaling and terminating jobs in a multi-user instance
```
grondo, line 369: commit message typo: forard

spec_15.rst
```diff
@@ -372,13 +372,25 @@ A multi-user instance of Flux not only requires the ability to execute
...
+as a proxy for the job by trapping common signals and forwarding them to
+the job shell.
+
+To enable the instance owner to fully clean up when the job shell is unable
+to do so, the IMP SHALL handle SIGUSR1 as a surrogate for SIGKILL.  Upon
+receipt of this signal, the IMP SHOULD deliver SIGKILL to all processes in
+the job's container, including the job shell.
+
+The mechanism by which processes are identified to receive SIGKILL is
+outside the scope of this document.
```
grondo, line 393: So is the consensus that we no longer need `flux-imp kill`? That would be kind of nice because it is a bit of code to maintain and has to read cgroup files to determine if the calling user has permission to signal a process.

It feels like we will need to add details here because there are two modes (at least) in which the IMP operates:
1. sdexec: The IMP should be the first processes in a cgroup created by the `flux` user systemd instance for the specific job. In this case the IMP can use `cgroup.procs` to gather the list of PIDs to forward signals.
2. without sdexec: The IMP is in a cgroup shared with the parent broker (system instance broker) and all other concurrently running jobs. It should _not_ use `cgroup.procs` to gather the list of PIDs to signal, since that would include everything Flux is running on the node including the broker.

If we can codify in the RFC how to differentiate these two environments, then the implementation of the IMP signal forwarding would be much simpler (and documented). For instance, to follow current practice we could state something like "The IMP shall get the basename of the current cgroup directory at startup. If the directory begins with `imp-shell`, then the IMP SHALL forward SIGKILL to all PIDs listed in `cgroup.procs`. Otherwise, the IMP SHALL forward SIGKILL only to its direct child and optionally MAY include descendants."

Ignore me if you already had this in mind (and is what you meant when you said details could be added later)

grondo (commented):
Thanks!! Just added one comment and a typo I happened to notice.

spec_15.rst
grondo: BTW, your succinct description of how this should work is excellent. Thanks!

garlick: I was hoping you would swoop in with those details and you did!  Let me put that in and we'll see how it looks.

Yes, I concur that `flux imp kill` isn't needed.   At least I can't think why we need it.  Dropping it will also simplify bulk-exec since the regular subprocess kill should work.

grondo: Great!

garlick: Oh wait a sec, when `flux imp run` is used, we might need `flux imp kill` unless we change that to work the same way.

grondo: Oh, good catch. :disappointed:

garlick:
Pushed those changes, thanks!

grondo (approved):
This LGTM. Once this is merged we can open an issue to implement the RFC in flux-security.

garlick:
I guess a question is should we address `flux imp run`?  We could have it wait for its children and forward signals too.  Or if we aren't going to do that we probably should add back `flux imp kill` to the RFC.

Actually let's deal with that in a follow on PR and get this merged since it's a somewhat independent issue.  I'll set MWP here.

PR #430, opened 2024-10-18
rfc15: describe flux imp run and how these processes are to be killed

garlick:
Problem: we've added `flux imp run` and `flux imp kill` but these are not described in RFC 15.

Describe `flux imp run`.

Instead of describing `flux imp kill`, propose a simplification where `flux imp run` lingers and forwards signals, analogous to the way `flux imp exec` was proposed to do for the shell.

There's also a bit of unrelated cleanup for RFC 15, split into separate commits.

### DIFF
--- a/spec_15.rst
+++ b/spec_15.rst
@@ -389,10 +389,10 @@ to do so, the IMP SHALL handle SIGUSR1 as a surrogate for SIGKILL.  Upon
 receipt of this signal, the IMP SHOULD deliver SIGKILL to all processes in
 the job's container, including the job shell.
 
-The IMP shall get the basename of the current cgroup directory at startup.
-If the directory begins with "imp-shell", then the IMP SHALL deliver SIGKILL
-to all PIDs listed in cgroup.procs. Otherwise, the IMP SHALL deliver SIGKILL
-only to its direct child and optionally MAY include descendants.
+The IMP SHALL get the basename of the current cgroup directory at startup.
+If the directory begins with ``imp-shell``, then the IMP SHALL deliver SIGKILL
+to all PIDs listed in ``cgroup.procs``. Otherwise, the IMP SHALL deliver
+SIGKILL only to its direct child and optionally MAY include descendants.
 
 IMP configuration
 =================
--- a/spec_15.rst
+++ b/spec_15.rst
@@ -372,6 +372,8 @@ A multi-user instance of Flux not only requires the ability to execute
 work as a guest user, but it must also have privilege to monitor and
 kill these processes as part of normal resource manager operation.
 
+.. _signal_handling:
+
 Signal Handling
 ---------------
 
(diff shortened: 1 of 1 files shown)


spec_15.rst
```diff
@@ -394,6 +394,16 @@ If the directory begins with ``imp-shell``, then the IMP SHALL deliver SIGKILL
...
 
+.. note::
+
+  The default system instance configuration is to launch the IMP as a direct
+  child of the broker, in the broker's cgroup.
+
+  Flux MAY be configured to run jobs in a ``flux`` owned systemd user
+  instance, enabling jobs to persist across a Flux restart.  In this
+  case the IMP, and hence the job, is started as a transient unit named
+  ``imp-shell-<rank>-<jobid>.service``, in its own cgroup.
```
grondo, line 407: I wonder if instead of documenting behavior of Flux, this RFC should specify it, i,e. instead of saying "In this case the IMP, and hence the job, is started as a transient.." the wording should be "In this case, the IMP and hence the job SHALL be started as a transient unit named ``imp-shell-<rank>-<jobid>.service``..." 

Or leave off specification of the entire name and just document that the unit SHALL begin with `imp-shell` which notifies the IMP that it SHOULD use a cgroup kill for SIGKILL?

Probably fine either way though!

spec_15.rst
```diff
@@ -429,6 +431,22 @@ allowed-environment
...
+.. note::
+
+  Flux MAY be configured to run prolog, epilog, and housekeeping scripts
+  as one-shot systemd services, which enables them to persist across a Flux
+  restart and run in a dedicated cgroup.
+
+  This is accomplished by configuring Flux-provided wrapper scripts as the
+  administrative executable.  The wrapper scripts trap SIGTERM and run
+  :program:`systemctl stop`, which tells systemd to terminate [#f2]_ all
+  processes in the unit's cgroup.
```
grondo, line 448: Should it be specified that Flux should avoid sending SIGKILL (via SIGUSR1) when a `flux run` target is a wrapper script for systemd service?

I guess also, since we're going to be working on the code here, we can take a step back and see if there's a better way to support run targets that are actually just systemd units.

grondo (approved):
Excellent! Thanks!

This LGTM. Questions inline could be addressed in future updates if preferred. (if updates are even needed.)

spec_15.rst
garlick: I'll add something about SIGKILL.

I'm drawing a blank right now on how to improve things with systemd units.  It is a bit icky-scripty right now.  Have to think about that one.  Maybe we can revisit if we have some good ideas?

garlick:
Repushed with those changes.  I guess the big question is whether we think it's reasonable to do away wtih `flux imp kill` in favor of the IMP lingering and forwarding signals in the run case.

spec_15.rst
grondo: Sounds good. I guess I was thinking something like we could design direct support for management of systemd units into the IMP. A new subcommand `flux-imp systemctl` could be used to solely to start/stop select systemd units, by configuration only. However, I think you're right, let's table that until we get the current issues resolved.

grondo:
> Repushed with those changes. I guess the big question is whether we think it's reasonable to do away wtih flux imp kill in favor of the IMP lingering and forwarding signals in the run case.

Yes that sounds good to me. Getting rid of code would be good.

garlick:
OK then, I'll go ahead and set MWP.  Apparently azure cloud or whatever can't get to the venti paper link even though I can from home, so the link check is failing.  I'll wait a bit in restart that builder.

PR #431, opened 2024-10-30
rfc15: IMP may linger post-shell to clean cgroup

garlick:
Problem: RFC 15 states that the IMP may exit after the job shell exits, but then it cannot forward SIGUSR1->SIGKILL to stuck processes if the shell exits early.

Change the wording to allow the imp to linger after the shell exits.


grondo (approved):
LGTM!

PR #432, opened 2024-11-15
RFC 14,25: add preemptible-after system attribute

garlick:
Problem: jobspec has no way to express preemptibility
    
Add the preemptible-after jobspec attribute.


grondo (approved):
This LGTM!

PR #433, opened 2024-11-15
rfc27: support partial hello responses

garlick:
This adds a `partial-ok` flag to the hello request, and (if that flag was true) an optional `allocated` idset key to the response which can indicate which ranks of R are allocated when some of the ranks have already been released.

The problem has been discussed in flux-framework/flux-core#6089

### DIFF
--- a/spec_27.rst
+++ b/spec_27.rst
@@ -153,6 +153,13 @@ userid
 t_submit
   (double) job submission time
 
+and the following OPTIONAL key:
+
+free
+  (string) An RFC 22 idset representing the ranks (execution targets)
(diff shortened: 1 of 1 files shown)


grondo (approved):
This looks straightforward to me, but likely you'll want an ACK from a Fluxion developer as well.

garlick:
After feedback from @trws, the `allocated` key is replaced with a `free` key, which makes more sense in retrospect.

grondo (approved):
LGTM! Since this got approval from @trws in principle, I think you could now set MWP.

PR #437, opened 2024-12-02
rfc7: break long test expressions to one per line

chu11:
Problem: We've adopted a defacto guideline that long test expressions connected by "AND" or "OR" operators should be broken to one test expression per line.  This is not documented.

Add this to RFC 7.

### DIFF
--- a/spec_7.rst
+++ b/spec_7.rst
@@ -86,6 +86,24 @@ to one parameter per line, at the same indent level.  Example:
                   FLUX_NODEID_ANY,
                   FLUX_RPC_RESPONSE);
 
+10. Long test expressions connected by "AND" or "OR" operators SHOULD
+be broken to one test expression per line at the same indent level.
+Example:
+
+  .. code-block:: c
+
+    if (out->type == MY_EXPECTED_TYPE
+        && out->bytes > LOW_END_RANGE
+        && out->bytes < HIGH_END_RANGE) {
+        ...
+    }
+
+  .. code-block:: c
+
+    if (flux_event_subscribe (h, "event.foo") < 0
+        || flux_event_subscribe (h, "event.bar") < 0) {
+        ...
+    }
 
 Variable Names
 ==============


cmoussa1 (approved):
LGTM!

spec_7.rst
```diff
@@ -86,6 +86,24 @@ to one parameter per line, at the same indent level.  Example:
...
 
+10. Long test expressions connected by "AND" or "OR" operators SHOULD
+be broken to one test expression per line at the same indent level.
+Example:
+
+  .. code-block:: c
+
+    if (out->type == MY_EXPECTED_TYPE
+        && (out->bytes > MIN_BYTES
+            && out->bytes <= MAX_BYTES)) {
```
garlick, line 97: er that doesn't appear to be at the same indent level?

garlick (approved):
LGTM except does example need to be fixed or maybe de-tabbed?

spec_7.rst
chu11: note the parens, its style is `if (A && (B &&C))` leading to the longer version

```
if (long-test-1
    && (long-test-2
           && long-test-3)) {
}
```

i added it to demonstrate that a long expression within an expression should also be broken up.

garlick: Oh sorry, I misread that.

However, do we want that?  I would not have put a line break on the second (compound) expression though.

chu11: ok, lets tweak for this example.  I changed it to

```
+    if (out->type == MY_EXPECTED_TYPE
+        && out->bytes > LOW_END_RANGE
+        && out->bytes < HIGH_END_RANGE) {
+        ...
+    }
```

chu11:
re-pushed, tweaking the example per conversation above

spec_7.rst
garlick: Controversy avoided :-)

PR #439, opened 2025-01-27
rfc20: clarify R_lite rank:host mapping

garlick:
Problem: the wording on how ranks and hostnames are ordered in R_lite is possibly a bit ambiguous, as discussed in flux-framework/flux-core#6584.

Make it very clear that the hostnames are in execution target rank order rather than the R_lite array order.


grondo (approved):
LGTM, Thanks!

PR #441, opened 2025-02-03
rfc25: clarify V1 restrictions in tasks section

sam-maloney:
Problem: the "Tasks" section of RFC 25 basically just points back to RFC 14, but in my testing there does seem to be several noteworthy restrictions on the "count" key that I couldn't find documented anywhere.

This PR is basically my attempt at clarifying what I would have found useful to have in the documentation, based on my best understanding of how the keys seem to behave in my testing. My apologies if I'm not totally correct on the true behavior, but hopefully it won't be too difficult for someone with more knowledge than me to tweak the wordings if necessary!

### DIFF
--- a/spec_25.rst
+++ b/spec_25.rst
@@ -143,20 +143,19 @@ key:
 
 -  count
 
-The ``count`` key SHALL contain at least one of the following keys:
+The ``count`` key SHALL contain exactly one of the following keys:
 
 -  per_slot
 
 -  total
 
-If both of these keys are set, the ``total`` key SHALL prevail. The definitions
-of these keys SHALL match those provided in :doc:`RFC 14 <spec_14>`, with the
-following restrictions in V1: if ``per_slot`` is used, its value MUST be one,
-and if ``total`` is used its value MUST be less than or equal to the ``count``
-key of the associated slot and greater than or equal to the number of allocated
-nodes (if no ``node`` resource vertex is explicitly given in the jobspec, then
-this minimum value will depend on the instance resource configuration and/or
-scheduler used).
+The definitions of these keys SHALL match those provided in
+:doc:`RFC 14 <spec_14>`, with the following restrictions in V1: if ``per_slot``
+is used, its value MUST be one, and if ``total`` is used its value MUST be less
+than or equal to the ``count`` key of the associated slot and greater than or
+equal to the number of allocated nodes (if no ``node`` resource vertex is
+explicitly given in the jobspec, then this minimum value will depend on the
+instance resource configuration and/or scheduler used).
 
 Attributes
 ==========
--- a/spec_25.rst
+++ b/spec_25.rst
@@ -130,10 +130,12 @@ well. Therefore, the complete enumeration of valid resource graphs in V1 is:
 Tasks
 =====
 
-The value of the ``tasks`` key SHALL be a strict list which MUST define exactly
-one task. The list element SHALL be a dictionary representing a task to run as
-part of the program. A task descriptor SHALL contain the following keys, whose
-definitions SHALL match those provided in RFC14:
+The ``tasks`` key SHALL be a strict list which MUST define exactly one task.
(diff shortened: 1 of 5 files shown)


spec_25.rst
```diff
@@ -133,25 +133,29 @@ Tasks
...
 
-   -  per_slot
+The ``count`` key SHALL contain at least one of the following keys:
 
-   -  total
+-  per_slot
+
+-  total
+
+If both of these keys are set, the last key parsed SHALL prevail. The definitions of these keys SHALL match those provided in :doc:`RFC 14 <spec_14>`, with the following restrictions in V1: if ``per_slot`` is used, it's value MUST be one, and if ``total`` is used it's value MUST be less than or equal to the ``count`` key of the associated slot and greater than or equal to the number of allocated nodes (if no ``node`` resource vertex is explicitly given in the jobspec, then this minimum value will depend on the instance resource configuration and/or scheduler used).
```
grondo, line 150: I wonder if we want to explicitly forbid both `total` and `per_slot` being used, then update the current implementation to enforce it.

sam-maloney: I would personally agree on that being a sensible restriction to make the (intended & actual) behavior of a given jobspec as clear as possible!

sam-maloney: After looking into this a bit further I realized that I was an idiot in my first go-round and was actually overwriting the keys in my Python script, hence why only the last one I specified was being used... mea culpa!

It turns out the actual behavior was that the `total` is used if both are specified.

sam-maloney: I've also pushed a change reflecting the possible change to exactly one of the keys being required.

sam-maloney: @grondo My follow-up question would then be whether this is just a V1 restriction, or if the RFC14 canonical jobspec should also be tweaked to require that exactly one of per_slot/per_resource/total be set?

grondo: That sounds right to me, but I'd like to get @trws opinion since Fluxion already supports more of canonical jobspec. There may be something I'm missing.

Thanks!

trws:
I'm perfectly ok with limiting to one of "per_slot" or "total" but why is this restricting "per_slot" to only support 1 as a value in the json schema? If that's all we support right now then it makes sense, and we could mention the limitation, but it was set up as a number to support over-subscription or allowing multiple tasks to "roam" within the resource bindings of a slot.

sam-maloney:
> I'm perfectly ok with limiting to one of "per_slot" or "total" but why is this restricting "per_slot" to only support 1 as a value in the json schema? If that's all we support right now then it makes sense, and we could mention the limitation, but it was set up as a number to support over-subscription or allowing multiple tasks to "roam" within the resource bindings of a slot.

Yes, it's just documenting what the shell currently handles; anything other than 1 triggers an error: https://github.com/flux-framework/flux-core/blob/master/src/shell/jobspec.c#L281

trws:
Ok, I'm alright with the schema change then but would like to have it noted in the RFC that it's limited to one, maybe "a positive integer, currently restricted to 1" or similar?

sam-maloney:
> Ok, I'm alright with the schema change then but would like to have it noted in the RFC that it's limited to one, maybe "a positive integer, currently restricted to 1" or similar?

That wording looks like it's coming from RFC14, but I don't think we want to restrict it in the Canonical Jobspec? This PR was only meant to clarify the actual resrictions present in V1 (or at least the current implementation). In `spec_25.rst` lines 153/154 I'd added "...with the following restrictions in V1: if ``per_slot`` is used, its value MUST be one..."

https://github.com/flux-framework/rfc/pull/441/files#diff-c0c7d9cda327a10f4b10a35105130a3d5de8e2afa0327211631ea89dc2267f71R154

Hopefully that's what you're after :)

trws:
Yup, that’s it, thanks!

trws (approved):

grondo (approved):
LGTM as well, thanks! Setting MWP.

PR #442, opened 2025-02-07
rfc14: allow idset and range strings for resource count

sam-maloney:
Problem: irregular ranges of resource counts that do not fit the current min/max/operator/operand paradigm cannot be specified.

Allow a third option (in addition to single integer and range dict) for resource counts to be a string containing an RFC22 idset of acceptable counts.

Obviously new operators could be defined for some cases (including the simple example I included), but for future-proofing it would seem desirable to have a fallback for users to specify any arbitrary set of counts they desire.

[SLURM allows similar strings for its node counts](https://slurm.schedmd.com/salloc.html#OPT_nodes) so this would ensure flux ends up at least as capable as SLURM in this regard ;)

(idset regex created/tested via https://regex101.com/r/65FyDf)

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,7 @@ Table of Contents
 - [42/Subprocess Server Protocol](spec_42.rst)
 - [43/Job List Service](spec_43.rst)
 - [44/Resource Events](spec_44.rst)
+- [45/Resource Range String Representation](spec_45.rst)
 
 Build Instructions
 ------------------
--- a/data/spec_14/schema.json
+++ b/data/spec_14/schema.json
@@ -23,6 +23,11 @@
       },
       "additionalProperties": false
     },
+    "idset": {
+      "description": "an RFC 22 idset",
+      "type": "string",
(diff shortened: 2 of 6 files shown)


grondo:
Good idea! I think @trws should comment since this would impact Fluxion mostly.

trws:
I don't have a problem with this, we currently don't really _do_ anything with the range, but that's an active area of research so it might change and it shouldn't add significant complexity to anything we do with it.

sam-maloney:
This has now been rebased on top of #453 to also allow compact string representation of ranges, and initial implementation is included in flux-framework/flux-core#6832

grondo (approved):
Thanks @sam-maloney, I'll set MWP here.

grondo:
> it might be worth just adding r'https://www.gnu.org/.*' to the linkcheck_ignore array so at least they don't keep choking up the CI...

Yeah, that is starting to sound like a reasonable approach. An alternative would be to make a linkcheck failure just a warning...

PR #445, opened 2025-02-14
rfc44: add truncate event

grondo:
Problem: RFC 44 does not provide for the possibility of truncating the resource journal of events.

Add a truncate event to RFC 44 for this purpose.


garlick (approved):
LGTM! Thanks!

PR #446, opened 2025-02-14
rfc44: add optional reason field to undrain event

grondo:
Problem: At least one documented use case could make use of a reason field in the undrain event, but RFC 44 does not allow that.

Add an optional reason key in the undrain context.


garlick (approved):
Thanks!

PR #447, opened 2025-02-18
RFC 28: add `shrink` key to resource acquisition response

grondo:
This PR adds the `shrink` key to RFC 28 as suggested in flux-framework/flux-core#6641.
This key contains an idset of execution targets which have been removed from the instance so should no longer be considered for scheduling.

### DIFF
--- a/spec_28.rst
+++ b/spec_28.rst
@@ -55,9 +55,6 @@ module to the scheduler.  The responses to this RPC define the resource
 set available for scheduling, and mark targets *up* or *down* as
 availability changes.
 
-Version 1 of this protocol supports a static resource set per Flux instance.
-Resource *grow* and *shrink* are to be handled by a future protocol revision.
-
 Design Criteria
 ***************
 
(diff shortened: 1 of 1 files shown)


grondo:
One thought here: I wonder if, for better backwards compatibility, it should be ok for an execution target to be in both the `down` and `shrink` idsets, and if shrink is supported, that should take precedence? This would allow a scheduler that doesn't yet support `shrink` to continue to work with the updated protocol (since in this PR at least, the protocol version hasn't been updated)

grondo:
Done. I've added:

> For backwards compatibility, targets in the ``shrink``
>  key SHALL also appear in the ``down`` key of the same response. If a
>  scheduler supports ``shrink`` then the ``shrink`` key SHALL take precedence.

spec_28.rst
```diff
@@ -167,15 +171,16 @@ Example:
...
    }
 
-If down resources are assigned to a job, the scheduler SHALL NOT raise an
-exception on the job.  The execution system takes the active role in handling
-failures in this case.  Eventually the scheduler will receive a ``sched.free``
-request for the offline resources.
+If removed (``shrink``) or down resources are assigned to a job, the
+scheduler SHALL NOT raise an exception on the job. The execution system
+takes the active role in handling failures in this case. Eventually the
+scheduler will receive a ``sched.free`` request for the offline resources.
```
garlick, line 183: Perhaps we should clarify that we'll send a `sched.free` for resources  even when they have been removed?   This reads a bit ambiguous to me in that regard, but it should be required for backwards compatibility.

garlick (approved):
LGTM, thanks!  One comment below - feel free to ignore.

spec_28.rst
grondo: Agree. Would this be sufficient:
```suggestion
If removed (``shrink``) or down resources are assigned to a job, the
scheduler SHALL NOT raise an exception on the job. The execution system
takes the active role in handling failures in this case. Eventually the
scheduler will receive a ``sched.free`` request for the offline or removed resources.
```

grondo: I made this change and will set MWP. Thanks!

PR #449, opened 2025-03-01
rfc11: support optional size key

chu11:
Problem: It would be convenient to know the size of the data in a val or valref object without decoding/reading the data.  However there is no metadata for that.

Support an optional size key in 'val' and 'valref' objects.  It allows us to know the size of the data without wasting cycles decoding or reading the data.

### DIFF
--- a/spec_11.rst
+++ b/spec_11.rst
@@ -56,8 +56,8 @@ Implementation
 In contrast to the original KVS prototype, in this specification, KVS
 values are unstructured, opaque data.
 
-All tree objects SHALL be represented as JSON objects containing three top
-level names:
+All tree objects SHALL be represented as JSON objects containing three
+REQUIRED top level names:
 
 -  *ver*, an integer version number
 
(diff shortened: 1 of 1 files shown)


spec_11.rst
```diff
@@ -82,6 +82,12 @@ There are six types:
 -  *symlink*, data is an object containing a target key and optionally
    a namespace.
 
+The following OPTIONAL names are supported:
+
+-  *size*, contains the size of the decoded data within a *val* object
+   or the total size of data stored in the blobrefs of a *valref* object.
+
+
```
garlick, line 90: This seems like it could be neat.

A thought on adding size to dir and dirref: since every KVS metadata update involves updating all parent metadata back to the root, we could use that fact to keep a cumulative size in dir and dirref objects.  That is, a directory would contain the sum of the sizes of files and directories it contains.  The root would then contain the total size of the namespace.  These sizes could be a little bit misleading due to the automatic de-duplication inherent in the content design, but it still might be useful and it would accurately reflect the size of the subtree if dumped with `flux-dump(1)`.

PR #455, opened 2025-07-03
rfc10: support validate target

chu11:
Problem: It would be convenient if the content service could simply validate if a blob digest is valid or not.  Without such a service, the "load" service could be used, but that would be very costly for a simple check since all data is sent in the response.

Add a validate target to the content service.

---

Note: see https://github.com/flux-framework/flux-core/issues/6825


garlick (approved):
LGTM

Issue #456, opened 2025-07-14
linkcheck failing


chu11:
builders are currently failing due to a failed linkcheck

```
(         spec_13: line  904) broken    https://www.mcs.anl.gov/papers/P1760.pdf - 403 Client Error: Forbidden for url: https://www.mcs.anl.gov/papers/P1760.pdf
```

it's not clear to me why this is happening.  The link works fine in a browser, but does fail when doing `make linkcheck` on the command line (atleast on my personal laptop).  perhaps should setup an "ignore" config for this specific url.

PR #459, opened 2025-07-30
Fix/update links throughout

sam-maloney:
Problem: Some links are intermittently timing out during the sphinx linkcheck, or are otherwise "out-of-date"

This PR implements several changes, which are all related to links and/or linkcheck, but are otherwise independent, so I have tried make each commit a standalone change in case not all of them are desired.

1. The https://www.mcs.anl.gov/papers/P1760.pdf link was consistently failing in recent CI runs, as per #456. It seems the `User-Agent` header is the problem, so instead of ignoring the link as in #457, this sets a working header. I initially tried spoofing a browser but wasn't successful; however, testing directly with the `python-requests` module did succeed, so I just set the header to a generic version of its `User-Agent` i.e. `python-requests`

2. The 2 links to https://www.gnu.org/licenses/* were always being rate-limited and often timing out in CI, e.g. [here](https://github.com/flux-framework/rfc/actions/runs/15179769618/job/44371499742) and [here](https://github.com/flux-framework/rfc/pull/451#pullrequestreview-2681987951). For this domain spoofing a browser `User-Agent` helps, albeit only in conjunction with setting a more permissive `Accept=*/*` header (which I made a default since it also seemed to reduce the number of rate-limits I got for links to the github.com domain).

3. Two github.com links were in the ignore list because previously GitHub was also much more picky about what `User-Agent` they would accept, but both links seem to validate fine now without any further change, so I removed them from the ignore list (the Twitter/X blog still fails due to a Cloudflare bot challenge, which uses Javascript, and python-requests doesn't support Javascript, so I couldn't find a simple solution other than continuing to ignore it).

4. A number of links have been permanently redirected, many of them simply from http to https, so I updated all links to their final target and to all use https. Everything passes without this, but it cleans things up a lot (and made the next change more noticeable).

5. The Google Drive link regarding "Process Management in MPICH Draft 2.1" redirects me to a page to request access, which isn't the most immediately helpful thing... I changed it to link to a [wiki page on the MPICH GitHub](https://github.com/pmodels/mpich/blob/main/doc/wiki/notes/pm.md#notes) that seems to contain the same information the reference was referring to.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -67,6 +67,10 @@ linkcheck_ignore = [
     r'https://blog\.x\.com/.*', # 403 Forbidden, Cloudflare bot challenge
 ]
 
+linkcheck_anchors_ignore_for_url = [
+    r'https://github\.com/.*',
+]
+
 linkcheck_request_headers = {
     'https://www.mcs.anl.gov/': {
         'User-Agent': 'python-requests'
--- a/spec_13.rst
+++ b/spec_13.rst
@@ -891,7 +891,7 @@ the mapping is unknown.
 References
 **********
 
-.. [#f1] `Process Management in MPICH Draft 2.1 <https://drive.google.com/file/d/0B273EWJxZUxsbS15SEkzZGtXU2c/view?usp=sharing>`__
+.. [#f1] `Process Management in MPICH <https://github.com/pmodels/mpich/blob/main/doc/wiki/notes/pm.md#notes>`__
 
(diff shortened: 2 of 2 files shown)


chu11:
@wihobbs I'm not sure.  Maybe something racy happened behind the scenes?  Lets kick it and see if that fixes things.

chu11:
@mergifyio refresh

chu11:
@Mergifyio requeue

chu11:
@Mergifyio dequeue

chu11 (approved):

PR #462, opened 2025-08-08
ci: pass with comment on linkcheck fail

sam-maloney:
Problem: linkcheck fails frequently for various out-of-our-control reasons

Make the workflow pass regardless of `make linkcheck` outcome, but post a comment to the PR notifying if linkcheck failed.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -43,5 +43,20 @@ jobs:
       run: |
         pip3 install --upgrade -r ./requirements.txt
     - name: make linkcheck
+      id: linkcheck
       run: |
-        make linkcheck
+        status=0
+        make linkcheck || status=$?
+        echo "status=${status}" >> "${GITHUB_OUTPUT}"
+        exit 0 # this step should never fail the overall workflow
+    - name: Comment on PR if linkcheck failed
+      if: ${{ github.event_name == 'pull_request' && steps.linkcheck.outputs.status != 0 }}
+      uses: actions/github-script@v7
+      with:
+        script: |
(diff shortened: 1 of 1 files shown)


grondo:
Another simpler approach would be to drop the linkcheck requirement from `.mergify.yml` and just treat the error as a warning. Though I do like the idea of a PR comment much better, perhaps the 2nd workflow is overkill in this situation.

I'm fine either way, though. Perhaps others can chime in on their preferences.

In any event, dropping the explicit required checks in `.mergify.yml` in preference for branch protections like we've done in other projects is probably warranted here, so I can make a PR for that.

sam-maloney:
I completely agree that it's overkill, haha! But it was the only way I could find to have the workflow show as 'passing' even if the linkcheck failed while still actually providing a warning that it had failed, and by that point I was in too deep and figured I'd just try implementing it to learn more about GitHub's CI :grin: Feel free to just make a simpler PR and close this though, if you prefer! 

Other options I saw were just to use `make linkcheck || true` to always pass, but then you have to manually check the logs to even see the status; or as you suggested, make the linkcheck optional for mergify so it doesn't actually stop merging, but then the workflow will be frequently displayed as failing (which isn't the end  of the world, but does seem to defeat the purpose of the status marks if every merged PR has a red X next to it!)

grondo:
I'm going to rebase then merge this @sam-maloney. Hope that's ok.

sam-maloney:
> I'm going to rebase then merge this @sam-maloney. Hope that's ok.

Always OK with me to rebase/edit/mangle my PRs in any way you see fit, haha! :grin:

PR #465, opened 2025-08-15
command line jobspec resources (alternative to PR 371)

sam-maloney:
Problem: we cannot define arbitrary jobspec resources lists easily on the command line

Define a compact string representation that can be mapped to a full jobspec resources request

Apologies for stepping on any toes, but I was reading through @vsoch's draft in #371 and the discussion in flux-framework/flux-core#2213 and this idea started forming in my head, and wouldn't leave me alone until I had written it down, and then it snowballed into this, hahaha!

I'm not sure if people are still interested in this feature or not, but figured I'd open the PR in any case. And I think my main takeaway would be that whatever is settled on it shouldn't be V1 specific, otherwise it's just one more thing that needs to be transitioned eventually. This syntax makes simple requests very simple and compact, but imposes no restrictions on arbitrarily complex RFC14 compliant resources lists. Of course at some point it may make little sense to ever express such things directly on the command line, but one might as well avoid restricting anything needlessly!

And the fundamental idea could likely be extended to full jobspec requests instead of just resources if that was ultimately desirable, but would likely make more sense as a later step/addition.

There's a prototype parser on a [branch in my flux-core fork](/sam-maloney/flux-core/tree/cli-jobspec-resources-parser) modeled after the constraint parser.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ Table of Contents
 - [43/Job List Service](spec_43.rst)
 - [44/Resource Events](spec_44.rst)
 - [45/Resource Range String Representation](spec_45.rst)
+- [46/Command Line Job](spec_46.rst)
 
 Build Instructions
 ------------------
--- a/index.rst
+++ b/index.rst
@@ -294,6 +294,12 @@ Changes to resource availability are recorded in an RFC 18 eventlog.
 This specification describes a compact form for expressing an RFC 14
 resource range defined by a min/max/operand/operator combination.
 
+:doc:`spec_46`
+~~~~~~~~~~~~~~
+
+This specification describes a compact form for expressing an RFC 14 jobspec
+resources list that can be provided to several Flux commands by a user.
+
 .. Each file must appear in a toctree
 .. toctree::
    :hidden:
@@ -341,3 +347,4 @@ resource range defined by a min/max/operand/operator combination.
    spec_43
    spec_44
    spec_45
+   spec_46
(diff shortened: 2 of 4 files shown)


grondo:
One idea for now  -- now that we have CLI plugins -- is to implement this as a CLI plugin that users requiring the more advanced specification can install and use. I think @wihobbs was talking about working on something similar that acts as a filter, but embedded in the cmdline via a plugin with a unique option namespace is an enticing way to offer the functionality.

Perhaps this would allow us to offer something without fully committing, eventually perhaps absorbing the implementation as a native option or new command when we feel it working as desired.

sam-maloney:
I think @garlick's example was what I had in mind, i.e. a single option (in my mind I was visualizing `--resources=...`) to precisely specify the desired resources (obviously one could just have such an option directly accept a valid JSON/YAML string representation, but that seems very verbose for the CLI).

I also agree with what @vsoch is saying, but it feels to me more like the current interface of specifying things through a set of more "intuitive" options (like `--nodes` `--ntasks` `--cores-per-task`) which I agree is _very_ useful for most users, and the evolution of that will be a challenging design problem for more complicated jobspec requests. But I feel like that's perhaps complementary to the idea here, which would target more advanced users who have (or are willing to learn) more of an understanding of the flux resource model and jobspec and want a way to directly specify exactly what they want.

So I guess what I'm saying is _both_ might have a place: something like this as the compact, direct, _power-user_ option, and then an evolution & expansion of the current options for the verbose, intuitive, _basic-user_ mode. :man_shrugging:

garlick:
@mergify-io rebase

PR #466, opened 2025-08-19
rfc20: add `nslots` to Rv1 spec

sam-maloney:
Problem: the number of slots decided on by the scheduler is not currently recorded in its allocated R

Add an (optional) nslots key to the execution dict of the Rv1 spec which SHOULD be included in a scheduler allocation

This update documents the design change proposed in flux-framework/flux-core#6632 ([specifically this comment](https://github.com/flux-framework/flux-core/pull/6632#issuecomment-2720404016))if this idea seems like it would also be feasible to implement in `fluxion`

### DIFF
--- a/data/spec_20/example1.json
+++ b/data/spec_20/example1.json
@@ -13,6 +13,7 @@
     "nodelist": [
       "node[186-189]"
     ],
+    "nslots": 32,
     "starttime": 1676560542,
     "expiration": 1676562342
   }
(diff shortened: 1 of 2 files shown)


trws:
That's a fair point.  As I said I don't see any problem with it, it's easy enough to supply.   @milroy do you have any thoughts here?

grondo (approved):
Yes, looks good and lets get this merged!

I agree about making the nslots required so we can remove the conditional logic eventually in the job shell.

PR #469, opened 2025-09-26
rfc42: add lflags to exec protocol

garlick:
Problem: there is no way to pass the subprocess API flags to remote subprocesses.

Add a separate flags field called "lflags" for this purpose.

See also flux-framework/flux-core#7086


grondo (approved):
Nice, LGTM!

PR #470, opened 2025-09-30
WIP: rfc42: add background flag to exec protocol

grondo:
This is a first cut at adding a `background` flag to the RFC 42 exec protocol.

Marking WIP because there are other ways to do this. Another approach for example would be a separate RPC topic string for background process execution that implements a non-streaming RPC -- where the response contains either the pid and rank of the subprocess, or an exec error. But I thought I'd throw up a version using a flag since this can probably result in the re-use of more code.

After playing around with the code for a bit. Probably the best implementation would be a new API function:
```c
flux_future_t *flux_rexec_bg (flux_t *h, int rank, flux_cmd_t *cmd);
```

### DIFF
--- a/spec_42.rst
+++ b/spec_42.rst
@@ -97,8 +97,10 @@ Implementation
 exec
 ====
 
-The streaming :program:`exec` RPC creates a new subprocess.  Payloads are
-are defined as follows:
+The exec RPC creates a new subprocess. The RPC MUST be initiated with a
+streaming request as defined in RFC 6 unless the ``background`` flag is set
+(see below), in which case it MUST be initiated with a non-streaming request.
+Payloads are defined as follows:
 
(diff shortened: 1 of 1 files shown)


garlick (approved):
LGTM!

Oh, this probably needs an update up top
> The streaming exec RPC creates a new subprocess. Payloads are are defined as follows:

Maybe
> The exec RPC creates a new subprocess.   The RPC is REQUIRED to be initiated with a streaming request, as defined in RFC 6, unless the `background` flag is set (see below), in which case it MUST be initiated with a non-streaming request.
Payloads are defined as follows:

grondo:
Thanks! Took your suggestion with trivial changes (use MUST in both clauses) and pushed the squashed result. I'll set MWP and go make requisite changes in the PR.

PR #471, opened 2025-10-01
requirements: require `urllib3>=2.5.0`

cmoussa1:
#### Problem

GitHub Dependabot reported a security vulnerability with `urllib3 < 2.5.0`, which the RFC repo uses.

---

This PR upgrades the `urllib3` version to be at least `2.5.0`.


cmoussa1:
Woohoo! No problem. It's a little discouraging that this change is going to be made to all of the flux-framework projects that use mergify, but it seems to be pretty trivial thus far 😅

PR #473, opened 2025-10-07
rfc42: use RPC streaming flag instead of explicit background flag

grondo:
The more I thought about @garlick's idea to drop the `background` flag, the more I liked it, so I thought I'd attempt changes to the RFC to implement this idea.


grondo:
Thanks! I made the suggested change and will set MWP.

PR #474, opened 2025-10-07
mergify: fix wip regex

grondo:
Problem: The condition in the mergify config which attempts to prevent WIP PRs from being merged has an incorrect regex.

Rewrite the regex without quotes so it will match PR titles. Simplify the expression using a case insensitive match, and only match one or zero brackets instead of zero or more.


garlick (approved):
LGTM!

I'll press the merge button.

PR #476, opened 2025-10-17
rfc42: add optional subprocess labels

grondo:
Problem: There is no convenient way to refer to subprocesses launched using the protocol specified in RFC 42.

Introduce an optional subprocess label which, when specified, can be used to refer to a subprocess by name instead of PID.


garlick (approved):
LGTM!  One suggestion inline.  Fine to ignore.

PR #480, opened 2025-12-01
feat: code of conduct

vsoch:
Problem: to join HPSF and encourage Flux to have a more broad community, we need a Code of Conduct.
Solution: add rfc spec 47 for a code of conduct.

I wasn't sure who to put for maintainers. I figure we should have one person of each gender, minimally, for different kinds of complaints. I can adjust / add / remove as requested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ Table of Contents
 - [44/Resource Events](spec_44.rst)
 - [45/Resource Range String Representation](spec_45.rst)
 - [46/Command Line Job](spec_46.rst)
+- [47/Flux Framework Contributor Code of Conduct](spec_47.rst)
 
 Build Instructions
 ------------------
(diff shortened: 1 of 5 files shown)


vsoch:
> I'm a little worried about the list of email addresses being repeated in various documents and going stale. I wonder if we could find a way to centralize them or provide a generic mail ailas? Maybe in this case contacts shouldn't be anonymized though.

I was thinking about this - I've had to file a report twice (unfortunately) and both times it was to a generic list. That's logistically easier, but I knew would be going to a small group (of both genders). Recollecting that, I don't know if this is politically incorrect to say, but I really wish I could have contacted a woman first. The way the communication comes in would likely vary between those two situations. That's at least why I made the choice here. For governance, we likely want more of a group list.

garlick:
That makes sense!

Well regarding the duplicate references that go stale, here's a thought.  Maintainers that are willing to be contacted could add their emails to their github profiles, and then we could create a "maintainers" team at https://github.com/orgs/flux-framework/teams that we could point to?

garlick:
Yes!

PR #482, opened 2025-12-03
rfc43: update job-list.list to streaming RPC

chu11:
Problem: job-list.list RPC currently responds to requests with a single large array of jobs.  This response can be extremely large, and could lead to head-of-line blocking.

Update the job-list.list RPC to be a streaming RPC.  It will instead stream jobs back to the caller.

### DIFF
--- a/spec_43.rst
+++ b/spec_43.rst
@@ -272,7 +272,7 @@ List
 
 The :program:`job-list.list` RPC fetches a list of jobs.
 
-The list of jobs shall be filtered in the following order.
+The list of jobs SHALL be filtered in the following order.
 
 - pending jobs
 - running jobs
@@ -281,6 +281,12 @@ The list of jobs shall be filtered in the following order.
 Pending jobs are returned ordered by priority (higher priority first),
 running jobs ordered by start time (most recent first), and inactive
 jobs ordered by completion (most recently finished first)
+
+The RPC request SHOULD specify FLUX_MSGFLAG_STREAMING as described in
+RFC 6 so that the server has the opportunity to manage response
+message size for large query results.  When that flag is specified,
+responses SHALL be split over multiple messages, terminated with error
+61, "No data available" (ENODATA).
 
 The RPC payloads are defined as follows:
 
@@ -313,7 +319,8 @@ The RPC payloads are defined as follows:
 
 .. object:: job-info.lookup response
 
-  The response SHALL consist of a JSON object with the following keys:
+  Each non-error response SHALL consist of a JSON object with the
+  following keys:
 
   .. object:: jobs
 


chu11:
> I'm talking about one version of Flux talking to another, e.g. as would happen with flux top or flux pstree if a subinstance is launched using a different version. We've been pretty good about keeping interfaces compatible up to this point.

I brought up this question in the flux-core issue.  Simply maintaining old behavior if the RPC isn't a streaming RPC would be trivial.  I just didn't know if that's something we considered important for this particular RPC.

The example you bring up, interaction between older version vs new versions of Flux is a good example use case.  We should definitely keep that working like before.  I'll update (both RFC and code)

chu11:
re-pushed, adding backwards compatibility.  I wasn't sure how to organize things.  So I just added a blurb at the end of the section saying backwards compatibility MAY be supported via the old response format.

grondo (approved):
LGTM!

spec_43.rst
```diff
@@ -270,9 +270,9 @@ In these examples the caller wants to filter jobs submitted to the queue foobar
 List
 ====
 
-The :program:`job-list.list` RPC fetches a list of jobs.
+The :program:`job-list.list` streaming RPC fetches and returns jobs.
 
-The list of jobs shall be filtered in the following order.
+The jobs to be streamed shall be filtered and returned in the following order.
```
garlick: Maybe s/jobs to be streamed/jobs/

garlick (approved):
Some suggestions inline but not critical.

spec_43.rst
```diff
@@ -313,13 +316,30 @@ The RPC payloads are defined as follows:
...
 
   .. object:: jobs
 
-    (*array of objects*, REQUIRED) A list of the jobs returned from
-    the request.  Each object will contain the requested attributes in
-    an object described in :ref:`spec_43_job_attributes`.
+    (*array of objects*, REQUIRED) A list of all the jobs matching
+    criteria from the request.  Each object will contain the requested
+    attributes in an object described in
+    :ref:`spec_43_job_attributes`.
```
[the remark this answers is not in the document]
garlick: Oh crud, my main comment must not have been added.  Let me recreate.

spec_43.rst
```diff
@@ -282,6 +282,9 @@ Pending jobs are returned ordered by priority (higher priority first),
 running jobs ordered by start time (most recent first), and inactive
 jobs ordered by completion (most recently finished first)
 
+The RPC stream SHALL be terminated with error 61, "No data available"
+(ENODATA) when all jobs have been sent.
+
```
garlick: Suggestion

> The RPC request SHOULD specify FLUX_MSGFLAG_STREAMING as described in RFC 6 so that the server has the opportunity to manage response message size for large query results.  When that flag is specified, responses SHALL be split over multiple messages, terminated with error 61, "No data available" (ENODATA).

chu11:
I don't think there are any gotchas, will set MWP

PR #484, opened 2025-12-04
rfc1: drop Project Administration section

garlick:
Problem: the Project Administration section of RFC 1 was inherited from 0MQ and is neither sufficient nor correct for flux-framework.

Remove that section from RFC 1.
A new project governance RFC will cover this topic.


grondo (approved):
LGTM!

PR #485, opened 2025-12-10
rfc43: update roles section

garlick:
@grondo, @trws, and I had a chat today about this and wanted to add Administrator as a new role in the gov doc.  This would be the people who are able to add/remove projects from the flux-framework organization and do other "root" like tasks within it.

This defines the new role and tries to distinguish between "the project" which is really the organization, and the individual projects within it which follow the rules, but may have different maintainers for example.  (They also may have different policies for development, within the constraints of RFC 1).

Anyway, @trws or @grondo please make sure I captured what we discussed.

p.s. I also added a README.md for the organization that lists me, @grondo, and @trws as the administrators and this doc links to that page.

### DIFF
--- a/spec_48.rst
+++ b/spec_48.rst
@@ -90,14 +90,14 @@ Maintainer Review
 Issue Governance
 ----------------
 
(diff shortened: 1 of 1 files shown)


grondo (approved):
This looks good to me. One suggestion and a couple questions inline.

PR #488, opened 2025-12-12
rfc43: support job-list.list response version

chu11:
Problem: The job-list.list service is forwards compatible, where older non-streaming clients can communicate with newer services. However it is not backwards compatible, where newer streaming clients can operate with older services.

Add a version field to the "job-list.list" RPC response.  This can inform newer clients on what protocol behavior is expected from both older and newer job-list.list services.


grondo (approved):
LGTM!

PR #489, opened 2026-01-07
rfc5: update broker module specification

garlick:
Problem: RFC 5 is pretty of date at this point.

Make an update pass.

### DIFF
--- a/spec_5.rst
+++ b/spec_5.rst
@@ -5,6 +5,10 @@
 5/Flux Broker Modules
 #####################
 
+.. highlight:: c
+
+.. default-domain:: c
+
 This specification describes the broker extension modules
 used to implement Flux services.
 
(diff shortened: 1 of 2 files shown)


spec_5.rst
```diff
@@ -27,205 +31,419 @@ Related Standards
...
-exit.  The module SHALL wait for a response to this message before exiting
-``mod_main()``.
+If unspecified, the module name MAY be derived from the path name used to
+load it.
 
-Example payload:
+The module name SHALL be registered as a default service provided by the
+module.  This means that request messages with topic strings that begin
+with ``NAME.`` are routed to the module for handling.  Additional services
+MAY be dynamically registered by the module.
```
chu11: perhaps add an example, like `kvs.lookup` and `kvs.commit` are both routed to the `kvs` module?

PR #490, opened 2026-01-08
code of conduct: update contacts

vsoch:
We should add back the more explicit list of emails and admin contacts. The team link we initially created will always be private to org members only. The link (as it is now) if you test not logged into GitHub is 404:

https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_47.html

-> https://github.com/orgs/flux-framework/teams/flux-code-of-conduct-contacts (404)

@trws is the two of us for emails sufficient? I remember you saying you were OK with that when we realized teams weren't an option. I'm definitely OK with having mine there. :+1:

### DIFF
--- a/spec_47.rst
+++ b/spec_47.rst
@@ -52,9 +52,9 @@ We pledge to act and interact in ways that contribute to an open, welcoming, div
 Need Help?
 ----------
 
-If any behavior makes you uncomfortable, or you believe it breaches the intent of this code of conduct, please contact a project maintainer:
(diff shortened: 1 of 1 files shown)


vsoch:
@garlick how should I be added there? I'm not a Flux admin but I should be a code of conduct contact (in case a woman is wanted for the contact).

garlick:
Oh sorry, well I guess my main point was just that github identities might be better than emails?

I can add you to the org admin page as a CoC contact and then this doc can just point there?
At least then we have one page to keep updated rather than two.

garlick (approved):
LGTM!  If you could please squash that commit into the first one we can get this in.

PR #493, opened 2026-02-11
rfc19: specify json encoding rules for FLUIDs

garlick:
Problem: it is not safe to encode FLUIDs as JSON numbers, yet we currently do that routinely in Flux.

Require a string encoding going forward.

Note: the proposed language allows any valid FLUID string encoding since they are unambiguous and the public API functions  `flux_jobid_encode(3)` and `flux_jobid_parse(3)` are available.  However, @dirkschubert-linaro requested that a single encoding be used for human readability.  Thoughts on whether that should be required?  It doesn't cost anything to narrow the requirement to one encoding AFAICS such as `f58plain`

References
- flux-framework/flux-core#6171

### DIFF
--- a/spec_19.rst
+++ b/spec_19.rst
@@ -200,3 +200,14 @@ SHALL use the following rules to decode the argument:
 
 An implementation decoding FLUID string representations SHALL
 ignore leading and trailing whitespace.
+
+JSON Encoding
+-------------
+
+When a FLUID value is added to a JSON object, it is RECOMMENDED to
+represent it as a JSON string, in F58 encoding as defined above.
+
+FLUDS SHALL NOT be represented as JSON numbers, since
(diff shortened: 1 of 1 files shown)


grondo (approved):

grondo:
I've thought about this for a bit and done some experiments and this seems good to me.

> RECOMMENDED to use the name `idstr` for the key

Is recommending the key name out of scope for this RFC?

Since most use cases currently of FLUIDs in JSON are to encode jobids, I wonder if a good long term plan would be to call the key simply `jobid`. For a time, externally used protocols would have both `id` (integer) and `jobid` (f58 encoded string). Eventually `id` could be dropped and payloads would just have `jobid` as if we planned it that way all along.

grondo:
Seems like we could leave it out. I just brought it up as a more general idea.

garlick:
OK, how's this look?

PR #494, opened 2026-02-19
remove reference to `flux mini` in RFC 36

wihobbs:
Problem: there's a reference left to `flux mini` in the RFCs.

Fix the script.


chu11 (approved):
doh! good catch

PR #514, opened 2026-04-15
rfc48: add github role mapping

garlick:
Problem: it's not clear how project roles map to github access control.

Propose a mapping.


grondo (approved):
Great! This seems very straightforward and reasonable to me. May want to get an ACK from @trws as well.

garlick:
Oh @trws verbally acked in our meeting so I'll set MWP

PR #518, opened 2026-05-14
rfc49: add treepool RFC

garlick:
A locality aware python scheduler under development in flux-core uses the `R.scheduling` key to carry topology information, but this format is undocumented.

Add a specification.


grondo (approved):
Hey, this is pretty great!  LGTM!

PR #519, opened 2026-05-23
RFC 15: describe external containment helper

grondo:
Problem: External projects that want or need to modify cgroup device containment would need to duplicate code from the IMP to construct and load BPF programs.

Specify a cgroup device containment helper that privileged processes can invoke to apply device containment to existing cgroups. The motivating use case is for flux-pam, which will want to manage device access for user slices on shared node systems.

Also, remove a stale `release` entry in the rfc project's `conf.py` so the RFC readthedocs site stops claiming `Flux 0.13.0`

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -33,10 +33,6 @@ copyright = '''Copyright 2014 Lawrence Livermore National Security, LLC and Flux
 SPDX-License-Identifier: LGPL-3.0'''
 author = 'This page is maintained by the Flux community.'
 
-# The full version, including alpha/beta/rc tags
-release = '0.13.0'
-
-
 # -- General configuration ---------------------------------------------------
 
 # Add any Sphinx extension module names here, as strings. They can be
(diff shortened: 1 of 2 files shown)


garlick (approved):
Works for me!

grondo:
While working on the implementation for the external helper, I realized that the current spec for the Device Containment options to the IMP was incorrect. I've added a commit here that corrects the section to describe what is currently in use. That could use a review.

grondo:
The intended use case for the external containment helper was to assist the prolog/housekeepingscripts in updating the device-allow BPF program on the user slice associated with a starting or ending job. However, the helper is actually unnecessary in this case: the user slice is managed by PID 1, which already has permission to load, modify, and remove BPF programs on that slice. The prolog/housekeeping scripts can therefore apply device properties at the same time as CPUsAllowed, MemsAllowed, etc., and the new device properties will take effect immediately.

Closing this PR as not needed.

PR #520, opened 2026-05-31
rfc15: update device containment section to match implementation

grondo:
As noted in #519, the specification of device options in the RFC 15 Device Containment ended up not matching the actual implementation.

Align the specification with the implementation.

Also, fix a errant `release = '0.13.0'` in `conf.py` which causes the RFC readthedocs page to always report version 0.13.0.

### DIFF
--- a/spec_15.rst
+++ b/spec_15.rst
@@ -271,37 +271,57 @@ Device Containment
 Two optional keys control device access for the job shell and its
 descendants.
 
-device-policy
-   A string specifying the device access policy. The only currently
-   defined value is "closed", which permits access to a baseline set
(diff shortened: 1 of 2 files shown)


spec_15.rst
```diff
@@ -271,59 +271,79 @@ Device Containment
 Two optional keys control device access for the job shell and its
 descendants.
```
garlick, line 272: Commit message: s/Containmnet/Containment/

garlick (approved):
Oof sorry I forget to get back and update that.  Looks good.

grondo:
Thanks! Fixed that typo and set MWP.

PR #521, opened 2026-06-12
rfc21: document exec eventlog events and add new `shell-exit` event

grondo:
In flux-framework/flux-core#7673 and flux-framework/flux-core#7660 it was observed that once the leader job shell exits, services it provides (MPIR proctable, standard I/O, the shell exception service) are permanently unavailable, but nothing in the job record marks the transition. Tools get a bare `ENOSYS` and cannot distinguish a transient error from permanent
loss of services, and diagnosing the hang required poking live flux module stats job-exec data that doesn't survive the job.

This PR first adds a new Execution Eventlog Events section to RFC 21 and documents the existing events posted by the job execution system. Then it proposes a new `shell-exit` event, posted when the execution system observes leader shell termination. This will allow tools and other eventlog users to know when leader shell services are no longer available for a job, and also may serve as the start of a shell-exit timeout applied by the job-exec module to jobs to avoid jobs that hang with a job shell that is stuck at exit.

### DIFF
--- a/spec_16.rst
+++ b/spec_16.rst
@@ -194,7 +194,8 @@ The *exec system* creates the job’s guest namespace and links it to
 ``job.<jobid>.guest``. Its initial contents are populated with
 
 ``exec.eventlog``
-   An eventlog for the use of *job shells*, TBD.
+   An eventlog containing events posted by the *exec system* and
+   *job shells*, described in :ref:`RFC 21 <spec_21_execution_eventlog>`.
 
 Once all *job shells* have exited and all outstanding writes to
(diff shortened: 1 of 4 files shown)


garlick:
I'm surprised we haven't done this already! 


However I wonder if this eventlog should have its own RFC instead of adding on to the job manager one?   It's related but we do have a clean separation between exec system and job manager.

grondo:
Sure. I had considered doing that but figured someone looking a job eventlog might naturally think to look at the "Job States and Events" RFC. I didn't see anywhere mentioned that this was the job manager RFC. Easy enough to add a new RFC.

PR #522, opened 2026-06-12
rfc50: add job execution eventlog specification

grondo:
The execution system maintains an eventlog at `exec.eventlog` in the job's guest KVS namespace, but the events posted there are not documented in any RFC. RFC 16 declares the eventlog's contents "TBD" and RFC 41 normatively requires its terminating "done" event with no specification to reference.

Add RFC 50/Job Execution Eventlog documenting the events posted by the execution system , note that these events do not drive job state transitions, and reserve the "shell." prefix for events posted by the job shell. Update dangling references to point readers to RFC 50. Finally, propose a new `shell-exit` event that the execution system posts when it records termination of the leader job shell.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ check: $(SCHEMA_DIRS) spelling
 	./indexcheck spec_*.rst
 
 $(SCHEMA_DIRS):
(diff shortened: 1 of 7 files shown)


garlick (approved):
LGTM!  Filling in a big gap in the docs!

PR #523, opened 2026-06-16
propose redacted R stored in KVS as `R_redacted`

grondo:
This PR updates relevant RFCs to require that an Rv1 object with any `scheduling` key be stored in the KVS as `R_redacted` co-located with any full `R`. This is to solve the issue of a large scheduling key being unnecessarily fetched from the KVS for consumers that only need the execution section. See flux-framework/flux-core#7615

A commit is also included that adds a missing requirement to the resource acquisition protocol that R be sent in the alloc response (de-facto requirement in the code now). The scheduling key is made OPTIONAL in this response in keeping with the spirit of this PR.

### DIFF
--- a/spec_16.rst
+++ b/spec_16.rst
@@ -124,8 +124,8 @@ Guests may access data in the primary KVS namespace only through instance
 services that allow selective guest access, by proxy or by staging copies
 to the guest namespace.
 
-Guest access for primary namespace contents ``R``, ``J``, ``jobspec``, and
-``eventlog`` is provided via a proxy service in the instance.
+Guest access for primary namespace contents ``R``, ``R_redacted``, ``J``,
+``jobspec``, and ``eventlog`` is provided via a proxy service in the instance.
 
 Event Log
 =========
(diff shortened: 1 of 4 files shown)


garlick:
Just brainstorming, but since not all schedulers need this, would it make sense to flip it so `R` (required) in the KVS is the redacted version and another key like `R_scheduling` (optional) is the non-redacted version?

Alternatively, and more in keeping with the original design, add a way for the `scheduling.writer` key to specify indirection to an alternate key containing only the scheduling object.  The scheduling frameworks (python scheduler class, libschedutil) could be put in charge of composing/decomposing R for for the scheduler.  Then the rest of the system could just use R with ride-along scheduling key as designed.

grondo:
As I commented in flux-framework/flux-core#7615, this won't work because it would break backwards compatibility: the scheduler and resource modules of subinstances fetch the `R` key from the parent. This change would break launching a previous version of Flux under the new version.

grondo:
> Alternatively, and more in keeping with the original design, add a way for the scheduling.writer key to specify indirection to an alternate key containing only the scheduling object. The scheduling frameworks (python scheduler class, libschedutil) could be put in charge of composing/decomposing R for for the scheduler. Then the rest of the system could just use R with ride-along scheduling key as designed.

I like this suggestion if we could figure out a path to avoid breaking inter-version compatibility. Maybe job-info could assemble the full R by default unless a new flag is used? Then RPCs from older versions of Flux would continue to work, and the job shell and other same-version job-info callers could add the flag. 

I wonder if it would be simpler to add a new optional `scheduling.location` or similar key that does the redirect. The `scheduling.writer` key then stays the same in the separate `scheduling` object. Either way would be fine with me.

PR #524, opened 2026-06-22
document `scheduling.location` in R to allow separate storage of `scheduling` key

grondo:
Based on @garlick's idea in #523, this alternate proposal specifies a new `scheduling.location` key in R which, when present, denotes a separate storage location for the R `scheduling` key in the KVS. Components reading R directly from the KVS will therefore read only the base R by default and will need to separately fetch any `scheduling` key if required.

A couple design decisions of note:
- A new `scheduling.location` key was chosen rather than extending the `scheduling.writer` as suggested in #523 as a separation of concerns approach. This allows `scheduling.writer` to appear in its original form in the separately stored `scheduling` object. The object itself is self-contained and the writer key can be read from there.
- RFC 27 is updated to explicitly require R as REQUIRED in a alloc success response. Since the job manager immediately deletes the R `scheduling` key anyway, the scheduler is allowed to drop the `scheduling` key in the response. Not sure if it would be useful to keep it when it only contains a `location` key.

This proposal as-is would break backwards compatibility with any component that assumes the full R will be returned when R is fetched from the KVS, job-info service, etc. This includes the `resource` and `sched-fluxion-resource` modules in subinstances, which make use of `job-info`. For a seamless transition, one plan would be to introduce two new flags in the `job-info` lookup protocol:
- `assemble` - if the key (e.g. R) is stored in parts, job-info returns the fully assembled result
- `no-assemble` - if the key is stored in parts, job-info does *not* assemble the result

Then, `assemble` is the default in the initial version of the implementation. This keeps backwards compatibility for subinstances of differing Flux versions. Components that do not need the scheduling key in R (like same-version job shells), add the `no-assemble` flag to the `job-info.lookup` RPC. After awhile, we switch the default and all components keep working as is. The `no-assemble` flag could eventually be retired.

Components that fetch R directly from the KVS will always get the unassembled version. However, these are all same-version components and for the most part (except for resource module restart) want only the unassembled R so I don't think there's an compatibility issue there.

### DIFF
--- a/spec_16.rst
+++ b/spec_16.rst
@@ -179,6 +179,21 @@ The scheduler allocates resources by writing a resource set
 as described in :doc:`RFC 20 <spec_20>`
 to ``job.<jobid>.R`` and answering the allocation request.
 
+``job.<jobid>.R`` is the canonical resource set for the job. Its
+:data:`scheduling` object (RFC 20), if present, MAY be stored out-of-line
+in ``job.<jobid>.scheduling`` and referenced by a :data:`scheduling.location`
+key rather than stored directly within *R*. Consumers that require the
+complete *R* (e.g. the scheduler during hello/replay, or a subinstance
+resource module) SHALL obtain it by reading both ``job.<jobid>.R`` and
+``job.<jobid>.scheduling`` and merging them, or by requesting the complete
+*R* from a service that performs this reassembly.
+
+``job.<jobid>.scheduling`` is an OPTIONAL write-once key that holds the
+complete :data:`scheduling` object referenced by :data:`scheduling.location`
+in ``job.<jobid>.R``. It SHALL be written before ``job.<jobid>.R`` is
+committed so that the referenced data is available as soon as the reference
+is visible.
+
 The scheduler frees resources by answering the free request,
 leaving ``R`` in place for job provenance. During a restart, the
 *job manager* uses the eventlog to determine whether ``R`` is currently
--- a/spec_20.rst
+++ b/spec_20.rst
@@ -233,12 +233,40 @@ R Format
     components SHALL be interpreted by the named scheduler.  If not
     provided, a value of ``fluxion`` SHALL be assumed.
 
+  .. data:: location
+
+    (*string*, OPTIONAL) A KVS key path at which the :data:`scheduling`
+    object is stored out-of-line. When :data:`scheduling` is stored
+    out-of-line, :data:`location` SHALL be present and the referenced KVS
+    key SHALL contain a complete :data:`scheduling` object. The referenced
+    key is written once and never rewritten. Per-job the conventional key
+    is ``job.<id>.scheduling``; for the instance resource inventory the
+    conventional key is ``resource.scheduling``.
+
+    A consumer that requires the complete :data:`scheduling` object SHALL
+    obtain it either by fetching the key named by :data:`location` itself,
+    or by requesting a complete *R* from a service that performs the
+    reassembly on its behalf. A service that provides *R* (such as a job
+    information service) MAY offer this reassembly on demand, controlled by
+    a request option. The default behavior of such a service is
+    implementation-defined and MAY change over time, so a requester that
+    requires a specific behavior SHOULD request it explicitly.
+
   Remaining scheduler-specific keys SHOULD be ignored by other Flux components.
 
-  When used, :data:`scheduling` SHALL ride along on the
-  resource acquisition protocol (RFC 28) and resource allocation protocol
-  (RFC 27) so that it may be included in static configuration, allocated to
-  jobs, and passed down a Flux instance hierarchy.
+  When used, :data:`scheduling` SHALL ride along on the resource acquisition
+  protocol (RFC 28) and resource allocation protocol (RFC 27) so that it may
+  be included in static configuration, allocated to jobs, and passed down a
+  Flux instance hierarchy. When :data:`scheduling` is stored out-of-line via
+  :data:`location`, it rides these protocols by virtue of being retrievable
+  from the KVS and reassembled by the persistence layer.
+
+  A component persisting *R* to the KVS MAY store the :data:`scheduling`
+  object out-of-line via :data:`location` to keep the stored *R* compact while
+  still providing access to the complete scheduler data via the referenced
+  key. Because :data:`scheduling` is independent of expiration, only the
+  stored *R* is rewritten when expiration is updated per RFC 21; the
+  out-of-line :data:`scheduling` key is never rewritten.
 
   Linkage to specific resources in :data:`R_lite` SHOULD use hostnames rather
   than execution targets since the scheduler-agnostic re-ranking of *R* that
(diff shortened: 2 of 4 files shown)


spec_27.rst
```diff
@@ -326,8 +326,24 @@ If resources can be allocated, the scheduler SHALL ensure that *R* has
...
-the OPTIONAL key:
+In addition to the above REQUIRED keys, the SUCCESS response includes the
+following REQUIRED key:
+
+R
+  (object) the allocated resource set (RFC 20). The job manager consumes this
+  resource set directly and need not read it back from the KVS. Since the job
+  manager requires only the execution portion, the scheduler MAY omit the
+  OPTIONAL :data:`scheduling` key from *R* in this response; the job manager
+  SHALL NOT rely on :data:`scheduling` being present here. The scheduler SHALL
```
garlick, line 337: Suggestion:  Just say the scheduler SHALL NOT include the scheduling key in this response rather than make baggage optional that the job manager is then ordered not to use.  Flux-core owns the code that composes this response for fluxion (libschedutil) so we can make this change in one place and it won't break interoperability.

spec_27.rst
```diff
@@ -326,8 +326,24 @@ If resources can be allocated, the scheduler SHALL ensure that *R* has
...
+  persist *R* to ``job.<jobid>.R`` (RFC 16) before responding; if the
+  scheduler stores :data:`scheduling` out-of-line (RFC 20), it SHALL write the
+  complete :data:`scheduling` object to ``job.<jobid>.scheduling`` and
+  reference it with a :data:`scheduling.location` key in ``job.<jobid>.R``.
```
garlick, line 341: Suggestion: just say that `location` is for now always set to `scheduling` (bare key, in same directory).  That simplifies descriptions and lets code fetch both keys in parallel, ignoring `scheduling` if it's not found or not referenced, retaining the self-describing nature of "location" and leaving the door open for other locations if need be.

spec_16.rst
```diff
@@ -179,6 +179,21 @@ The scheduler allocates resources by writing a resource set
...
+resource module) SHALL obtain it by reading both ``job.<jobid>.R`` and
+``job.<jobid>.scheduling`` and merging them, or by requesting the complete
+*R* from a service that performs this reassembly.
+
+``job.<jobid>.scheduling`` is an OPTIONAL write-once key that holds the
+complete :data:`scheduling` object referenced by :data:`scheduling.location`
+in ``job.<jobid>.R``. It SHALL be written before ``job.<jobid>.R`` is
+committed so that the referenced data is available as soon as the reference
+is visible.
+
```
garlick, line 196: Suggestion: s/written before/committed before or at the same time as/

spec_20.rst
```diff
@@ -233,12 +233,40 @@ R Format
...
 
+  .. data:: location
+
+    (*string*, OPTIONAL) A KVS key path at which the :data:`scheduling`
+    object is stored out-of-line. When :data:`scheduling` is stored
+    out-of-line, :data:`location` SHALL be present and the referenced KVS
+    key SHALL contain a complete :data:`scheduling` object. The referenced
+    key is written once and never rewritten. Per-job the conventional key
+    is ``job.<id>.scheduling``; for the instance resource inventory the
+    conventional key is ``resource.scheduling``.
```
garlick, line 244: If suggestion to hardwire a value of `scheduling` is accepted, the two conventional key examples can be dropped.

"written once" is covered below (redundant here?)

spec_20.rst
```diff
@@ -233,12 +233,40 @@ R Format
...
+
+    A consumer that requires the complete :data:`scheduling` object SHALL
+    obtain it either by fetching the key named by :data:`location` itself,
+    or by requesting a complete *R* from a service that performs the
+    reassembly on its behalf. A service that provides *R* (such as a job
+    information service) MAY offer this reassembly on demand, controlled by
+    a request option. The default behavior of such a service is
+    implementation-defined and MAY change over time, so a requester that
+    requires a specific behavior SHOULD request it explicitly.
```
garlick, line 253: Suggestion: this interrupts the flow of the object definition a bit.  Move to note box or combine with other expository paragraphs below?

spec_20.rst
```diff
@@ -233,12 +233,40 @@ R Format
...
-  When used, :data:`scheduling` SHALL ride along on the
-  resource acquisition protocol (RFC 28) and resource allocation protocol
-  (RFC 27) so that it may be included in static configuration, allocated to
-  jobs, and passed down a Flux instance hierarchy.
+  When used, :data:`scheduling` SHALL ride along on the resource acquisition
+  protocol (RFC 28) and resource allocation protocol (RFC 27) so that it may
+  be included in static configuration, allocated to jobs, and passed down a
+  Flux instance hierarchy. When :data:`scheduling` is stored out-of-line via
+  :data:`location`, it rides these protocols by virtue of being retrievable
+  from the KVS and reassembled by the persistence layer.
```
garlick, line 262: This paragraph is now confusing as to whether scheduling "rides along" or is stored separately, and in fact the RFC 27 allocation protocol describes optionally storing scheduling separately in the KVS and maybe being included in the alloc RPC response, while the RFC 28 acquisition protocol presumably must continue to provide the fully composed R to the scheduler.

Since those protocols are both defined elsewhere, maybe this paragraph could be dropped?

garlick (approved):
I made a few suggestions but I don't feel terribly strongly about them.

I guess my main suggestion is that we make `location` a key name in the same directory as `R` rather than a path, and for now to require it to be set to `scheduling`.  This would tighten up the spec, require fewer words to describe, and give consumers the option of fetching the keys in parallel rather than have to fetch R first to determine the location value.

spec_27.rst
grondo: This would be a larger change in behavior orthogonal to adding a `scheduling.location` key, and so perhaps would be appropriate for a separate PR and corresponding flux-core PR to make the change. However, if you think it should be folded into this PR, I'm fine with that, but that's why I held back from proposing SHALL NOT here.

spec_27.rst
grondo: > location is for now always set to `scheduling` (bare key, in same directory)

The KVS doesn't have a concept of relative paths, so it would be difficult in the future to tell if `"scheduling"` indicates the `"scheduling"` key in the top-level KVS directory or the same directory as R. Also if for some reason an R object is copied or moved to a different directory, then the `location` becomes stale. While these misgivings are highly unlikely, it does make me question if `scheduling.location` is the right solution if all we want to do is specify that the scheduling key SHALL be separately stored in a `scheduling` key in the same directory as R.

spec_20.rst
grondo: It rides along the protocol by virtue of being referenced by location. Agreed that this could be dropped though.

spec_20.rst
grondo: I think possibly hardwire of `location` to `scheduling`  may require a more nuanced rewrite of these RFC changes. I was just trying to get the discussion started. I'll close this one and the next person with the energy can take a stab at it.

PR #525, opened 2026-06-24
rfc43: add private mode specification for `job-list` service

grondo:
This PR adds a private mode spec to the job-list RFC. Private mode specifies that guests can only get information about their own jobs and cannot glean information about other jobs e.g. by querying `job-list.list-id` and differentiating `ENOENT` vs `EPERM`.


garlick (approved):
LGTM

grondo:
Thanks. Will set MWP.

PR #528, opened 2026-07-08
rfc1: Assisted-by example should use periods

garlick:
Problem: the Assisted-by example uses opus-4-8 but it's probably more natural to use periods.

Change the example.

### DIFF
--- a/spec_1.rst
+++ b/spec_1.rst
@@ -131,7 +131,7 @@ Patch Requirements
 
 -  An AI coding assistant SHOULD NOT be credited as an author or co-author of a patch (e.g. via a ``Co-authored-by`` trailer), because authorship implies a responsibility that only a human Contributor can hold. This does not apply to human co-authors, who SHOULD continue to be credited.
 
--  Where an AI coding assistant was used to prepare a patch, the commit message body SHOULD record this with an ``Assisted-by:`` trailer of the form ``Assisted-by: NAME:MODEL_VERSION``, for example ``Assisted-by: Claude:claude-opus-4-8``.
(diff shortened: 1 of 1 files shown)


chu11 (approved):
LGTM!

grondo:
One point that might need clarification: can MODEL_VERSION contain whitespace?

chu11:
> One point that might need clarification: can MODEL_VERSION contain whitespace?

On that note, do we want to be specific about format or perhaps keep it semi broadly open?

In https://github.com/flux-framework/flux-core/pull/7715  @grondo thought it should be `Claude:claude-opus-4.8`, but when I asked Claude to put in a trailer, it chose `Claude:Opus 4.8`.   Version can be somewhat ambiguous or broad.

wihobbs:
One thing I just noticed is the [LKML "standard"](https://docs.kernel.org/process/coding-assistants.html) isn't cited anywhere on the page even though that's what we're using. Perhaps it should go in "further reading"?

Saying "follow the LKML guidelines" is, IMHO, as unambiguous as you can get for this.

garlick:
I'm not sure we need to cite the kernel here, or get too prescriptive.  The kernel's guidelines are also pretty open ended about things like periods and spaces.

grondo:
Ok, sounds like "Claude: Opus 4.8" is just fine. The example implies an attempt to avoid whitespace (e.g. to perhaps allow later parsing when there are multiple items in the `Assisted-by:` line), but sounds like that isn't necessary. That was my only point of confusion.

garlick:
works for me.  Hopefully I'm not being shortsighted about parsing, but unless were to specify a full on schema like SPDX, it seems like it doesn't matter much.

PR #529, opened 2026-07-09
rfc27: add `sched.selection_type` key to example annotations

wihobbs:
I have a work-in-progress prototype of this behavior on my fork of flux-sched (see the [`backfill_annotate`](https://github.com/wihobbs/flux-sched/tree/backfill-annotate) branch.)

This proposal is a small change to the resource allocation protocol to add a "start method" to the list of example scheduler annotations for a successful job. This would capture data about the scheduling 'method' that was invoked when the job was scheduled. 

I don't mind if this waits until a scheduler actually implements it to go in, but wanted to put it up.

### DIFF
--- a/spec_27.rst
+++ b/spec_27.rst
@@ -404,6 +404,10 @@ sched.resource_summary
 sched.queue
   (string) human readable identification of job queue
 
+sched.selection_type
+  (string) human readable identification of the scheduler's selected method 
+  invoked to schedule the job (i.e. immediate, reserved, backfilled)
+
 user
   (dictionary) dictionary object containing user specific annotations
 
--- a/spell.en.pws
+++ b/spell.en.pws
@@ -531,3 +531,4 @@ Nvidia
 proctable
 Claude
 validators
+backfilled


grondo (commented):
Suggestion: `sched.start_method` could be confusing because it is the job execution system that handles the `start` request (and subsequent `start` event, RFC 32). Since this RFC describes the alloc protocol, would `alloc_method` be more clear?

Also, commit message needs Problem: statement.

trws:
A quick glance at the sched PR that end looks good, would rather get the interface right here before we merge though since we have a bit before next reasonable release window for this.

trws:
I'm leaning toward something like your suggestion @grondo, maybe `alloc_type` or `sched_selection_type`? That and make the strings the same here and in the PR and I think we're probably good.

wihobbs:
Oh, of the options so far I like `sched.selection_type` the most. @cmoussa1 @garlick your thoughts?

cmoussa1:
I think `sched.selection_type` makes sense to me too!

wihobbs:
Ok, that seems to have consensus. Can someone push the approve button? Then I'll go update my PR to Fluxion.

cmoussa1 (approved):
LGTM!

wihobbs:
@grondo Okay with adding MWP to this?

grondo (approved):
LGTM!

PR #531, opened 2026-07-11
rfc42: guarantee EOF delivery on attach

garlick:
Problem: The attach RPC reports the flags the subprocess was originally started with, which tell the client which output streams to expect, but if a stream reaches EOF prior to the attach, the specification does not require an EOF response to be synthesized.

State in the attach response stream requirements that the server sends exactly one EOF output response for each forwarded stream, whether the stream ends before or after the attach and on every attach

Assisted-by: Claude:opus-4.8


grondo (approved):
LGTM!

PR #532, opened 2026-07-13
rfc42: clarify background subproc input handling

garlick:
Problem: the specification does not state how input (standard input and writable auxiliary channels) should be handled in a background subprocess.

There is no strong use case for input processing in background subprocesses, but since the spec does not address this, implementations would presumably need to add support for resuming an input stream on reattach.

Close off this area by stating that a background subprocess's standard input and writable auxiliary channels are at EOF.  As a consequence, the write-credit and stdio-fallthrough flags, which request input handling, are errors when combined with a non-streaming request.

Assisted-by: Claude:opus-4.8


grondo (commented):
👍

grondo (approved):

garlick:
Thanks!  Setting MWP.

PR #533, opened 2026-07-14
rfc42: update attach request/response parameters

garlick:
Problem: the `attach` request and response do not contain enough information to handle channel output properly.

- add `flags` to the `attach` request, so (like with `exec`) the client can indicate which callbacks are registered, so that the server doesn't forward channels that have nowhere to go
- add the command object to the initial `attach` response, so the client can add plumbing for all output channels and configure buffering
- drop `flags` from the initial `attach` response, since those were unused


grondo (approved):
Nice!

PR #534, opened 2026-07-14
rfc: document or-slots and xor-slots in jobspec specifications

zekemorton:
Add documentation for or-slots and xor-slots to RFC 14 and RFC 25.
  
## RFC 14 (Canonical Jobspec)
- Extended `slot` definition to describe or-slot behavior (sibling slots as alternatives)
- Added new `xor_slot` reserved resource type specification
- Added comparison table and three use cases with examples
- Updated JSON schema to support `xor_slot` type
  
## RFC 25 (Jobspec V1)
- Added note clarifying V1 does not support or-slots or xor-slots
- References RFC 14 for users needing alternative slot configurations
  
Or-slots use dynamic programming for optimal slot selection during traversal.
Xor-slots expand into distinct candidate jobspecs before traversal with first-match selection.
Both require the flexible traverser implementation.

### DIFF
--- a/spec_14.rst
+++ b/spec_14.rst
@@ -244,6 +244,72 @@ Reserved Resource Types
    to the program described in the jobspec, unless otherwise specified
    in the ``exclusive`` field of the associated resource.
 
+   When multiple ``slot`` resources appear as siblings at the same hierarchical
+   level (referred to informally as "or-slots"), they represent alternative slot
+   configurations. The flexible traverser MAY select any one of the sibling slots
+   to satisfy the request. This form is useful when alternatives share the same
+   task slot label and live at the same jobspec level. The traverser uses a
+   dynamic-programming approach to choose the slot configuration that yields the
+   largest total number of satisfiable slot matches. Once a slot shape has been
+   selected, concrete resources for each matched slot are chosen according to
+   the active match policy.
+
+**xor_slot**
+   A resource type of ``type: xor_slot`` SHALL indicate an alternative
+   resource grouping that is expanded into distinct jobspec branches before
+   traversal. Exactly one of the xor-slot alternatives will be allocated or
+   reserved. Each ``xor_slot`` is converted into a normal ``slot`` candidate
+   internally, and the flexible traverser tries the expanded branches sequentially
+   until a match is found (first-match strategy).
+
+   An ``xor_slot`` SHALL follow the same structural requirements as ``slot``,
+   including the requirement for a ``label`` key and at least one edge specified
+   using ``with:``.
+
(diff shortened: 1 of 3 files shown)


zekemorton:
@milroy do you mind giving this a first pass when you get a chance?

spec_14.rst
```diff
@@ -244,6 +244,67 @@ Reserved Resource Types
...
+   to satisfy the request. This form is useful when alternatives share the same
+   task slot label and live at the same jobspec level. The traverser uses a
+   dynamic-programming approach to choose the slot configuration that yields the
+   largest total number of satisfiable slot matches. Once a slot shape has been
+   selected, concrete resources for each matched slot are chosen according to
+   the active match policy.
+
+**xor_slot**
+   A resource type of ``type: xor_slot`` SHALL indicate an exclusive alternative
+   resource grouping that is expanded into distinct jobspec branches before
```
milroy: I think you want to disambiguate the "exclusive" adjective here as we'll soon support non-exclusive resources in a slot. This statement should make it clear that exactly one of the XOR slots will be allocated or reserved.

spec_14.rst
```diff
@@ -244,6 +244,67 @@ Reserved Resource Types
...
+   traversal. Each ``xor_slot`` is converted into a normal ``slot`` candidate
+   internally, and the flexible traverser tries the expanded branches sequentially
+   until a match is found (first-match strategy).
```
milroy, line 263: You may want to emphasize that the XOR slot order effectively encodes preference.

spec_14.rst
```diff
@@ -244,6 +244,67 @@ Reserved Resource Types
...
+
+   This type is particularly useful when alternatives carry different subtree
+   prefixes (e.g., different hierarchical paths from the root) or when exclusive
+   same-level alternatives are needed. Unlike sibling ``slot`` resources which
+   are evaluated during traversal, ``xor_slot`` resources are expanded into
+   separate candidate jobspecs before matching begins.
+
+   Nested ``xor_slot`` resources are supported. When processing nested xor-slots,
+   the expansion produces all combinations of the xor branches, and each
+   combination is tried as a separate candidate jobspec.
```
milroy: Please add a phrase mentioning the need to limit the number of candidates, and the configuration option that sets the limit.

milroy (commented):
Looks good so far! Please add a bit more description where I indicated. After that this is good to merge as far as I'm concerned!

milroy (approved):
This PR looks good to me. Please change the commit messages to conform to the "Problem: ... Add" structure before merging. It would also be good to get a review from @garlick or @grondo.

PR #536, opened 2026-07-20
rfc33: clarify queue access policy enforcement semantics

grondo:
Problem: RFC 33 does not specify behavior of queue access policy under multiple validators, but this could cause confusion when multiple validators are composed (e.g. flux-core internal vs an external entity such as flux-accounting), leading to unexpected job rejection or acceptance.

Clarify how multiple access validators compose (e.g. the built-in tables alongside flux-accounting's database-driven access control), and what a validator should do with a queue it does not manage.

### DIFF
--- a/spec_33.rst
+++ b/spec_33.rst
@@ -177,11 +177,27 @@ policy.access.allow-group
   (list of strings) Specify a list of UNIX group names that are to be granted
   access.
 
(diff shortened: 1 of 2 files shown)


garlick (approved):
LGTM!

grondo:
Thanks! Setting MWP.

PR #537, opened 2026-07-21
rfc33: add queue aliases

grondo:
Opening this RFC 33 proposal to get some discussion going.

Problem: Creating an alternate job submission path with different limits, access, or priority (the "expedite"/"exempt" patterns of flux-framework/flux-core#4306) currently requires configuring a fully independent queue, duplicating resource configuration and creating a separate scheduling queue for no scheduling reason — something RFC 33's design criteria explicitly warn against.

This amendment adds *queue aliases*: an optional `queues.NAME.parent` key declaring a queue to be an alternate name for an existing queue. An alias inherits its parent's resource subset and policy, may override inherited policy at ingest, and shares its parent's scheduling order. The job's queue attribute is never rewritten.

Notes for discussion:

- **Key naming**: `parent` is the proposal; `alias-of` and `inherits-from` were also considered. "Parent" is admittedly overloaded with the parent *instance* concept, but the key only appears within a `queues.NAME.` table where that reading isn't plausible, and prose always qualifies as "parent queue."
- **Priority**: initially handled by flux-accounting's per-queue priority factor (no accounting changes needed). A core-side follow-up could add simple built-in policies, e.g. per-queue default urgency (`urgency = 16` makes an "expedite" alias work out of the box; see also flux-framework/flux-core#5795). Deliberately not specified here.
- **Job listing**: requesting jobs by parent name SHOULD include alias jobs (single scheduling order); requesting by alias name is exact. (More thought needed on implementation here)
- **Administrative state**: submission enable/disable is independent per alias. Started/stopped state follows the parent, and drain/idle are not supported on an alias (even once flux-framework/flux-core#4821 is fixed), since alias jobs are scheduled as part of the parent queue and an alias has no resources of its own to quiesce. Conversely, drain/idle on a queue must count jobs submitted to its aliases.
- **v1 restriction**: an alias's parent must not itself be an alias.

Fluxion support will require one small change to detect aliases and direct jobs to the parent queue internally.

### DIFF
--- a/spec_33.rst
+++ b/spec_33.rst
@@ -219,6 +219,45 @@ queues.NAME.policy
   top-level ``policy`` table  and a queue-specific ``policy`` table, the
   queue-specific value takes precedence for jobs submitted to that queue.
 
+queues.NAME.parent
+  (string) Declare this queue to be a virtual queue backed by another
+  configured queue, as described in `Virtual Queues`_ below.
+
+Virtual Queues
+--------------
+
+A queue configured with a ``parent`` key is a *virtual queue*: an
+alternate name for submitting jobs to the parent queue's resources
+with different policy applied at ingest.
+
+Configuration SHALL be rejected if ``parent`` does not name a
+configured queue, or if the named queue is itself a virtual queue.
+
+A virtual queue SHALL inherit the resource subset and effective policy
+of its parent queue.  It MAY override inherited policy values in its
+own ``policy`` table, but MUST NOT be configured with ``requires`` or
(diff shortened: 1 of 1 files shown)


grondo:
Actually, @garlick had suggested the name "virtual queues" or vqueues instead of queue aliases, and after reading this proposal back, I think virtual is a better term than alias. So I'm going to rewrite the addition with that terminology instead. Otherwise, the behavior called out in the RFC will be the same.

grondo:
I'll open a new PR so we don't have the old "alias" name in the merged branch.

PR #539, opened 2026-07-22
rfc42: add output key to wait response

garlick:
Problem: the subprocess wait RPC cannot return any output, but this may be useful in a future design where wait reaps background flux-shell processes that may have emitted important failure telemetry on stderr.

Add an optional 'output' key to the wait response.  The key is purposefully specified to be incomplete and optional so that server retention (and thus the wait response message size) can be capped to scalable levels.  This framing also permits a server implementation such as sdexec to remain conformant while allowing an I/O monitoring lapse across a server restart.

Assisted-by: Claude:opus-4.8

### DIFF
--- a/spec_42.rst
+++ b/spec_42.rst
@@ -501,12 +501,29 @@ with the waitable flag set.
 
 .. object:: wait response
 
-  A successful response SHALL consist of a JSON object with the following key:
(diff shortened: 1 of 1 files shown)


grondo (approved):
LGTM!

garlick:
Thanks!

PR #540, opened 2026-07-24
rfc32: support reload of job-exec module with running jobs

garlick:
Problem: the protocol between the job manager and execution system includes (early/prototype) support for instance restart with running jobs, but precludes reattaching to jobs across a `job-exec` module reload.  This contradicts a stated goal of the RFC and precludes a useful test mode.

Relax the spec to allow this.

Also fix a couple of omissions to match the spec to current reality.

Note: to make each diff more easily reviewable, I reflowed paragraphs as a final commit.  If that's not acceptable I can split that last commit up and squash it with the others.

### DIFF
--- a/spec_32.rst
+++ b/spec_32.rst
@@ -148,8 +148,14 @@ jobspec
   (object) *jobspec* object (RFC 14)
 
 reattach
-  (boolean) Set to True if broker has been restarted and job should still
-  be running.
+  (boolean) Set to True if the job's shells were already launched by a
+  previous execution service instance and should be recovered rather than
+  launched anew.  Whether the shells actually survive a given restart is a
+  property of the execution mechanism and outside the scope of this protocol;
+  ``reattach`` asserts only that they were launched by a previous instance
+  and, if still present, are to be recovered.  The execution service SHALL
+  respond with ``reattached`` in place of ``start`` once recovery completes,
+  or raise a job exception as for any other unrecoverable start failure.
 
 
 Example:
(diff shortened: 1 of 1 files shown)


spec_32.rst
```diff
@@ -150,6 +151,13 @@ userid
 jobspec
   (object) *jobspec* object (RFC 14)
 
+R
+  (object) *R* resource set object (RFC 20).  Together with ``id`` and
+  ``jobspec`` this makes the ``start`` request self-sufficient: a newly
+  registered execution service reconstructs what it needs to reattach from
+  the request payload alone, with no state handed off from a previous
+  instance and no capability advertised in the ``exec-hello`` request.
```
grondo: This description implies that the _R_ object is only required for `reattach`, but it is also required for a normal `start` request without `reattach`, because this is how the job execution system determines the ranks on which a job shell needs to be started.

spec_32.rst
```diff
@@ -121,12 +121,15 @@ Example:
...
-the new one to override it.
-
-The response payload SHALL be empty on success.  The job manager SHALL issue
-a failure response if any jobs have an outstanding ``start`` request to an
-existing execution service.  The execution service SHALL treat a failure
+If an execution service is already registered, the job manager SHALL allow
+the new registration to replace it, whether it is the same service reloaded
+or a different one overriding it.  A job that had an outstanding ``start``
+request to the previous execution service instance had its shells launched by
+that instance; after responding, the job manager SHALL re-issue such a job's
```
grondo: Suggestion: `; after reponsding, the job manager SHALL` -> `; after responding to ``exec-hello``, the job manager SHALL`

grondo (approved):
LGTM! A couple suggestions inline.

garlick:
Thanks!  I fixed those things and will set MWP.

PR #541, opened 2026-07-24
rfc50: add recoverable event

garlick:
Problem: RFC 32 lets a reloaded execution service recover a job's already-running shells "if still present," but nothing in the execution eventlog records whether a job ever reached a state from which its shells can be recovered, so a reattaching service cannot deterministically decide between adopting the job and raising an exception.

Add an optional "recoverable" event to the RFC 50 execution eventlog, posted by the execution service when a job first becomes recoverable. The event is deliberately mechanism-agnostic: whether and when a job becomes recoverable is a property of the execution mechanism, so a service that cannot recover running shells never posts it.

Assisted-by: Claude:opus-4.8

### DIFF
--- a/spec_50.rst
+++ b/spec_50.rst
@@ -30,6 +30,7 @@ Related Standards
 - :doc:`spec_18`
 - :doc:`spec_21`
 - :doc:`spec_22`
+- :doc:`spec_32`
 - :doc:`spec_41`
 
 Background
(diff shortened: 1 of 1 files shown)


grondo (approved):
:+1:

garlick:
Thanks!

PR #542, opened 2026-07-25
rfc33: document queue status response and add `blocked` state for vqueues

grondo:
Problem: When a virtual queue is started but its parent queue is stopped, the queue is not scheduling: its jobs wait until the parent starts. Reporting this state truthfully with the existing payload forces a bad choice: report `start: true` and the queue looks like it should be scheduling when it isn't (an admin debugging "why won't my job run" is misled), or report `start: false` and it looks like someone stopped the queue when no one did - hiding that it is armed to begin scheduling the moment its parent starts, which an admin needs to know before starting the parent.

This PR first documents the existing queue status protocol, then introduces a `blocked` flag to indicate that a queue has `start: false` due to an external condition, i.e. if a vqueue has `start: false` and `blocked: true` then the vqueue is stopped by an external condition, o/w if `blocked: false` then the vqueue is stopped by an admin and needs a `flux queue start`. This also resolves a similar ambiguity when a normal queue is stopped due to the scheduler being unloaded.

### DIFF
--- a/spec_33.rst
+++ b/spec_33.rst
@@ -316,13 +316,44 @@ stop_reason
   (string) the reason scheduling is stopped.  This key MAY be present
   when ``start`` is false.
 
+blocked
+  (boolean) the queue is started but prevented from scheduling by an
+  external condition.  This key SHALL be present and true when such a
+  condition exists, and SHALL be omitted otherwise.
+
+parent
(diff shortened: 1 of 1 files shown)


garlick (approved):
That makes sense.

PR #543, opened 2026-07-29
rfc33: report blocked reason as a keyword

grondo:
Problem: The RFC 33 queue status "blocked" boolean is insufficient for consumers to determine the exact reason a started queue is prevented from scheduling. A consumer cannot tell the scheduler being offline from a virtual queue's parent being stopped without parsing the human-readable "stop_reason" string.

Redefine the "blocked" field as a string reason naming the condition ("scheduler" or "parent") rather than a boolean. Since consumers need only check for the key's presence, the recoverability of a queue's own started state (start true OR blocked present) is unchanged.

Additionally, relax the requirement that a blocked queue include a "stop_reason" from SHALL to SHOULD, since the key is defined as OPTIONAL when "start" is false.

Assisted-by: Claude:Opus-4.8

### DIFF
--- a/spec_33.rst
+++ b/spec_33.rst
@@ -317,9 +317,19 @@ stop_reason
   when ``start`` is false.
 
 blocked
-  (boolean) the queue is started but prevented from scheduling by an
-  external condition.  This key SHALL be present and true when such a
-  condition exists, and SHALL be omitted otherwise.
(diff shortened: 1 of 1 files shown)


garlick (approved):
👍

grondo:
Thanks! Setting MWP.

PR #544, opened 2026-08-01
rfc33: define queue membership by hostlist

grondo:
Problem: RFC 33 queue membership is defined indirectly. An admin assigns a resource property to nodes, then binds a queue to it with queues.NAME.requires. In practice nearly every configuration uses one property per queue named after the queue, so the indirection forces tools to re-implement the property-to-queue join and offers no way to validate membership.

Add an optional queues.NAME.hosts key that declares queue membership directly as an RFC 29 Hostlist (or the literal string "all"). In this case, the queue property is then implicitly assigned the queue name, and jobs have the proper constraint automatically assigned. Properties that match queue names are then reserved. Document in the spec that hosts is the preferred method for defining queues, and that `requires` may be deprecated in the future.

Assisted-by: Claude:Opus-4.8


grondo:
Hold off on reviewing this. It might be nice to make automatically assigned queue properties marked in some way, e.g. with a `+` prefix. Then the Rv1 spec could specify that properties with a leading `+` are "locally or automatically assigned"  can be discarded when creating a new R for an allocation. This would make it so that queue-only properties for new queues assigned with `hosts=` do not show up in jobs' `flux resource list` output, a minor annoyance now.

grondo:
Ok, now based on top of #545. A leading `+` is added to automatic queue properties. This indicates a property has been auto-assigned by Flux and need not be copied into R objects generated for allocations.

grondo:
Rebased after #545 was merged

grondo:
Thanks!

PR #545, opened 2026-08-01
rfc20: reserve leading + for instance-local properties

grondo:
Problem: There is no way to differentiate properties in a resource set that were applied automatically by Flux apart from a property configured manually, so tooling cannot know which properties describe the instance versus the resources. This leads to allocated R objects inheriting properties that may have no meaning in a subinstance.

Reserve a leading + in a property name to mark a property that Flux added automatically rather than one configured manually.  The literal property still applies to its execution target ranks, but because it describes the containing instance it is intended for internal use and MAY be omitted from a resource set allocated to a job.

Assisted-by: Claude:Opus-4.8

### DIFF
--- a/spec_20.rst
+++ b/spec_20.rst
@@ -194,6 +194,15 @@ R Format
     result in both ``amd-mi50@gpu`` and ``amd-mi50`` being valid properties
     for resources in the instance.
 
+    A leading ``+`` character marks an *instance-local* property: one that
+    describes the containing instance rather than an attribute of the
+    hardware, for example a property derived from queue membership. The
+    literal property SHALL apply to the defined execution target ranks as
+    usual. Because instance-local properties describe the instance and not
+    the resources themselves, a scheduler SHALL omit them from the *R* it
+    generates to allocate resources to a job. Other components that
+    generate an *R* subset MAY likewise omit them.
+
   .. data:: starttime
 
     (*number*, OPTIONAL) The start time at which the resource set is valid.


grondo:
Tweaked slightly to allow properties with leading + to indication "instance local" more generally, instead of just auto-added.

spec_20.rst
```diff
@@ -194,6 +194,14 @@ R Format
     result in both ``amd-mi50@gpu`` and ``amd-mi50`` being valid properties
     for resources in the instance.
 
+    A leading ``+`` character marks an *instance-local* property: one that
+    describes the containing instance rather than an attribute of the
+    hardware, for example a property derived from queue membership. The
+    literal property SHALL apply to the defined execution target ranks as
+    usual. Because it describes the instance and not the resources
+    themselves, an instance-local property MAY be omitted from a resource
+    set generated for a job or a child instance.
```
garlick: Would these properties be stripped from allocated _R_ sets (e.g. what gets stored in the job KVS dir) by the scheduler?  Or just stripped by the sub-instance resource module when it builds the inventory?

If the scheduler should strip these, the change currently wouldn't require a fluxion change since R is stored via libschedutil.

grondo: I was thinking the scheduler would strip them, but intentionally left that unspecified in the spec.

grondo: To be more clear, by "the scheduler" I did indeed mean the thing generating the resource set for a job, i.e. the R stored in the KVS would have instance-local properties stripped, since they're somewhat meaningless in that context. Whether that is done by the actual scheduler or via something after the fact (libschedutil or the Python equivalent) is probably just an implementation detail and doesn't need to be in the spec (unless you'd prefer that?)

garlick: That seems right to me and you're right, libschedutil vs scheduler is implementation detail.  I was asking whether the stripping was occuring in the enclosing instance or the sub-instance.  I'm hearing enclosing instnace, which i agree with.

It's probably OK IMHO to be firm and use SHALL here as well as call out that the scheduler writes R and is responsible for this.

grondo: Ok, reworded to:

> Because instance-local properties describe the instance and not
> the resources themselves, a scheduler SHALL omit them from the *R* it
> generates to allocate resources to a job. Other components that
> generate an *R* subset MAY likewise omit them.

I added the "Other components" language here as a hedge against future work where the scheduler isn't the only thing that can subset an instance R.

garlick (approved):
LGTM

garlick (approved):
LGTM!

PR #546, opened 2026-08-10
rfc50: drop re-starting event

garlick:
Problem: the exec.eventlog `re-starting` event description doesn't match what its name implies and it's actually not needed.

Drop it.

Also clarify the `reattach` event description.


grondo (approved):
LGTM!

garlick:
Thanks!

PR #547, opened 2026-08-12
rfc27: drop sched.queue as scheduler annotation

grondo:
Problem: RFC 27 lists `sched.queue` as a known scheduler annotation, but this annotation is not used anywhere since schedulers do not assign queues.

Drop the annotation to avoid confusion.

For background, Fluxion did set a `sched.queue` annotation early on when it had a default queue config parameter, but this was removed in flux-framework/flux-sched#971.

### DIFF
--- a/spec_27.rst
+++ b/spec_27.rst
@@ -401,9 +401,6 @@ sched.reason_pending
 sched.resource_summary
   (string) human readable overview of assigned resources
 
-sched.queue
-  (string) human readable identification of job queue
-
 sched.selection_type
   (string) human readable identification of the scheduler's selected method 
   invoked to schedule the job (i.e. immediate, reserved, backfilled)


wihobbs:
I was thinking about this the other day and here's one reason to consider keeping it:

- With the inclusion of vqueues it might be good to have a record of the non-virtual queue the job actually was scheduled in for historical purposes. Is that data gathered anywhere else in the job's metadata?

edit: if that data is already collected elsewhere, I'm happy to approve this PR.

@milroy agreed with your assessment, btw, that this annotation was probably a relic from early Fluxion.

grondo:
> With the inclusion of vqueues it might be good to have a record of the non-virtual queue the job actually was scheduled in for historical purposes. Is that data gathered anywhere else in the job's metadata?

I don't think it is, but if that was added, it probably wouldn't be under the `sched.*` namespace, since queue management is now done at the flux-core level. Some sort of memo or other annotation could be added if sites wish to keep a record of the parent queue to which a vqueue maps on a per-job basis.

wihobbs:
A simple test on tuolumne and a `git grep` ~~suggests~~ edit: corroborates :) this isn't set anywhere by Fluxion. Approving.

wihobbs (approved):

grondo:
Thanks! Setting MWP.

